PR #6, opened 2018-06-06
chore: Dockerfile/ci fixes

pjenvey:
- fix the final RUN command
- include ssl for rusoto/reqwest and upgrade to latest rust
- possibly fix branch triggers (or is circleci just acting up)?
- remove unnecessary WITH_RUST to travis

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.26.0-stretch as builder
+FROM rust:1.26.2-stretch as builder
 
 ADD . /app
 WORKDIR /app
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -8,8 +8,7 @@ RUN \
     cargo --version && \
     rustc --version && \
     mkdir -m 755 bin && \
-    cargo build --release && \
-    cp /app/target/release/autopush_rs /app/bin
+    cargo install --root /app
```
bbangert: I copied this from pushbox, so that prolly needs a similar fix.

pjenvey: no fix here, just a 'refactor'. pushbox ran into a nice rust-nightly bug requiring this workaround for a few days

jrconlin: I should check if that bug is still in play, TBH.

pjenvey: I've verified it's fixed as of megaphone's pegged nightly-2018-05-05 or somewhere near there

PR #31, opened 2018-07-11
 fix: drop users with too many stored messages 

### LINKED ISSUE #25: Limit sent messages to clients
We missed porting this over to rust: the MessageOverloadException related code in websocket.py that limits the maximum amount of messages sent to clients

pjenvey:
(split into a refactor commit and the fix)

Closes #25


tests/test_integration.py
```diff
@@ -61,6 +61,7 @@
 
 ROUTER_TABLE = os.environ.get("ROUTER_TABLE", "router_int_test")
 MESSAGE_TABLE = os.environ.get("MESSAGE_TABLE", "message_int_test")
+MSG_LIMIT = 20
```
bbangert: Our current message limit is 100, do we want to hardcode 20 as the default?

PR #32, opened 2018-07-12
fix: typo and image link fix

bbangert:


jrconlin:
You know, I should remove that word from my spell check dictionary.

PR #37, opened 2018-07-17
feat: add dockerflow requirements

### LINKED ISSUE #29: Add dockerflow
Ops and SecOps have adopted [DockerFlow](https://github.com/mozilla-services/Dockerflow) which specify common set of behaviors for products.

We should add those to autopush to make life easier for Ops.

jrconlin:
Closes #29


src/server/mod.rs
```diff
@@ -946,6 +950,27 @@ fn write_status(socket: WebpushIo) -> MyFuture<()> {
     )
 }
 
+fn write_version_file(socket: WebpushIo) -> MyFuture<()> {
+    let mut file = match File::open("version.json"){
```
pjenvey: rust added an fs::read_to_string function that could remove a few lines of code here

Though I've used the include_str! macro in megaphone/pushbox instead. It compiles the version.json contents into the binary. If there's any problem reading the file rustc fails at compile time, so no need for error handling in the app

PR #39, opened 2018-07-20
feat: emit metrics for any failed notif conversions 

### LINKED ISSUE #33: Metric notification drops
The dynamodb notification loading has a few spots where messages will be dropped silently that fail deserialization. We should do basic metric counters on these just so that we understand how frequently it happens and whether we need more detailed logging of these occurrences.

pjenvey:
a separate refactor commit

Closes #33

### DIFF
--- a/src/db/commands.rs
+++ b/src/db/commands.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::fmt::Display;
 use std::rc::Rc;
 use std::result::Result as StdResult;
 use uuid::Uuid;
@@ -49,6 +50,7 @@ pub fn list_tables(
 
 pub fn fetch_messages(
     ddb: Rc<Box<DynamoDb>>,
+    metrics: &Rc<StatsdClient>,
     table_name: &str,
     uaid: &Uuid,
     limit: u32,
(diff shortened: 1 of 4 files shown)


src/db/commands.rs
```diff
@@ -66,19 +68,22 @@ pub fn fetch_messages(
         ..Default::default()
     };
 
+    let metrics = Rc::clone(metrics);
```
pjenvey: I just noticed the rust team began recommending [Rc::clone vs .clone](https://github.com/rust-lang/rust/pull/42137#issuecomment-304908345) about a year ago, makes it clearer that this is a cheap Rc clone vs worrying about it being a different clone

src/db/commands.rs
```diff
@@ -417,3 +434,30 @@ fn handle_user_result(
...
+        }
+    }
+}
+
+/// Log/metric errors during conversions to Notification
+fn conversion_err<E>(metrics: &StatsdClient, err: E, name: &'static str)
+where
+    E: Display,
+{
+    error!("Failed {} conversion: {}", name, err);
```
pjenvey: I had this as debug! but figured let's error! for now so it's seen on prod

src/db/commands.rs
```diff
@@ -417,3 +434,30 @@ fn handle_user_result(
     user.connected_at = connected_at;
     Ok(user)
 }
+
+/// Like Result::ok, convert from Result<T, E> to Option<T> but applying a
+/// function to the Err value
+fn ok_or_inspect<T, E, F>(result: StdResult<T, E>, op: F) -> Option<T>
```
pjenvey: almost tempted to make this into a whole trait extension, time will tell how useful it is

src/db/commands.rs
jrconlin: It makes sense for containers like RC (and I suppose vec may be similar), but I keep feeling that a given object should be responsible for making it's own clone, since it knows best how deep it needs to copy itself. I wonder if we'll see this pattern play out for other, similar structures?

src/db/commands.rs
bbangert: :+1:

src/db/commands.rs
jrconlin: Well, right now you're using it just to contain a call to `conversion_err`. (I *almost* wonder if it might just be better to combine them into one call to reduce some of the templating, but that's a super minor nit.)

It'll be interesting to see how these play out in the future.

src/db/commands.rs
pjenvey: Not Vec but it makes sense for smart pointers. the original RFC (which suggested a new method new_ref instead that was rejected) mentioned Rc/Arc and the associated Weak pointer types. Here's a pretty good example it outlines:

https://github.com/nical/rfcs/blob/rc-newref-clone/text/0000-rc-newref-clone.md#example

src/db/commands.rs
pjenvey: I had it that way to start with then ended up here. Already had in mind I wanted an Option::ok_or_else like call for Result (but it doesn't have one because it would be dropping Err in that case).

It had "ua.notification_read.error" either embedded inside of it or that also needed to be an argument, so went for the generic combinator I really wanted

PR #54, opened 2018-08-03
feat: log out disconnect reason with session statistics

bbangert:


src/client.rs
```diff
@@ -447,12 +448,14 @@ where
...
         session_complete: &'a mut RentToOwn<'a, AwaitSessionComplete<T>>,
     ) -> Poll<AfterAwaitSessionComplete, Error> {
-        // xxx: handle error cases with maybe a log message?
-        let _error = {
+        let error = {
             match session_complete.auth_state_machine.poll() {
                 Ok(Async::Ready(_)) => None,
                 Ok(Async::NotReady) => return Ok(Async::NotReady),
-                Err(e) => Some(e),
+                Err(e) => match Error::from(e) {
```
pjenvey: is the Error::from actually needed?

PR #56, opened 2018-08-06
feat: support webpush API pings

### LINKED ISSUE #55: Support webpush API pings
We only support emitting websocket level ping frames. Clients also ping us with "older" style webpush API pings (empty json dicts "{}"), which they expect a response to.

pjenvey:
Closes #55

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -884,6 +884,20 @@ class TestRustWebPush(unittest.TestCase):
         assert client.uaid != uaid
         yield self.shut_down(client)
 
+    @inlineCallbacks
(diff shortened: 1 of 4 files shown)


tests/test_integration.py
```diff
@@ -884,6 +884,20 @@ def test_msg_limit(self):
         assert client.uaid != uaid
         yield self.shut_down(client)
 
+    @inlineCallbacks
+    def test_can_ping(self):
+        client = yield self.quick_register()
+        yield client.ping()
```
bbangert: Verify the response came back down?

PR #62, opened 2018-08-24
feat: notify other nodes if user has reconnected for missed messages

### LINKED ISSUE #58: no lookup/notify of nodes after storing unacked messages on disconnects
I don't recall us intentionally omitting this from autopush-rs:

https://github.com/mozilla-services/autopush/blob/5a2c3d0/autopush/websocket.py#L591

When storing unacked messages due to a client disconnecting, we also check if they've already reconnected to another node and if so notify them to check storage

bbangert:
Closes #58


src/client.rs
```diff
@@ -542,6 +539,55 @@ where
...
+                    }
+                } else {
+                    future::err("No new node_id, notify not needed".into())
+                }
+            })
+            .and_then(|(client, uaid, node_id)| {
+                // Send the notify to the user
+                let notify_url = format!("{}/notif/{}", node_id, uaid.simple());
+                client
+                    .post(&notify_url)
```
pjenvey: it's a PUT not POST

bbangert: Ooo, nice catch!

PR #65, opened 2018-08-29
feat: upgrade sentry to 0.8.0 and log out errors

### LINKED ISSUE #5: Upgrade Sentry in Rust connection node to the newest version
https://blog.sentry.io/2018/05/01/sentry-for-rust announced 0.3, they're now past 0.5.

This does require a new Sentry in our deployment env, so its blocked right now.

bbangert:
Previously we only logged panics to sentry, this commit add's logging
of all errors except several ignored classifications.

Closes #5


bbangert:
The test had to be skipped, because the local HTTP server is not getting the data for some reason. However when running it and having an ```nc``` server get it, it does work. I've dug into the HTTP server Python sets up and added more debugging, but nothing seems to come in for unknown reasons.

I'd like to see if this works on dev/stage so we can proceed with a release tag tomorrow, and then spend however much time is needed to determine why the local Python HTTP server is not getting the request when the data seems to be sent in a full test environment.

PR #73, opened 2018-09-06
fix: use latest sentry

bbangert:
A prior version of Sentry was less strict on guard handling, the
latest fully tears down the transport. The guard now lives for the
duration of the server to avoid the teardown.


src/server/mod.rs
```diff
@@ -92,10 +92,23 @@ impl AutopushServer {
         }
     }
 
-    pub fn start(&self) {
+    pub fn start(&self) -> Option<sentry::internals::ClientInitGuard> {
         logging::init_logging(!self.opts.human_logs).expect("init_logging failed");
         let handles = Server::start(&self.opts).expect("failed to start server");
         self.shutdown_handles.set(Some(handles));
+        if let Ok(dsn) = env::var("SENTRY_DSN") {
```
pjenvey: nit: we could just store the guard inside of AutopushServer. I suppose this could all go in new above to make that easier

PR #79, opened 2018-10-04
chore: upgrade dependencies

### LINKED ISSUE #78: Update dependencies [Q4]
howto: https://github.com/mozilla-services/autopush/issues/1170#issuecomment-378776031

I may attempt to finish my previous effort from last quarter to upgrade to hyper 0.12

pjenvey:
split in 2 commits for review

Closes #78

Also see https://github.com/mozilla-services/autopush-rs/compare/chore/78...chore/hyper0.12-take2 and its commit message for my 2nd attempt to upgrade hyper (because misery loves company)

It appears to work but there's some caveats w/ ClientRegistry's locking that I don't want to spend any more time on at the moment

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -17,7 +17,7 @@ use futures::sync::mpsc;
 use futures::sync::oneshot::Receiver;
 use futures::AsyncSink;
 use futures::{Async, Future, Poll, Sink, Stream};
-use reqwest::unstable::async::Client as AsyncClient;
+use reqwest::async::Client as AsyncClient;
 use rusoto_dynamodb::UpdateItemOutput;
 use sentry;
 use sentry::integrations::error_chain::event_from_error_chain;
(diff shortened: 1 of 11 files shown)


src/client.rs
```diff
@@ -586,7 +585,7 @@ fn save_and_notify_undelivered_messages(
                 } else if let Some(node_id) = user.node_id {
                     let result = AsyncClient::builder()
                         .timeout(Duration::from_secs(1))
-                        .build(&srv3.handle);
+                        .build();
```
bbangert: The async client uses its own event loop then?

pjenvey: w/ the tokio reform, going forward you don't really pass Handles around. Now there's kind of a global event loop. Taskish things generally get an executor to schedule themself from a globalish CurrentThread executor or DefaultExecutor, I'm not entirely sure which one in this case.

But basically it should be running on the current executor (which sets itself as one of those when it starts)

bbangert: That's a lot of ish-ish! ;)

pjenvey: very ish-ish hand waivey, I don't have a good grip on all this yet

PR #88, opened 2018-12-12
chore: cargo fmt (1.31.0)

pjenvey:
tidies up the overhanging chained methods, so also revert:
https://github.com/mozilla-services/autopush/commit/fd48405


src/client.rs
```diff
@@ -510,11 +510,21 @@ where
...
+                .insert("ua_os_family".to_string(), metrics_os.to_string());
+            event
+                .tags
+                .insert("ua_os_ver".to_string(), ua_result.os_version.to_string());
+            event
+                .tags
+                .insert("ua_browser_family".to_string(), metrics_browser.to_string());
+            event
+                .tags
+                .insert("ua_browser_ver".to_string(), ua_result.version.to_string());
```
bbangert: :man_shrugging:

pjenvey: :do_not_litter:
rustfmt :x:

PR #90, opened 2019-01-03
fix: return correct not found for disconnected client and fix tests

### LINKED ISSUE #89: For offline notifications, connection endpoint replies back 502 Bad Gateway response code to Router Endpoint instead of replying 404 Not Found
Connection endpoint replies back 502 Bad Gateway response code to Router Endpoint instead of replying 404 Not Found. When sending a notification while the browser is disconnected (offline),  connection endpoint (or Rust websocket server), reports 502 to the Python router, a response code that is not expected by the router. In the original implementation, and in such scenario, router used to receive 404 from the Python connection node.

bbangert:
Tests weren't passing because autopush had a dependency change for FCM.
This updates the travis test to pull down the current dependencies for
autopush for integration tests.

Closes #89


src/http.rs
```diff
@@ -57,7 +57,7 @@ impl Service for Push {
                             Ok(hyper::Response::new().with_status(StatusCode::Ok))
                         } else {
                             Ok(hyper::Response::new()
-                                .with_status(StatusCode::BadGateway)
+                                .with_status(StatusCode::NotFound)
```
pjenvey: The check_client_storage call below should also potentially 404 too

PR #104, opened 2019-02-14
chore: tag 1.53.1

pjenvey:


Cargo.lock
```diff
@@ -192,7 +192,7 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.7.2"
+version = "0.7.0"
```
pjenvey: I run `cargo update -p autopush` to set the correct autopush version in Cargo.lock

It checked in w/ crates.io and downgraded block-buffer because the last 2 versions were yanked for an unknown reason. 

autopush-rs 1.5.0 was already using 0.7.0 anyway so ¯\_(ツ)_/¯

PR #111, opened 2019-07-29
chore: update deps minus hyper 0.12

### LINKED ISSUE #110: Update dependencies [Q3]
Let's update the rust deps when we get to autopush's updates: https://github.com/mozilla-services/autopush/issues/1339

TODO: should add "update the python/rust in the Dockerfile" to the dep update checklist

pjenvey:
Closes #110


autopush/src/client.rs
```diff
@@ -863,7 +861,7 @@ where
                     Ok(endpoint) => {
                         data.srv
```
jrconlin: silly question, but why not use a ref to the `srv` you just cloned above?

pjenvey: good question actually. the Rc::clone above wasn't even needed apparently

PR #116, opened 2019-09-19
fix: restore the log crate

### LINKED ISSUE #115: Quiet excessive logging
We lost the log crate dependency along with its Cargo `features = ["max_level_info", "release_max_level_info"]` when we split the code into 2 separate crates.

This causes excessive trace/debug logging to be on by default, so we'll need to add this back.

pjenvey:
along with its max log level restrictions to quiet excessive logging

also include env_logger

Closes #115


jrconlin:
https://travis-ci.org/mozilla-services/autopush-rs/builds/587146509 and https://travis-ci.org/mozilla-services/autopush-rs/builds/587146507 are both green, but are not reporting back here correctly. going to override for now.

PR #120, opened 2019-10-02
fix: report Megaphone errors to Sentry

### LINKED ISSUE #118: Notify sentry on MegaphoneUpdater failures
Currently we write to debug log when MegaphoneUpdater attempts fail, with a TODO to notify sentry.

With debug logs being hidden by default on prod, we should begin notifying sentry about issues communicating with megaphone instead.

For the sake of diagnosing such issues: https://bugzilla.mozilla.org/show_bug.cgi?id=1585042

leplatrem:
fixes #118 

I wanted to use `capture_error(&error)` but didn't manage to convert errors types from  `autopush_common::errors` :grin:

### DIFF
--- a/autopush/src/server/mod.rs
+++ b/autopush/src/server/mod.rs
@@ -22,7 +22,7 @@ use hyper::{self, StatusCode};
 use openssl::hash;
 use openssl::ssl::SslAcceptor;
 use reqwest;
-use sentry::{self, integrations::panic::register_panic_handler};
+use sentry::{self, capture_message, integrations::panic::register_panic_handler};
 use serde_json::{self, json};
 use time;
 use tokio_core::net::TcpListener;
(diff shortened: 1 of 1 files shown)


autopush/src/server/mod.rs
```diff
@@ -639,9 +639,12 @@ impl Future for MegaphoneUpdater {
                             }
                         }
                         Ok(Async::NotReady) => return Ok(Async::NotReady),
-                        Err(_) => {
-                            // TODO: Flag sentry that we can't poll megaphone API
-                            debug!("Failed to get response, queue again");
+                        Err(error) => {
+                            error!("Failed to get response, queue again {:?}", error);
+                            capture_message(
```
pjenvey: Sentry also has a `capture_error_chain` that might work (though I've never tried it..)

jrconlin:
Going to merge this as is. 

@leplatrem I don't know if there is a way to test it locally. We can open up another issue later if we decide that `capture_error_chain` is a better approach.

pjenvey:
In the past my sentry "test" has been running a local listening netcat alongside a SENTRY_DSN=http://foo:bar@localhost:9999/etc :)

PR #134, opened 2020-02-29
chore: update deps

### LINKED ISSUE #131: [Q1] Update dependencies
Most rust deps & rust itself

pjenvey:
- continue avoiding hyper 0.12/13 and futures 0.3 for now
- avoid updates to the tungstenite crates and
  reqwest/tokio-openssl/bytes which assume a more recent hyper or
  futures
- replace the deprecated chan_signal crate w/ crossbeam_channel +
  signal_hook

Closes #131


autopush/src/main.rs
```diff
@@ -61,3 +60,17 @@ fn main() -> Result<()> {
     signal.recv().unwrap();
     server.stop().chain_err(|| "Failed to shutdown properly")
 }
+
+/// Create a new channel subscribed to the given signals
```
jrconlin: So, if I understand, `notify` starts a rust thread to monitor for POSIX signals, and then direct them back to the main thread via a `crossbeam_channel` buffered for 100 messages? (To replace the deprecated `chan_signal` library.)

I'm curious why the buffer is so large, particularly for 2 kill signals.

fzzzy: why not have it be large? memory is cheaper than impossible to reproduce runtime errors.

pjenvey: It's 100 because I copied the recommended impl. from https://crates.io/crates/chan-signal and it's the same buffer size `chan_signal::notify` used.

It could be lower for our use case but it's not terribly important

PR #141, opened 2020-05-01
133 update to hyper 13

fzzzy:


autopush-common/src/db/mod.rs
```diff
@@ -336,74 +336,100 @@ impl DynamoStorage {
...
+        }))
+    }
+
+    pub async fn check_storage(
+        &self,
+        table_name: &str,
+        uaid: &Uuid,
+        include_topic: bool,
+        timestamp: Option<u64>,
+    ) -> LocalBoxFuture<'static, Result<CheckStorageResponse>> {
```
fzzzy, line 423: This needs to be changed to `) -> Result <CheckStorageResponse> {`

autopush-common/src/db/mod.rs
```diff
@@ -336,74 +336,100 @@ impl DynamoStorage {
...
+        let uaid = *uaid;
+        let table_name = table_name.clone().to_string();
+
+        let resp = self.check_include_topic(
+            include_topic,
+            table_name.clone(),
+            &uaid
+        ).await;
+        self.check_if_empty(table_name, &uaid, include_topic, timestamp, resp)
```
fzzzy, line 432: And this needs to change to `self.check_if_empty(table_name, &uaid, include_topic, timestamp, resp).await`

Issue #148, opened 2020-06-02
Switch integration tests to circle-ci


jrconlin:
Travis is not being very reliable. To not be blocked on builds, we should switch to circle-ci.

(part of https://github.com/mozilla-services/services-engineering/issues/60)

PR #149, opened 2020-06-02
feat: Switch integration tests to circleci

jrconlin:
Issue: #148


.circleci/config.yml
```diff
@@ -7,6 +7,34 @@
 #
 version: 2
 jobs:
+  test:
+    docker:
+      - image: circleci/python:2.7.1
+    steps:
+      - checkout
+      - run:
+          # can this be saved? The python integration tests generally don't change.
```
Mcat12, line 16: We could use circle's cache, though the challenge is invalidating it when the autopush dependencies change.

jrconlin: Yeah, I believe we can probably lock the python integration test for now without much concern. Version updates inevitably will start pulling in py3 dependencies and until we can move these tests to py3, we're probably going to have to not use those.

jrconlin:
Let's close this for now.

PR #150, opened 2020-06-03
feat: Add autoendpoint crate

### LINKED ISSUE #102: Setup endpoint crate structure and required dependencies
Create the autoendpoint crate, copy/update main/settings from autopush and setup basic module structure with the actix framework.

Mcat12:
Based off of the https://github.com/mozilla-services/autoendpoint-rs template, but with some changes:
- Removed unused dependencies
- Replaced `failure` (deprecated) with `thiserror`
- Some small refactoring

Closes #102


autoendpoint/src/error.rs
```diff
@@ -0,0 +1,109 @@
...
+    Io(#[from] std::io::Error),
+
+    #[error("{0}")]
+    Metrics(#[from] cadence::MetricError),
+
+    #[error("{0}")]
+    Internal(String),
+}
+
+impl ApiError {
```
pjenvey, line 33: One note about our experience w/ a top level ApiError in syncstorage: I'm beginning to wonder if it's really worth having -- it was originally created w/ an older version of actix-web w/ a slightly different Error type.

Being the centralized error, it's nice in that it provides a centralized place for implementing the actix-web ResponseError (that renders the error), and a couple other misc bits.

On the downside it's a little annoying having to convert from say e.g. `DbError` .into() an `ApiError`, then .into() again a `ResponseError` (in some cases [we have 3 into()s](https://github.com/mozilla-services/syncstorage-rs/blob/7203b8f/src/web/handlers.rs#L221), the first for the `DbErrorKind`. granted, async/await might allow more usages of `?` which probably avoid these into()s entirely). Identifying certain types of errors at this level is also [a bit cumbersome](https://github.com/mozilla-services/syncstorage-rs/blob/b2d9341/src/error.rs#L106).

Metrics and io Error likely only happen during startup (`main`)?, so they wouldn't necessarily benefit from ApiError's features.

Could we avoid the ApiError and just have multiple types of errors specific to modules handle rendering themselves (without lots of code duplication everywhere)? Just something to keep in mind going forward: there's possibly room for improvement here.

autoendpoint/src/error.rs
```diff
@@ -0,0 +1,109 @@
...
+    dev::{HttpResponseBuilder, ServiceResponse},
+    error::ResponseError,
+    http::StatusCode,
+    middleware::errhandlers::ErrorHandlerResponse,
+    HttpResponse, Result,
+};
+use serde::ser::SerializeMap;
+use serde::{Serialize, Serializer};
+use std::convert::From;
+use thiserror::Error;
```
pjenvey, line 13: 👍

autoendpoint/src/error.rs
Mcat12: I think I've addressed the issues you brought up in the latest commit:
- The `ApiError` enum is renamed to `ApiErrorKind` and a new `ApiError` struct holds both a kind and a backtrace.
- The number of `From`/`Into` conversions can be reduced by using error kind enums for underlying errors (`DbErrorKind`) and wrapping them in a variant on `ApiErrorKind`, ex. `ApiErrorKind(DbErrorKind)`. The functionality brought about by `DbError` is mostly backtrace support, which is already present in `ApiError`. Additionally, I've added a `From` impl to `ApiError` which forwards the `From` impls from `ApiErrorKind`:
  https://github.com/mozilla-services/autopush-rs/blob/18b1063fc1cc600dfd434e5f46b52a3a61944de8/autoendpoint/src/error.rs#L88-L100
- Related to the previous point, in the case where we want to further isolate a database module such that it does not know about `ApiError` but still needs to act as an HTTP response, we can fallback to using `DbError` and add a "short-cut" impl: `impl From<DbError> for HttpResponse` which does the chain of conversion to `ApiError` and then `HttpResponse`. This would hide the repeated conversions during normal use.
- Identifying the type of error is easier now, as you can do a direct match:
  ```rust
  match error.kind {
      ApiErrorKind::Db(DbErrorKind::BatchNotFound) => {
          // ...
      },
      // some other type of error...
  }
  ```
- While the metrics and IO errors may only happen during startup, I don't see how they are disadvantaged by using the `ApiError` type. We could remove them from the list of errors and handle them explicitly, but that does not affect the decision to use a top level error or not. Additionally, using `ApiError` allows us to handle a function which could return either error type, such as `Server::with_settings` (and we have backtrace support, which `Box<dyn Error>` does not necessarily have).

If we do use multiple root-level error types instead of one, I would be concerned in two ways:
- Their rendering may not be the same, which would be bad for API consumers. For example, if DB errors had a different schema compared to logic errors.
- At some point we would likely have to handle two different error types in the same function. How should we return the error? As a `Box<dyn Error>`? This is worse than using a single root-level error type.

Let me know what you think of this, and if I missed anything.

fzzzy: > add a "short-cut" impl: impl From<DbError> for HttpResponse which does the chain of conversion to ApiError and then HttpResponse. This would hide the repeated conversions during normal use.

 👍👍👍

fzzzy: > match error.kind {
>     ApiErrorKind::Db(DbErrorKind::BatchNotFound) => {

This helped me to understand why people use the ApiError/ApiErrorKind division. Thanks.

PR #152, opened 2020-06-11
feat: only register new users on new subscription requests

### LINKED ISSUE #137: Only register user on new subscription request
[Bug 1617136](https://bugzilla.mozilla.org/show_bug.cgi?id=1617136) will increase the number of writes we currently are making to DynamoDB. 

In order to reduce those, we should consider only recording the now quite disposable UAID we issue in the `hello` message when we absolutely need to.

~From doing a cursory overview of the code, I believe it's possible to alter `DynamoDbUser` to include the `chids`, modify `handle_user_result` to fetch the uaid meta record containing the chids, and then use that to determine if `hello` should call `register_user` if `chids` are not empty.~ (Optionally, we could also include a `registered` boolean inside of DynamoDB to indicate if this user is registered or not)

*IF* the client is well behaved, we should only get a UAID when there are existing ChannelIDs. We could verify by performing an UPDATE write to the Router Table, returning a new UAID if that fails (and possibly resetting any `registered` flags, since the UA is now effectively unregistered.)

On any future `register` call, we can either first check if `registered` is `false` or just call `register_user` with the values already set in `DynamoDbUser` to record the user into the `router` table, then proceed to record the new information for the `message*` table.

pjenvey:
when no UAID is specified from desktop clients, defer writing them to
the router table until they subscribe to a channel

Avoids unwriting unnecessary router entries, especially important
after https://bugzilla.mozilla.org/show_bug.cgi?id=1617136

Co-authored by: jrconlin <jr+git@mozilla.com>

Closes #137

### DIFF
--- a/autopush-common/src/db/mod.rs
+++ b/autopush-common/src/db/mod.rs
@@ -27,7 +27,7 @@ use self::commands::{
     retryable_batchwriteitem_error, retryable_delete_error, retryable_updateitem_error,
     FetchMessageResponse,
 };
-use self::models::{DynamoDbNotification, DynamoDbUser};
+pub use self::models::{DynamoDbNotification, DynamoDbUser};
 
 const MAX_EXPIRY: u64 = 2_592_000;
 const USER_RECORD_VERSION: u8 = 1;
@@ -46,6 +46,8 @@ pub struct HelloResponse {
     pub reset_uaid: bool,
     pub rotate_message_table: bool,
     pub connected_at: u64,
+    // Exists when we didn't register this user during HELLO
+    pub deferred_user_registration: Option<DynamoDbUser>,
 }
 
 pub struct CheckStorageResponse {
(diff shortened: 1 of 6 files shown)


autopush-common/src/db/mod.rs
```diff
@@ -46,6 +46,8 @@ pub struct HelloResponse {
     pub reset_uaid: bool,
     pub rotate_message_table: bool,
     pub connected_at: u64,
+    // Exists when we didn't register this user during HELLO
+    pub deferred_user_registration: Option<DynamoDbUser>,
```
jrconlin, line 50: much better name. 👍

autopush-common/src/db/mod.rs
```diff
@@ -199,12 +222,40 @@ impl DynamoStorage {
         channel_id: &Uuid,
         message_month: &str,
         endpoint: &str,
+        register_user: Option<&DynamoDbUser>,
```
jrconlin, line 225: I wonder if we should call this `deferred_user`. If so, we can if someone is aesthetically inclined.

Issue #103, opened 2019-02-14
Create extractor/ validators for autoendpoint API methods


bbangert:
autoendpoint has a fairly large amount of validation required, this should all be ported in as actix extractors and custom validation logic as needed (syncstorage-rs has some examples of this).

Depends on #102.

PR #153, opened 2020-06-12
feat: Add more autoendpoint WebPush validation

Mcat12:
Generally, this PR fills in the validation TODOs in the `Subscription` extractor.

Specific changes:
- Validate the VAPID public key against the v2 token
- Add and enforce maximum payload size
- Validate and store encryption headers
- Fix some issues with Fernet key configuration and decryption
- Renamed `WebPushHeaders` to `NotificationHeaders`
- Other minor refactoring

Some of the remaining work before #103 can be closed includes JWT token validation and any extractors used by the other endpoints (e.g. registration endpoints), but this PR is already big enough (perhaps too big).

Related to #103.


Mcat12:
Restarting CI after https://github.com/mozilla-services/autopush/pull/1399

autoendpoint/src/server/extractors/subscription.rs
```diff
@@ -1,66 +1,158 @@
...
+            let state: Data<ServerState> =
+                Data::extract(&req).await.expect("No server state found");
+            let fernet = state.fernet.as_ref();
+
+            // Decrypt the token
+            let token = fernet
+                .decrypt(&repad_base64(&token_info.token))
+                .map_err(|_| ApiErrorKind::InvalidToken)?;
+
+            // Parse VAPID and extract public key
```
pjenvey, line 43: nit: I think this block could be restricted to only running when `api_version == "v2"`.

Mcat12:
> Also for the future, keep an eye out for the errnos (on top of status codes) we associate with error responses.

I've opened an issue to make sure I clean this up: #159

autoendpoint/src/server/extractors/subscription.rs
Mcat12: VAPID can still be provided for v1 endpoints, so if we gate VAPID on v2 we will lose metrics for VAPID usage on v1 endpoints. Is that expected?

jrconlin: No, we want to know overall vapid use, not just how many use it for subscription locking. Likewise, we want to disallow invalid VAPID headers because it breaks the agreement around VAPID use (even if we're not using the data).

pjenvey: > VAPID can still be provided for v1 endpoints, so if we gate VAPID on v2 we will lose metrics for VAPID usage on v1 endpoints (in addition to allowing invalid VAPID headers). Is that expected?

Ah I've misread the Python, you are correct

PR #160, opened 2020-06-22
feat: Validate user subscription data in autoendpoint

### LINKED ISSUE #156: [autoendpoint] Validate the user and subscription when handling incoming notifications
See `WebPushSubscriptionSchema::validate_uaid_month_and_chid` in the Python autopush server.

This validation makes sure that the user and subscription/channel exist and are well formed (not missing required data). There is also some logic for updating users and dropping inactive users. Lastly, the user data is made available in the subscription object.

This functionality may need some refactoring; there's a lot of different actions being done in this "validation". This will also be one of the first autoendpoint features which uses the database, so that may increase the estimate of this issue.

Mcat12:
User subscription data is now validated as part of the Actix extractor validations. Specifically:
- Validate router type
- (WebPush) Check that the subscription/channel exists
- (WebPush) Drop user if inactive

This is the first autoendpoint feature to use the database, so support has been added. Part of those changes simplified `DynamoStorage` and made it `Send + Sync`.

Closes #156

### DIFF
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -59,6 +59,9 @@ pub enum ApiErrorKind {
     #[error(transparent)]
     VapidError(#[from] VapidError),
 
+    #[error(transparent)]
+    Uuid(#[from] uuid::Error),
+
     #[error("Error while validating token")]
     TokenHashValidation(#[from] openssl::error::ErrorStack),
 
@@ -67,6 +70,9 @@ pub enum ApiErrorKind {
 
     #[error("Invalid token")]
     InvalidToken,
+
+    #[error("No such subscription")]
+    NoSubscription,
 
     /// A specific issue with the encryption headers
     #[error("{0}")]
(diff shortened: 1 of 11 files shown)


autoendpoint/src/error.rs
```diff
@@ -59,12 +59,21 @@ pub enum ApiErrorKind {
     #[error(transparent)]
     VapidError(#[from] VapidError),
 
+    #[error(transparent)]
+    Uuid(#[from] uuid::Error),
```
jrconlin, line 63: _(mutters something about preferring errors be discretely named values like `ApiErrorKind::UuidError`, but realizes it's probably not worth fighting over.)_

Mcat12: I was going back and forth about this. It's nice to be clear that it's an error type, but `ApiErrorKind::Uuid` is pretty clear that it's an error. Adding `Error` on to the end of every variant seems pretty redundant.

jrconlin: yeah, like i said, not really worth fighting over, particularly since this instance of `Uuid` (or `Database`) should never appear without it's type.

PR #166, opened 2020-06-23
feat: Record the encoding in a metric if there is an encrypted payload

Mcat12:
This improves parity with Python autoendpoint.


Mcat12:
Going to let #154 merge first before rebasing this PR.

PR #167, opened 2020-06-25
feat: Route notifications to autopush connection servers

### LINKED ISSUE #161: [autoendpoint] Route notifications to autopush connection servers
WebPush notifications (destined for desktop Firefox user agents) should be routed to the autopush connection servers.

This is the first router implementation, which may increase the estimate.

Mcat12:
Major changes:
- Added `routers` module to handle routing notifications.
- WebPush notifications are routed to Autopush connection servers, following the same logic used in Python autoendpoint.
- Added `Routers` extractor to expose routers to HTTP handlers.

Minor changes:
- Notification message IDs are now generated
- Use a `RouterType` enum to make better guarantees about the possible `router_type` values.
- TTL defaults to 0. There's a TODO questioning if this is correct, as the Python code is somewhat inconsistent. Related to https://github.com/mozilla-services/autopush/issues/358
- Removed unused database settings and settings banner. Actix already prints the banner info.

Closes #161

### DIFF
--- a/autoendpoint/src/server/extractors/notification_headers.rs
+++ b/autoendpoint/src/server/extractors/notification_headers.rs
@@ -19,7 +19,7 @@ const MAX_TTL: i64 = 60 * 60 * 24 * 60;
 pub struct NotificationHeaders {
     // TTL is a signed value so that validation can catch negative inputs
     #[validate(range(min = 0, message = "TTL must be greater than 0", code = "114"))]
-    pub ttl: Option<i64>,
+    pub ttl: i64,
 
     #[validate(
         length(
@@ -53,7 +53,12 @@ impl NotificationHeaders {
         let ttl = get_header(req, "ttl")
             .and_then(|ttl| ttl.parse().ok())
             // Enforce a maximum TTL, but don't error
-            .map(|ttl| min(ttl, MAX_TTL));
+            .map(|ttl| min(ttl, MAX_TTL))
+            // TODO: Should we return an error if no TTL is given instead of defaulting to 0?
+            //       The Python code often defaults to 0, but there's at least one case where an
+            //       error is returned if the TTL is missing:
+            //       https://github.com/mozilla-services/autopush/commit/1f01cd70f52de3c22f74a7389019dfafd1d90ea7#diff-170ab1658a6a917eee357db65564b076R111-R112
+            .unwrap_or(0);
         let topic = get_owned_header(req, "topic");
         let content_encoding = get_owned_header(req, "content-encoding");
         let encryption = get_owned_header(req, "encryption");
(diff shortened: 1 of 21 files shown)


autoendpoint/src/server/extractors/notification_headers.rs
```diff
@@ -53,17 +79,22 @@ impl NotificationHeaders {
         let ttl = get_header(req, "ttl")
             .and_then(|ttl| ttl.parse().ok())
             // Enforce a maximum TTL, but don't error
-            .map(|ttl| min(ttl, MAX_TTL));
+            .map(|ttl| min(ttl, MAX_TTL))
+            // TODO: Should we return an error if no TTL is given instead of defaulting to 0?
```
jrconlin, line 83: TTL was introduced after the commonly used draft but before the RFC was released. At this point, I think it's fine to return an error:
https://tools.ietf.org/html/rfc8030#section-5.2

autoendpoint/src/server/extractors/user.rs
```diff
@@ -5,31 +5,59 @@ use crate::server::ServerState;
...
+        match s {
+            "webpush" => Ok(RouterType::WebPush),
+            "gcm" => Ok(RouterType::GCM),
+            "fcm" => Ok(RouterType::FCM),
+            "apns" => Ok(RouterType::APNS),
+            "adm" => Ok(RouterType::ADM),
+            _ => Err(()),
+        }
+    }
+}
```
jrconlin, line 34: This is such a common pattern, I'm surprised there's not a macro for it.

autoendpoint/src/server/routers/webpush.rs
```diff
@@ -0,0 +1,217 @@
...
+        self.metrics.incr("updates.client.host_gone").ok();
+
+        self.ddb
+            .remove_node_id(&user.uaid, node_id, user.connected_at)
+            .compat()
+            .await
+            .map_err(|e| ApiErrorKind::Database(e).into())
+    }
+
+    /// Update metrics and create a response for when a notification has been directly forwarded to
```
jrconlin, line 176: Not like it's a huge deal, but why the intermediary function? It's adding a few static args. 

"Aesthetics" is a reasonable answer.

autoendpoint/src/server/extractors/user.rs
Mcat12: There are third-party derives available, but I'm not sure they're worth the extra dependencies and compile time.
https://docs.rs/enum-utils/0.1.1/enum_utils/derive.FromStr.html

autoendpoint/src/server/routers/webpush.rs
Mcat12: 1. The function is used multiple times and this avoids duplicating those static args. Now we only need to change the args in one place.
2. The Python code had this pattern, and it seemed reasonable enough to reuse. One difference is that the Python code duplicated the metrics and response object builder code, which I've moved into `make_response`.

PR #170, opened 2020-06-29
feat: Return detailed autoendpoint errors

### LINKED ISSUE #159: [autoendpoint] Add errno values to errors and improve specificity of errors
Autoendpoint should emit detailed errors with associated error numbers and a link to http://autopush.readthedocs.io/en/latest/http.html#error-codes.

Currently, the errors are relatively detailed but lack error numbers and do not fully conform to the expected HTTP codes.

This work is more about polish, and it should probably happen after the major autoendpoint features are implemented.

Mcat12:
Errors now include an optional error number and detailed errors are returned to the application server.

A minor change is that `DynamoStorage::get_user` now returns an `Option`. 

Closes #159

### DIFF
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -74,6 +74,9 @@ pub enum ApiErrorKind {
     #[error("Invalid token")]
     InvalidToken,
 
+    #[error("UAID not found")]
+    NoUser,
+
     #[error("No such subscription")]
     NoSubscription,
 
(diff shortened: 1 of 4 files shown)


autoendpoint/src/error.rs
```diff
@@ -174,40 +194,23 @@ impl From<ApiError> for HttpResponse {
...
-        // To return a descriptive error response, this would work. We do not
-        // unfortunately do that so that we can retain Sync 1.1 backwards compatibility
-        // as the Python one does.
-        // HttpResponse::build(self.status).json(self)
-        //
-        // So instead we translate our error to a backwards compatible one
-        HttpResponse::build(self.kind.status())
-            .header("Retry-After", RETRY_AFTER.to_string())
-            .finish()
+        HttpResponse::build(self.kind.status()).json(self)
```
jrconlin, line 197: Why are we not using `Retry-After`?
Should it be an option for `error_response`?

Mcat12: The Python code does not use `Retry-After`. This is an artifact of the template taken from sync-storage ([only set if the error is `DbErrorKind::Conflict`](https://github.com/mozilla-services/syncstorage-rs/blob/b2d9341824d3bb7b722e75a5aaaa2e4096007e20/src/error.rs#L226-L228)).

jrconlin: Ah! Sorry, you're right.

PR #171, opened 2020-07-09
feat: Route notifications to FCM (Android)

### LINKED ISSUE #162: [autoendpoint] Route FCM notifications
WebPush notifications (destined for Android user agents using FCM) should be routed through Firebase Cloud Messaging.

Questions:
- Do both versions of FCM need to be supported (legacy and v1)?
  Answer: Only support v1

Mcat12:
This branch is based on `feat/autoendpoint-webpush-router`, but will be based on master once #167 merges. Until then this PR will be a draft.

Major changes:
- Add the Firebase Cloud Messaging (FCM) router to support routing notifications to Android devices.

Minor changes:
- Added `.gitguardian.yml` to ignore the fake OAuth credentials used during the FCM tests. [config docs](https://github.com/GitGuardian/gg-shield#configuration)
- Added a note to `operation_notes.md` about how to set nested settings via environment variables.
- Unknown settings are now an error (serde's `deny_unknown_fields`)
- Added an `InsertOpt` trait to `autopush_common` to make conditionally inserting an `Option` into a `HashMap` easier.

Closes #162


autoendpoint/src/server/routers/fcm/client.rs
```diff
@@ -0,0 +1,301 @@
...
+use reqwest::StatusCode;
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::time::Duration;
+use url::Url;
+use yup_oauth2::authenticator::DefaultAuthenticator;
+use yup_oauth2::ServiceAccountAuthenticator;
+
+/// The timeout for FCM requests, in seconds
+const TIMEOUT: Duration = Duration::from_secs(3);
```
jrconlin, line 12: kinda wonder if this should be a config option.

autoendpoint/src/server/routers/fcm/client.rs
```diff
@@ -0,0 +1,301 @@
...
+    use tempfile::NamedTempFile;
+    use url::Url;
+
+    pub const PROJECT_ID: &str = "yup-test-243420";
+    const ACCESS_TOKEN: &str = "ya29.c.ElouBywiys0LyNaZoLPJcp1Fdi2KjFMxzvYKLXkTdvM-rDfqKlvEq6PiMhGoGHx97t5FAvz3eb_ahdwlBjSStxHtDVQB4ZPRJQ_EOi-iS7PnayahU2S9Jp8S6rk";
+
+    /// Write service data to a temporary file
+    pub fn make_service_file() -> NamedTempFile {
+        // Taken from the yup-oauth2 tests
+        let contents = serde_json::json!({
```
jrconlin, line 134: kinda also wondering if we should put this as a json file to avoid some of the potential key flagging we're seeing. We could have the CI system include the key info as part of the build.

autoendpoint/src/server/routers/fcm/router.rs
```diff
@@ -0,0 +1,398 @@
...
+    use crate::server::FcmSettings;
+    use autopush_common::db::DynamoDbUser;
+    use cadence::StatsdClient;
+    use std::collections::HashMap;
+    use std::path::PathBuf;
+    use url::Url;
+    use uuid::Uuid;
+
+    const FCM_TOKEN: &str = "test-token";
+    const CHANNEL_ID: &str = "4530d3a6-13f9-4639-87f9-2ff731824f34";
```
jrconlin, line 230: Traditionally, I've used UUIDs that start with something like "DEADBEEF". This lets me scan and purge the database if the test leaks. 

(And it may not be your code that leaks. Someone later on could add something, use the testing value and that winds up corrupting things.)

autoendpoint/src/server/routers/fcm/client.rs
Mcat12: The JSON file would still be flagged by the tool, since it's a secret that's checked into Git. Hiding it in CI would break local testing, unless everyone had a copy of the config (which causes issues for new people when first cloning and testing the repo).

Telling GitGuardian to ignore the file seems to be the most correct fix.

autoendpoint/src/server/routers/fcm/router.rs
```diff
@@ -0,0 +1,398 @@
...
+            .and_then(Value::as_str)
+            .ok_or(FcmError::NoAppId)?;
+        let ttl = MAX_TTL.min(self.settings.ttl.max(notification.headers.ttl as usize));
+        let message_data = self.build_message_data(notification)?;
+
+        // Send the notification to FCM
+        let client = self.clients.get(app_id).ok_or(FcmError::InvalidAppId)?;
+        trace!("Sending message to FCM: {:?}", message_data);
+        if let Err(e) = client.send(message_data, fcm_token.to_string(), ttl).await {
+            return Err(self.handle_error(e));
```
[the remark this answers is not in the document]
jrconlin: Added: https://github.com/mozilla-services/autopush-rs/issues/190

PR #174, opened 2020-07-10
fix: Copy and upgrade parts of DynamoStorage into autoendpoint

### LINKED ISSUE #172: Fix incompatibility between Tokio 0.2 and 0.1
Background: https://github.com/actix/actix-web/issues/1593

This incompatibility causes a crash in autoendpoint after running for a few minutes, when handling a request.

Due to this incompatibility, we must copy and update whatever code `autoendpoint` uses from `autopush_common` into Tokio 0.2 (futures 0.3) compatible code. Until Actix fixes the runtime issue or we upgrade the rest of the autopush codebase to Tokio 0.2, autoendpoint cannot use any Tokio 0.1 code.

Mcat12:
Major changes:
- Autoendpoint no longer uses any Tokio 0.1 code (and consequently doesn't crash). The autopush_common database code has been copied and converted into Tokio 0.2 (futures 0.3) code.
- The autoendpoint `message_table_name` setting is now the exact name of the message table to use (previously it was a prefix, used in legacy table rotation code).

Minor changes:
- A `database.retry` metric is incremented whenever a database operation is retried.
- Some of the autopush_common macros are exported so they can be used in the autoendpoint database client.
- The autoendpoint database client returns an option for get_user instead of an error if it doesn't exist.

Closes #172

### DIFF
--- a/autoendpoint/src/db/client.rs
+++ b/autoendpoint/src/db/client.rs
@@ -1,0 +1,184 @@
+use crate::db::error::{DbError, DbResult};
+use crate::db::retry::{
+    retry_policy, retryable_delete_error, retryable_getitem_error, retryable_putitem_error,
+    retryable_updateitem_error,
+};
+use autopush_common::db::{DynamoDbNotification, DynamoDbUser};
+use autopush_common::notification::Notification;
+use autopush_common::{ddb_item, hashmap, val};
+use cadence::StatsdClient;
+use rusoto_core::credential::StaticProvider;
+use rusoto_core::{HttpClient, Region};
+use rusoto_dynamodb::{
+    AttributeValue, DeleteItemInput, DynamoDb, DynamoDbClient, GetItemInput, PutItemInput,
+    UpdateItemInput,
+};
+use std::collections::HashSet;
+use std::env;
+use uuid::Uuid;
+
+/// Provides high-level operations over the DynamoDB database
+#[derive(Clone)]
+pub struct DbClient {
+    ddb: DynamoDbClient,
+    metrics: StatsdClient,
+    router_table: String,
+    pub message_table: String,
+}
+
+impl DbClient {
+    pub fn new(
+        metrics: StatsdClient,
+        router_table: String,
+        message_table: String,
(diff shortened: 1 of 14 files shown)


autoendpoint/src/db/retry.rs
```diff
@@ -0,0 +1,34 @@
...
+use std::time::Duration;
+
+/// Create a retry function for the given error
+macro_rules! retryable_error {
+    ($name:ident, $error:tt, $error_tag:expr) => {
+        pub fn $name(metrics: StatsdClient) -> impl Fn(&RusotoError<$error>) -> bool {
+            move |err| match err {
+                RusotoError::Service($error::InternalServerError(_))
+                | RusotoError::Service($error::ProvisionedThroughputExceeded(_)) => {
+                    metrics
```
jrconlin, line 14: Nice. adding the metric call here ensures we don't miss recording this.

autoendpoint/src/db/client.rs
```diff
@@ -0,0 +1,184 @@
...
+            .await?;
+
+        Ok(())
+    }
+
+    /// Store a single message
+    pub async fn store_message(&self, uaid: Uuid, message: Notification) -> DbResult<()> {
+        let put_item = PutItemInput {
+            item: serde_dynamodb::to_hashmap(&DynamoDbNotification::from_notif(&uaid, message))
+                .unwrap(),
```
jrconlin, line 170: Should we have this return a proper error rather than a panic? (In case someone gives us a TTL of `None` or something stupid?) Could probably have that return a 400 in those cases, which might help the abuse filters.

Mcat12: > Serialization can fail if T's implementation of Serialize decides to fail, or if T contains a map with non-string keys.

https://docs.rs/serde_dynamodb/0.5.1/serde_dynamodb/fn.to_hashmap.html

I don't think either of these cases can occur here, but in future PRs I did change `DbError` to support serialization in addition to deserialization, so I will backport that change and handle the error here.

PR #182, opened 2020-07-10
refactor: Flatten autoendpoint server module

### LINKED ISSUE #157: [autoendpoint] Flatten module structure a bit
This is a real small issue, but I'm making an issue here to track it. Additionally, if we use the autoendpoint crate as a template in the future, we want it to be well architected.

The `server` module is a little too big. It should only handle setting up the Actix server. The extractors, routes, business logic, etc should be contained in "peer" modules.

This change should be done once there are no "major" PRs open for autoendpoint (perhaps after #154).

Current layout:
```
├── server
│   ├── extractors
│   │   ├── mod.rs
│   │   ├── notification.rs
│   │   ├── notification_headers.rs
│   │   ├── subscription.rs
│   │   └── token_info.rs
│   ├── headers
│   │   ├── crypto_key.rs
│   │   ├── mod.rs
│   │   ├── util.rs
│   │   └── vapid.rs
│   ├── mod.rs
│   └── routes
│       ├── health.rs
│       ├── mod.rs
│       └── webpush.rs
```

Proposed layout:
```
├── extractors
│   ├── mod.rs
│   ├── notification.rs
│   ├── notification_headers.rs
│   ├── subscription.rs
│   └── token_info.rs
├── headers
│   ├── crypto_key.rs
│   ├── mod.rs
│   ├── util.rs
│   └── vapid.rs
├── routes
│   ├── health.rs
│   ├── mod.rs
│   └── webpush.rs
└── server.rs
```

Mcat12:
This PR will likely require some merge conflict resolution and rebasing, but I wanted to get it out so we fix this structure issue sooner rather than later.

Closes #157


autoendpoint/src/extractors/notification.rs
```diff
@@ -1,6 +1,6 @@
 use crate::error::{ApiError, ApiErrorKind};
-use crate::server::extractors::notification_headers::NotificationHeaders;
-use crate::server::extractors::subscription::Subscription;
+use crate::extractors::notification_headers::NotificationHeaders;
+use crate::extractors::subscription::Subscription;
```
jrconlin, line 3: optional, could probably collapse these:
```
use crate::extractors::{
    notification_headers:::NotificationHeaders,
    subscription::Subscription,
};
```
if you do, just need to be consistent with other, similar includes. Could save a bit of typing for any future moves like this.

autoendpoint/src/error.rs
```diff
@@ -1,6 +1,7 @@
 //! Error types and transformations
 
-use crate::server::{RouterError, VapidError};
+use crate::headers::vapid::VapidError;
+use crate::routers::RouterError;
```
jrconlin, line 4: Probably also a nit, but we've previous used a python approach to library sections:

so includes are broken into groups of

* System/internal (e.g. `std`)
* External libraries (e.g. `actix::*`, usually in alphabetical order)
* Package internal (e.g. `crate::`) 

It's optional, and makes zero difference to the compiler or linker.

Mcat12: I like the idea, but I've found that it doesn't work well with auto-importing. At least for IntelliJ Rust, the import will often be placed in the wrong block which messes up the groups and requires manual intervention.

Additionally, rustfmt will already order the imports nicely (crate first, then alphabetically). Leaving it up to rustfmt seems more productive to me, and the CI can check it.

autoendpoint/src/extractors/notification.rs
Mcat12: We can make rustfmt handle this automatically:
https://rust-lang.github.io/rustfmt/?version=master&search=#merge_imports

jrconlin: Yeah, it's the clippy battle. I think clippy really wanted folks to collapse when possible, probably because of tons of one line `use`s.

autoendpoint/src/error.rs
jrconlin: Ah, it makes sense if you're using the auto-import feature. That's why these are nits more than anything.

autoendpoint/src/extractors/notification.rs
Mcat12: Oh, is Clippy complaining about this? I don't see any relevant lints in https://rust-lang.github.io/rust-clippy/stable/index.html

PR #184, opened 2020-07-13
fix: Use errnos from validation errors

Mcat12:
Validation errors contain error codes, so we should use those numbers in our error response. Example:
https://github.com/mozilla-services/autopush-rs/blob/e73dff17e9a9909743fd57bbd9d6769789a455e1/autoendpoint/src/server/extractors/notification_headers.rs#L21-L23

Other minor changes:
- Removed the unnecessary `Uuid` error, as it is an impossible case (the error is only thrown if the input is not 16 bytes long).


autoendpoint/src/server/extractors/subscription.rs
```diff
@@ -58,8 +58,8 @@ impl FromRequest for Subscription {
             }
 
             // Load and validate user data
-            let uaid = Uuid::from_slice(&token[..16])?;
-            let channel_id = Uuid::from_slice(&token[16..32])?;
+            let uaid = Uuid::from_slice(&token[..16]).unwrap();
```
jrconlin, line 61: I'll bite, why `unwrap()` rather than handle the Uuid error?

Mcat12: An error will only be returned if the slice length is not 16:
https://docs.rs/uuid/0.8.1/uuid/struct.Uuid.html#method.from_slice

I can add a clarifying comment.

PR #185, opened 2020-07-15
feat: User registration

### LINKED ISSUE #176: [autoendpoint] Support user registration endpoint
Endpoint: `POST /v1/{type}/{app_id}/registration`
Documentation: https://autopush.readthedocs.io/en/latest/http.html#registration

Registers a user with a bridge router.

**Questions:**
- The Python code appears to support registering WebPush clients via this endpoint. Should this still be supported, since they should instead be registered when connecting via WebSocket to the connection server?

Mcat12:
Major changes:
- Support registering users with routers. Currently only FCM is implemented (and webpush, but that should be happening via the connection servers).

Minor changes:
- Read the payload via the `Bytes`/`Json` extractors to simplify code and to use their built-in size limiting functionality.
- Return UUIDs as lowercase hyphenated strings (coincidentally the default `Display` impl for `Uuid`).

Closes #176

### DIFF
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -23,6 +23,7 @@ pub struct Settings {
 
     pub max_data_bytes: usize,
     pub crypto_keys: String,
+    pub auth_keys: String,
     pub human_logs: bool,
 
     pub statsd_host: Option<String>,
@@ -43,6 +44,7 @@ impl Default for Settings {
             message_table_name: "message".to_string(),
             max_data_bytes: 4096,
             crypto_keys: format!("[{}]", Fernet::generate_key()),
+            auth_keys: "[AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB=]".to_string(),
             human_logs: false,
             statsd_host: None,
             statsd_port: 8125,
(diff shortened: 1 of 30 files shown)


autoendpoint/src/settings.rs
```diff
@@ -85,17 +87,30 @@ impl Settings {
         })
     }
 
-    /// Initialize the fernet encryption instance
-    pub fn make_fernet(&self) -> MultiFernet {
-        if !(self.crypto_keys.starts_with('[') && self.crypto_keys.ends_with(']')) {
-            panic!("Invalid AUTOEND_CRYPTO_KEY");
+    /// Convert a string like `[item1,item2]` into a iterator over `item1` and `item2`.
```
jrconlin, line 90: technically, it's a JSON list, but I agree that pulling JSON into this is probably both overkill and a potential security issue.

autopush-common/src/endpoint.rs
```diff
@@ -0,0 +1,37 @@
...
+    endpoint_url: &str,
+    fernet: &MultiFernet,
+) -> Result<String> {
+    let root = Url::parse(endpoint_url)
+        .and_then(|root| root.join("/wpush/"))
+        .chain_err(|| "Invalid endpoint_url")?;
+    let mut base = hex::decode(uaid.to_simple().to_string()).chain_err(|| "Error decoding")?;
+    base.extend(hex::decode(chid.to_simple().to_string()).chain_err(|| "Error decoding")?);
+
+    if let Some(k) = key {
```
jrconlin, line 25: So, part of me wonders if we should create a NONCE key when there's none specified. This would ensure that the new URLs have a consistent length and be a little less subject to potential scrutiny.

Not saying we should do that here, but just thinking about it a bit more with the new system we're putting out.

autopush-common/src/endpoint.rs
```diff
@@ -0,0 +1,37 @@
...
+use uuid::Uuid;
+
+/// Create an v1 or v2 WebPush endpoint from the identifiers
+///
+/// Both endpoints use bytes instead of hex to reduce ID length.
+//  v1 is the uaid + chid
+//  v2 is the uaid + chid + sha256(key).bytes
+pub fn make_endpoint(
+    uaid: &Uuid,
+    chid: &Uuid,
```
jrconlin, line 14: As noted in https://github.com/mozilla-services/autopush-rs/pull/189, we want to make sure that we don't accidentally modify the channelid that's provided to us from the Client, since they treat them as strings. We may want to alter this to do the same.

autoendpoint/src/settings.rs
Mcat12: Unfortunately it's not a JSON list because the items aren't quoted.

autopush-common/src/endpoint.rs
Mcat12: v2 endpoints require VAPID, at minimum to validate the public key hash in the endpoint URL. This would require changes to the clients who don't currently provide a key, because they probably also don't send VAPID.

autopush-common/src/endpoint.rs
Mcat12: We need to have it as a `Uuid` here because we use the raw UUID bytes to compute the endpoint URL. Having it in hyphenated form (if that's what the client gave us) would produce an invalid URL.
(side note: we could just call `Uuid::as_bytes` instead of formatting as a simple UUID and then decoding the hex...)

autopush-common/src/endpoint.rs
jrconlin: Right, but we'd use the nonce for `v1` endpoints that don't include the key. That shouldn't require any modification of the client.

autopush-common/src/endpoint.rs
jrconlin: Yeah, we decode because it's far more efficient. At least we have an answer from client that they want lowercase, dashed so we can default to that when we compose messages out.

PR #188, opened 2020-07-17
feat: Update token endpoint

### LINKED ISSUE #177: [autoendpoint] Support token update endpoint
Endpoint: `PUT /v1/{type}/{app_id}/registration/{uaid}`
Documentation: https://autopush.readthedocs.io/en/latest/http.html#token-updates

Updates the user's bridge token.

Mcat12:
Adds the `PUT /v1/{router_type}/{app_id}/registration/{uaid}` endpoint.

Closes #177

### DIFF
--- a/autoendpoint/src/server/extractors/authorization_check.rs
+++ b/autoendpoint/src/server/extractors/authorization_check.rs
@@ -61,7 +61,7 @@ fn get_token_from_auth_header(header: &str) -> Option<&str> {
     let mut split = header.splitn(2, ' ');
     let scheme = split.next()?;
 
-    if scheme != "Bearer" {
+    if scheme.to_lowercase() != "bearer" {
         return None;
     }
 
--- a/autoendpoint/src/server/extractors/authorization_check.rs
+++ b/autoendpoint/src/server/extractors/authorization_check.rs
@@ -7,6 +7,7 @@ use actix_web::web::Data;
 use actix_web::{FromRequest, HttpRequest};
 use futures::future::LocalBoxFuture;
 use futures::FutureExt;
+use uuid::Uuid;
 
 /// Verifies the request authorization via the authorization header.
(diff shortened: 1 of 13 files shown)


autoendpoint/src/server/extractors/authorization_check.rs
```diff
@@ -0,0 +1,69 @@
...
+        .boxed_local()
+    }
+}
+
+/// Get the token from a bearer authorization header
+fn get_token_from_auth_header(header: &str) -> Option<&str> {
+    let mut split = header.splitn(2, ' ');
+    let scheme = split.next()?;
+
+    if scheme.to_lowercase() != "bearer" {
```
Mcat12, line 64: The python code re-uses `AUTH_SCHEMES` when checking this header, but that allows schemes such as `webpush` and `vapid`. I've restricted it to just `Bearer`, but do we need to support those other schemes?

jrconlin: those are older auth schema specifications. I believe that [RFC 8292](https://tools.ietf.org/html/rfc8292#section-2) requires schema "vapid", [prior to that](https://tools.ietf.org/html/draft-ietf-webpush-vapid-01#section-2) it was "webpush". "bearer" was the oldest form. We may want to either continue accepting the old forms or just limit to the RFC 8292 spec and return an appropriate error that no one will probably read.

Mcat12: This isn't VAPID, this is our own authentication mechanism where the key is the hashed UAID. That shouldn't be covered under any RFC?

jrconlin: Ah, right. this is just the simple auth check we run. `Bearer` is fine because that's exactly what it is.

PR #189, opened 2020-07-17
feat: New channel endpoint

### LINKED ISSUE #178: [autoendpoint] Support create channel subscription endpoint
Endpoint: `POST /v1/{router_type}/{app_id}/registration/{uaid}/subscription`
Documentation: https://autopush.readthedocs.io/en/latest/http.html#channel-subscription

Creates a new channel ID for the user.

Mcat12:
Adds the `POST /v1/{router_type}/{app_id}/registration/{uaid}/subscription` endpoint.
Also fixed a warn log statement which produced redundant output.

Closes #178

### DIFF
--- a/autoendpoint/src/extractors/mod.rs
+++ b/autoendpoint/src/extractors/mod.rs
@@ -3,6 +3,7 @@
 
 pub mod authorization_check;
 pub mod message_id;
+pub mod new_channel_data;
 pub mod notification;
 pub mod notification_headers;
 pub mod registration_path_args;
--- a/autoendpoint/src/extractors/new_channel_data.rs
+++ b/autoendpoint/src/extractors/new_channel_data.rs
@@ -1,0 +1,10 @@
+use uuid::Uuid;
+
+/// The data provided when creating a new channel for an existing user. Extract
+/// from the request via the `Json` extractor.
+#[derive(serde::Deserialize, Default)]
+pub struct NewChannelData {
+    #[serde(rename = "channelID")]
+    pub channel_id: Option<Uuid>,
+    pub key: Option<String>,
+}
--- a/autoendpoint/src/routers/fcm/router.rs
+++ b/autoendpoint/src/routers/fcm/router.rs
@@ -119,7 +119,7 @@ impl FcmRouter {
                 }
             }
             FcmError::FcmUpstream { .. } | FcmError::FcmUnknown => {
-                warn!("FCM error: {error}", error = error.to_string());
+                warn!("{}", error.to_string());
                 self.incr_error_metric("server_error");
             }
             _ => {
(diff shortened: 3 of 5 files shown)


autoendpoint/src/server/routes/registration.rs
```diff
@@ -108,6 +109,38 @@ pub async fn update_token_route(
...
+pub async fn new_channel_route(
+    _auth: AuthorizationCheck,
+    path_args: RegistrationPathArgsWithUaid,
+    channel_data: Option<Json<NewChannelData>>,
+    state: Data<ServerState>,
+) -> ApiResult<HttpResponse> {
+    // Add the channel
+    debug!("Adding a channel to UAID {}", path_args.uaid);
+    let channel_data = channel_data.map(Json::into_inner).unwrap_or_default();
+    let channel_id = channel_data.channel_id.unwrap_or_else(Uuid::new_v4);
```
jrconlin, line 122: It's really important that we don't modify how we get the channelid. The client stores and handles this as a string, and if we change the format, the matches fail. 

This does mean that we should probably just store the channelid as a string with appropriate comments.

autoendpoint/src/server/extractors/new_channel_data.rs
```diff
@@ -0,0 +1,10 @@
+use uuid::Uuid;
+
+/// The data provided when creating a new channel for an existing user. Extract
+/// from the request via the `Json` extractor.
+#[derive(serde::Deserialize, Default)]
+pub struct NewChannelData {
+    #[serde(rename = "channelID")]
+    pub channel_id: Option<Uuid>,
```
jrconlin, line 8: See note later about why we may want to store this as a string instead of a UUID.

autoendpoint/src/server/routes/registration.rs
Mcat12: Can we fix the client so it parses the UUID as a UUID? Looking at the Python FCM code, it always sends the channel ID in the message as a simple hex string, so I'm confused where these "matches" you mention live?

fzzzy: I think these matches are in Sync. When a sync client does a PUT it can choose the format of the uuid (unfortunately). We could normalize the uuid representation in the backend, but then when we go to generate urls we can't generate the uuid format the client is expecting.

This is only a problem because sync is an open protocol and people write their own clients.

I'm not sure if this same problem exists in push. I don't think push has to generate urls with uuids in them to give back to the client, but I'm not positive.

Mcat12: This is fixed in the latest commit (return lowercase hyphenated UUIDs to match the client's expected format). Additionally, the UUIDs returned during user registration (#185) were adjusted.

PR #197, opened 2020-07-23
fix: Drop unknown FCM users

### LINKED ISSUE #173: Drop user routing information if bridge reports 404/410 or similar error
If the bridge system reports a user is no longer registered or found, we should drop that user from our routing table. 

Companion to:

https://github.com/mozilla-services/autopush/issues/1414

Mcat12:
Major changes:
- If FCM says the user does not exist (404, [docs]), then we drop the user from the router table.
- `DbClient` is now abstracted via a trait and can be mocked with `MockDbClient`. [Mocking library documentation][mockall].

Closes #173 

[docs]: https://firebase.google.com/docs/reference/fcm/rest/v1/ErrorCode
[mockall]: https://docs.rs/mockall/0.7.1/mockall

### DIFF
--- a/autoendpoint/src/db/client.rs
+++ b/autoendpoint/src/db/client.rs
@@ -1,7 +1,7 @@
 use crate::db::error::{DbError, DbResult};
 use crate::db::retry::{
-    retry_policy, retryable_delete_error, retryable_getitem_error, retryable_putitem_error,
-    retryable_updateitem_error,
+    retry_policy, retryable_delete_error, retryable_describe_table_error, retryable_getitem_error,
+    retryable_putitem_error, retryable_updateitem_error,
 };
 use async_trait::async_trait;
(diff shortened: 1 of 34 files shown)


jrconlin:
Heh, kinda burying the lede here that this also wraps in DbClient mocking.

autoendpoint/src/db/client.rs
```diff
@@ -95,6 +104,32 @@ impl DbClientImpl {
...
+            }
+            Err(e) => return Err(e.into()),
+        };
+
+        let status = output
+            .table
+            .and_then(|table| table.table_status)
+            .ok_or(DbError::TableStatusUnknown)?;
+
+        Ok(["CREATING", "UPDATING", "ACTIVE"].contains(&status.as_str()))
```
jrconlin, line 131: nit: I know that this will probably always be upper case, but I wonder if it might be good to force status to upper case to avoid any future issues.

PR #200, opened 2020-07-23
fix: Strip padding and double quotes from encryption and crypto-key headers

### LINKED ISSUE #192: Clean up padding and quotes in headers
This duplicates [an issue for autopush](https://github.com/mozilla-services/autopush/issues/1419)

Client reports issues with quotes and padding appearing in values passed from Encryption and Crypto-Key headers. 

These are legacy "aesgcm" encryption formats that are still in use. Server should probably clean these values up (remove quotes and padding) before relaying.

Client may also take steps to resolve this issue. 

image

Mcat12:
The regex is a little more complicated than the Python version because Rust's regex library does not support look-ahead, for performance reasons.

Closes #192

### DIFF
--- a/autoendpoint/src/extractors/notification_headers.rs
+++ b/autoendpoint/src/extractors/notification_headers.rs
@@ -12,6 +12,8 @@ use validator_derive::Validate;
 
 lazy_static! {
     static ref VALID_BASE64_URL: Regex = Regex::new(r"^[0-9A-Za-z\-_]+=*$").unwrap();
+    static ref STRIP_PADDING: Regex =
+        Regex::new(r"(?P<head>[0-9A-Za-z\-_]+)=+(?P<tail>[,;]|$)").unwrap();
 }
 
 const MAX_TTL: i64 = 60 * 60 * 24 * 60;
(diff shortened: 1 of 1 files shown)


autoendpoint/src/extractors/notification_headers.rs
```diff
@@ -411,5 +423,31 @@ mod tests {
...
 
+    /// The encryption and crypto-key headers are stripped of Base64 padding and
+    /// double-quotes.
+    #[test]
+    fn strip_headers() {
+        let req = TestRequest::post()
+            .header("TTL", "10")
+            .header("Content-Encoding", "aesgcm")
+            .header("Encryption", "salt=\"foo\"")
+            .header("Crypto-Key", "dh=\"deadbeef==\"")
```
jrconlin, line 434: I don't think you'll have a problem here, but could you add the subcomponent to this?

(e.g.  `.header("Crypto-Key", "keyid=\"p256dh\";dh=\"deadbeef==\""_)` )

I'd like to make sure that future fixes break if they can't parse that too.

PR #201, opened 2020-07-24
feat: APNS Router

### LINKED ISSUE #164: [autoendpoint] Route APNS notifications
WebPush notifications (destined for iOS user agents) should be routed through Apple Push Notification Service.

Mcat12:
This adds a router for Apple Push Notification Services. The router uses the [`a2`] library to send notifications to APNS. Some small changes were necessary, so we're currently using a [fork], but all of the changes have been submitted to the upstream repo as PRs.

Closes #164 

[`a2`]: https://github.com/pimeys/a2
[fork]: https://github.com/Mcat12/a2


autoendpoint/src/routers/apns/router.rs
```diff
@@ -0,0 +1,536 @@
...
+
+    /// Update metrics after successfully routing the notification
+    fn incr_success_metrics(&self, notification: &Notification, channel: &str) {
+        self.metrics
+            .incr_with_tags("notification.bridge.sent")
+            .with_tag("platform", "apns")
+            .with_tag("application", channel)
+            .send();
+        self.metrics
+            .incr_with_tags(&format!("updates.client.bridge.apns.{}.sent", channel))
```
jrconlin, line 115: Lets drop this second metric. It's redundant.

autoendpoint/src/routers/apns/router.rs
```diff
@@ -0,0 +1,536 @@
...
+                .into_iter()
+                .map(|(k, v)| (k, Value::String(v)))
+                .collect(),
+            device_token: token,
+            options: NotificationOptions {
+                apns_id: None,
+                apns_priority: Some(Priority::High),
+                apns_topic: Some(topic),
+                apns_collapse_id: None,
+                apns_expiration: None,
```
jrconlin, line 243: hrm, this should probably be the message TTL. We don't want messages appearing after they've expired. (I'll file a separate issue for that since the python never got updated either.)

jrconlin: https://github.com/mozilla-services/autopush-rs/issues/202

autoendpoint/src/routers/apns/router.rs
```diff
@@ -0,0 +1,536 @@
...
+            .unwrap_or_else(Self::default_aps);
+        let mut message_data = build_message_data(notification, self.settings.max_data)?;
+        message_data.insert("ver", notification.message_id.clone());
+
+        // Get client and build payload
+        let ApnsClientData { client, topic } = self
+            .clients
+            .get(channel)
+            .ok_or(ApnsError::InvalidReleaseChannel)?;
+        let payload = Payload {
```
jrconlin, line 231: I know that we're checking subscription data size with the app, but each bridge also has a data limit which is smaller than the general message size (this is because we may need to re-encode the message being sent through, which can reduce the effective total message size.)

I don't see where we're checking against that, except if the service rejects the message, which could result in user confusion. Am I missing something?

autoendpoint/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Mark Drobnak <mdrobnak@mozilla.com>", "jrconlin <me+crypt@jrconlin.c
 edition = "2018"
 
 [dependencies]
+a2 = { git = "https://github.com/Mcat12/a2.git", branch = "autoendpoint" }
```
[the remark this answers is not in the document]
Mcat12: These are the three PRs:
- https://github.com/pimeys/a2/pull/49
- https://github.com/pimeys/a2/pull/48
- https://github.com/pimeys/a2/pull/47

autoendpoint/src/routers/apns/router.rs
Mcat12: No, you're right we should be checking that here.

autoendpoint/src/routers/apns/router.rs
Mcat12: This is easily fixed in this PR. Can we move the issue to the Python repo instead?

autoendpoint/src/routers/apns/router.rs
Mcat12: It looks like the Python code does not do any size checks for APNS or ADM, and only checks the notification `data` payload in the FCM router:
https://github.com/mozilla-services/autopush/blob/91db1fa6c564af84d7dbbdb4e20e0619579edf0b/autopush/router/fcm_v1.py#L85

I'll fix this issue for the APNS router and make an issue to fix the FCM router (Rust). Should we also go back and fix the Python routers?

Mcat12: #203

autoendpoint/Cargo.toml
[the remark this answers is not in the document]
Mcat12: The `autoendpoint` branch is just those three branches merged together.

Mcat12:
Converting to draft because this still needs to implement dropping users when they're unregistered. Testing that functionality relies on #197.

autoendpoint/src/routers/apns/router.rs
jrconlin: Probably not a bad idea to fix up the python router too. https://github.com/mozilla-services/autopush/issues/1422

Mcat12:
This PR should be ready to review now.

autoendpoint/Cargo.toml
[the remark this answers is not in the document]
Mcat12: We can move the fork into the `mozilla-services` namespace if that would help things.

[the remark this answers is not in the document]
jrconlin: It's more a question of how fast he lands patches like that and how well we can monitor or fix things ourselves. Chances are pretty good that both aren't a problem, but we've had instances in the past where it's not the case. A really good, current example of that is the mozilla-sdk. 

If we're tied to your repo and forget that this is the case, we could miss out on future security patches. I'd be fine adding a commented out line to the Cargo.toml that points to the proper repo, a comment, and maybe with a version one tick above current release so it's a bit more obvious what's going on.

autoendpoint/src/routers/apns/router.rs
jrconlin: Sure. done.

autoendpoint/Cargo.toml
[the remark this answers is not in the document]
Mcat12: I've added these details to Cargo.toml in the latest commit.

Issue #10, opened 2018-06-21
Create example configuration files with all settings.


oremj:
There should be example config files in this repo with all settings and what they do.

Mcat12:
Fixed by #214 and #216

PR #214, opened 2020-08-10
docs: Add a sample config for autoendpoint and normalize router settings

Mcat12:
This might be useful when we start deploying autoendpoint-rs.

Related to #10

### DIFF
--- a/configs/autoendpoint.toml.sample
+++ b/configs/autoendpoint.toml.sample
@@ -1,0 +1,107 @@
+# The URI scheme to use when referencing this server
+#scheme = "http"
+
+# The host to use
+#host = "127.0.0.1"
+
+# The port to use
+#port = 8000
+
+# The router table name
+#router_table_name = "router"
+
+# The message table name
+#message_table_name = "message"
+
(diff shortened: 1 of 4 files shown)


configs/autoendpoint.toml.sample
```diff
@@ -0,0 +1,107 @@
...
+# The router table name
+#router_table_name = "router"
+
+# The message table name
+#message_table_name = "message"
+
+# The maximum payload size to accept in HTTP requests to this server
+#max_data_bytes = 4096
+
+# The fernet keys to use. Multiple are allowed when separated by a comma.
```
jrconlin, line 19: nit: in the old python version, we had a note on how to generate this key. Many of the potential users of this package are not as savvy about how to do this. 

I wonder if it makes sense to at least include the old `autokey.py` script in the `/tools` directory. Probably should be a separate issue.

jrconlin: Added: https://github.com/mozilla-services/autopush-rs/issues/217

PR #216, opened 2020-08-11
docs: Add a sample config for autopush and fix some settings

Mcat12:
- Removed unused `debug` setting
- Added default for `endpoint_hostname`
- Fixed invalid default for `crypto_key`

Related to #10

### DIFF
--- a/autopush/src/settings.rs
+++ b/autopush/src/settings.rs
@@ -27,7 +27,6 @@ fn include_port(scheme: &str, port: u16) -> bool {
 
 #[derive(Debug, Default, Deserialize)]
 pub struct Settings {
-    pub debug: bool,
     pub port: u16,
     pub hostname: Option<String>,
     pub resolve_hostname: bool,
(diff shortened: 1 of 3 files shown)


autopush/src/settings.rs
```diff
@@ -73,8 +72,9 @@ impl Settings {
         s.set_default("max_connections", 0)?;
         s.set_default("close_handshake_timeout", 0)?;
         s.set_default("endpoint_scheme", "http")?;
+        s.set_default("endpoint_hostname", "localhost")?;
         s.set_default("endpoint_port", 8082)?;
-        s.set_default("crypto_key", Fernet::generate_key())?;
```
jrconlin, line 77: I know that we're switching to a CSV list for this value.
Don't we reject any string that has "[]" as a wrapper in `ServerOptions::from_settings`?

configs/autopush.toml.sample
```diff
@@ -0,0 +1,83 @@
...
+
+# The name of the router table
+#router_tablename = "router"
+
+# The prefix of the message table(s)
+#message_tablename = "message"
+
+# A (stringified) list of Fernet keys to use when encrypting the notification
+# endpoint URL. The default is a single auto-generated key.
+#crypto_key = "[]"
```
jrconlin, line 66: remove the brackets, or use some pre-generated test value (noting that it should not be used in production in the comment?)

autopush/src/settings.rs
Mcat12: We reject the opposite: if the string does not have a `[]` wrapper.

jrconlin: Ah! you're right. I missed the leading `!`.

PR #220, opened 2020-08-12
feat: Use CircleCI instead of Travis for tests

### LINKED ISSUE #148: Switch integration tests to circle-ci

Mcat12:
Travis is no longer used. The tests (including format check) run in parallel alongside the build and audit jobs. Currently, the tests take about 3-7 minutes (depending on cache availability), compared to 22 minutes in Travis.

Some of the changes:
- We're now using a DynamoDB image (`circleci/dynamodb`) instead of starting one inside of the tests. Support for using an existing database has been added to the integration test suites (set `AWS_LOCAL_DYNAMODB`).
- Because the database now persists between the two integration test suites, some issues related to message table rotation had to be fixed. Because of this and the fact that we don't do message table rotation anymore, it has been disabled during the tests and the associated tests are removed. More info in https://github.com/mozilla-services/autopush-rs/commit/cf2167c3bcb9e32d207154108d86f4f528e330c1.
- Due to a different Python environment setup (no automatic virtualenv), git dependencies are no longer installed as editable. See also https://github.com/mozilla-services/autopush/pull/1431.

This also requires a change to the GitHub repo to change from requiring Travis checks to CircleCI.

Closes #148


.circleci/config.yml
```diff
@@ -19,6 +19,68 @@ jobs:
...
+      - restore_cache:
+          name: Restoring Rust cache
+          key: rust-v1-{{ .Branch }}-{{ checksum "Cargo.lock" }}
+      - run:
+          name: Set up Python
+          command: |
+            pip install --upgrade pip
+            pip install bottle
+            pip install -r requirements.txt
+            pip install -r test-requirements.txt
```
jrconlin, line 51: IIRC, `test-requirements.txt` begins with a `-r requirements.txt`. You should be able to just bring it in and it'll grab both files.

.circleci/config.yml
```diff
@@ -144,6 +209,7 @@ workflows:
           repo: ${DOCKERHUB_ENDPOINT_REPO}
           requires:
             - build-autoendpoint
+            - test
```
Mcat12, line 212: Now that the autoendpoint CI changes are in, `test` should be a dependency of `deploy-autopush` as well.

Issue #233, opened 2020-11-13
Address RUSTSEC-2020-0059


tublitzed:
Looks like the latest [build failure](https://app.circleci.com/pipelines/github/mozilla-services/autopush-rs) is caused by [RUSTSEC-2020-0059](https://rustsec.org/advisories/RUSTSEC-2020-0059) and should be fixable by upgrading futures-util.

pjenvey:
Closed in #235

PR #235, opened 2020-11-16
Bug/234

### LINKED ISSUE #234: Autopush should accept JSON formatted env args
Some args are compound and are normally expressed as JSON.  (e.g. AUTOEND_CRYPTO_KEYS, fcm creds, etc.)
Ops already has these as JSON, so we should make sure they're accepted.

jrconlin:
## Description
AUTH_KEYS were being rejected due to invalid characters. This was due to the fact that the while the sample config was not JSON Formatted, the ops key was. 

This is a stupid fix to ensure that keys are accepted.

## Testing

Tests included.

Closes #234, #233

### DIFF
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -45,7 +45,7 @@ impl Default for Settings {
             message_table_name: "message".to_string(),
             max_data_bytes: 4096,
             crypto_keys: format!("[{}]", Fernet::generate_key()),
-            auth_keys: "[AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB=]".to_string(),
+            auth_keys: r#"["AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB="]"#.to_string(),
             human_logs: false,
             statsd_host: None,
             statsd_port: 8125,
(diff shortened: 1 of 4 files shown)


tublitzed:
Hey, JR I _think_ after poking around a bit last week that we might need to fix [this one](https://github.com/mozilla-services/autopush-rs/issues/233) before CI passes.

jrconlin:
Ah! missed that one. Looks like I incorporated it into this PR.

autoendpoint/src/settings.rs
```diff
@@ -106,15 +106,19 @@ impl Settings {
 
     /// Initialize the fernet encryption instance
     pub fn make_fernet(&self) -> MultiFernet {
-        let fernets = Self::read_list_from_str(&self.crypto_keys, "Invalid AUTOEND_CRYPTO_KEYS")
+        let keys = &self.crypto_keys.replace('"', "").replace(" ", "");
+        let fernets = Self::read_list_from_str(&keys, "Invalid AUTOEND_CRYPTO_KEYS")
```
Mcat12, line 110: Definitely a simple fix, but just noting here that something like `serde_json::decode::<Vec<String>>` should work.

jrconlin: Yeah, I considered that, but the example we had was not properly formatted JSON. For now, I figured I should apply that fix until I can dig into seeing if we allowed similar values in Python.

Mcat12: Ah, ok so you want to allow the previous input as well :+1:

jrconlin: Heh, well, if this is going to be a drop-in replacement, it needs to follow the same rules we had before.

Not super happy about it, but it's something we can address later if need be.

jrconlin: Checked, and the autopush python [sample file](https://github.com/mozilla-services/autopush/blob/master/configs/autopush_endpoint.ini.sample#L28) also used non-JSON values.

pjenvey (approved):
It'd be nice to switch to strictly the JSON parsing later once ops switches over (and remove the extra parsing step from `auth_keys`), can we log an issue so we don't forget?

PR #245, opened 2020-12-15
chore: move a2 under mozilla-services

### LINKED ISSUE #236: Move forked "a2" dependency into mozilla-services org
Following best practices to not tie production code to our personal repos, let's move [this dep](https://github.com/mozilla-services/autopush-rs/blob/master/autoendpoint/Cargo.toml#L15) out of a personal repo and into the mozilla-services org before we deploy this to prod.

jrconlin:
There has been no action on pimeys/a2#49 since
Aug 12 2020. Using a moz controlled branch for now.

Closes #236

### DIFF
--- a/tests/test_integration_all_rust.py
+++ b/tests/test_integration_all_rust.py
@@ -965,7 +965,7 @@ class TestRustWebPush(unittest.TestCase):
     # Client.delete_notification to fail.
 
     # Skipping test for now.
-    # """
+    """
     @inlineCallbacks
     def test_delete_saved_notification(self):
         client = yield self.quick_register()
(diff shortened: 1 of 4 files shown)


tests/test_integration_all_rust.py
```diff
@@ -957,7 +957,6 @@ def test_empty_message_with_crypto_headers(self):
 
         yield self.shut_down(client)
 
-    """
```
pjenvey, line 960: Can this test be re-enabled now that we reverted the 202/201 change?

Probably better as a separate issue

pjenvey (approved):

tests/test_integration_all_rust.py
jrconlin: Honestly, it's another weird one. Locally, this passes fine, however on Circle it fails with the first `send_notification()` returning a 202. I have no clue why. 

I'll file an

PR #250, opened 2021-02-03
chore: Update dependencies where possible.

jrconlin:
* Pinned dependencies contain comments indicating why they're pinned.

### DIFF
--- a/autoendpoint/Cargo.toml
+++ b/autoendpoint/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 # The version of a2 at the time of the fork is v0.5.3.
 a2 = { git = "https://github.com/mozilla-services/a2.git", branch = "autoendpoint" }
 # several of these libraries are pinned due to https://github.com/mozilla-services/autopush-rs/issues/249
-actix-web = "3.2"  # 3.3+ requires futures 0.3+
+actix-web = "3.3"
 actix-rt = "1.1"   # 2.0+ requires futures 0.3+
 actix-cors = "0.5"
 again = { version = "0.1.2", default-features = false, features = ["log"] }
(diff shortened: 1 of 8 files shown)


autoendpoint/Cargo.toml
```diff
@@ -13,15 +13,16 @@ edition = "2018"
 # The `autoendpoint` branch merges these three PRs together.
 # The version of a2 at the time of the fork is v0.5.3.
 a2 = { git = "https://github.com/mozilla-services/a2.git", branch = "autoendpoint" }
-actix-web = "3.2"
-actix-rt = "1.1"
+# several of these libraries are pinned due to https://github.com/mozilla-services/autopush-rs/issues/249
```
Mcat12, line 16: Futures 0.3 is already used by autoendpoint, so these updates shouldn't necessarily be blocked:
https://github.com/mozilla-services/autopush-rs/blob/365cb29bf86f197f6cc8d42424d27808f41c71b8/Cargo.lock#L1411

jrconlin: From what i was seeing it's a compounding problem. Autoendpoint uses futures 0.3, but common doesn't always, and there are a bunch of changes rolled into the other dependencies that need to be made which we'd need to prioritize against what we've got upcoming. 

Basically, "Yay! More tech debt due to wear out bits!". 

Sigh.

autoendpoint/Cargo.toml
```diff
@@ -13,15 +13,16 @@ edition = "2018"
...
+actix-web = "3.2"  # 3.3+ requires futures 0.3+
```
pjenvey, line 17: 3.3 is still on the older tokio/futures, autoendpoint's already using it according to Cargo.lock

pjenvey (changes requested):

pjenvey (approved):

PR #260, opened 2021-03-10
bug Always return status 201 for success

### LINKED ISSUE #259: Return 201 for message acceptance
Follow up for https://github.com/mozilla-services/autopush/issues/1443, we should ensure that we return 201 for all message accepts.

jrconlin:
Closes #259


pjenvey (changes requested):
The tests also need an update

jrconlin:
Yep, copying over the Client to the integration tests to match the existing autopush one. Keeping things in different commits to keep things visually clearer for review.

pjenvey (approved):

Issue #266, opened 2021-03-18
Drop support for aesgcm128


jrconlin:
This is a very old version of the protocol. We no longer see messages in this format. We should block messages going to the client.

![image](https://user-images.githubusercontent.com/35755/111708708-9621a980-8803-11eb-9d71-1c687b878941.png)

PR #268, opened 2021-03-23
feat: Drop aesgcm128 support

jrconlin:
## Description

Drops support for legacy `aesgcm128` encryption format.  This is in parallel with https://github.com/mozilla/rust-ece/issues/53 which will drop support for the format for mobile devices.

We're also seeing zero messages containing this form, so it's absolutely safe to drop it.

## Testing

messages encoded with `aesgcm128` should be rejected.

## Issue

Closes: #266

### DIFF
--- a/autoendpoint/src/extractors/notification_headers.rs
+++ b/autoendpoint/src/extractors/notification_headers.rs
@@ -122,7 +122,6 @@ impl NotificationHeaders {
         })?;
 
         match encoding {
-            "aesgcm128" => self.validate_encryption_01_rules()?,
             "aesgcm" => self.validate_encryption_04_rules()?,
             "aes128gcm" => self.validate_encryption_06_rules()?,
             _ => {
@@ -132,16 +131,6 @@ impl NotificationHeaders {
                 .into());
             }
         }
-
-        Ok(())
-    }
-
-    /// Validates encryption headers according to
-    /// draft-ietf-webpush-encryption-01
-    fn validate_encryption_01_rules(&self) -> ApiResult<()> {
-        Self::assert_base64_item_exists("Encryption", self.encryption.as_deref(), "salt")?;
-        Self::assert_base64_item_exists("Encryption-Key", self.encryption_key.as_deref(), "dh")?;
-        Self::assert_not_exists("aesgcm128 Crypto-Key", self.crypto_key.as_deref(), "dh")?;
 
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


pjenvey:
> We're also seeing zero messages containing this form, so it's absolutely safe to drop it.

It's not quite zero but it's very low now, not even 1-2 requests per second

autoendpoint/src/extractors/notification_headers.rs
```diff
@@ -351,32 +340,7 @@ mod tests {
...
-                encoding: Some("aesgcm128".to_string()),
-                encryption: Some("salt=foo".to_string()),
-                encryption_key: Some("dh=bar".to_string()),
-                crypto_key: None
-            }
-        );
-    }
-
-    /// Valid 04 draft encryption passes validation
+    /// Valid 04 draft encryption passes validationgit o
```
pjenvey, line 343: nit: some cruft ended up at the end of this comment

pjenvey (changes requested):

autoendpoint/src/extractors/notification_headers.rs
jrconlin: crap. Thanks!

pjenvey:
aesgcm128 appears to be the sole user of the `Encryption-Key` header. We could also further remove it, killing the `encryption_key` field in here and related `enckey` entry in the router message data.

The only reference to it we _might_ need going forward is the [explicit invalidation if it was included](https://github.com/mozilla-services/autopush-rs/blob/2b2f6326dfc8482825d176c3ec97f1b74dc09915/autoendpoint/src/extractors/notification_headers.rs#L145) in the request. This error's only triggered for the `04` encoding but I don't see why it couldn't apply to `06` too, so the check could occur in `from_request` instead so the struct field wouldn't be needed.

Maybe we don't even need the check any longer either?

I'm happy to defer this till later but I think it's worth the effort even for such a minor cleanup -- otherwise this field's likely to linger around forever. The various bits of the different crypto standards are complicated enough as is.

pjenvey (approved):
aesgcm128 appears to be the sole user of the `Encryption-Key` header. We could also further remove it, killing the `encryption_key` field in here and related `enckey` entry in the router message data.

The only reference to it we _might_ need going forward is the [explicit invalidation if it was included](https://github.com/mozilla-services/autopush-rs/blob/2b2f6326dfc8482825d176c3ec97f1b74dc09915/autoendpoint/src/extractors/notification_headers.rs#L145) in the request. This error's only triggered for the `04` encoding but I don't see why it couldn't apply to `06` too, so the check could occur in `from_request` instead so the struct field wouldn't be needed.

Maybe we don't even need the check any longer either?

I'm happy to defer this till later but I think it's worth the effort even for such a minor cleanup -- otherwise this field's likely to linger around forever. The various bits of the different crypto standards are complicated enough as is.

PR #270, opened 2021-03-24
bug: Add explicit `endpoint_url` setting

### LINKED ISSUE #269: Explicity specify Endpoint_url as a setting instead of crafting it from the host
In deployment, we're seeing the host be set to `0.0.0.0`, which is a valid host IP address for a server. We should explicitly set the endpoint URL so that VAPID aud can match it regardless of local deployment configurations.

jrconlin:
## Description
This will add an explicit setting to specify the endpoint URL

###OPS
Note that this will include `AUTOEND_ENDPOINT_URL` as a new setting, the
URL must be a fully qualified string (e.g. `https://updates.push.services.mozilla.com/`)

## Testing

Testing included. Additional testing can be done by specifying an endpoint URL and validating using VAPID header `aud` fields.

## Issue

Closes #269

### DIFF
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -17,6 +17,7 @@ pub struct Settings {
     pub scheme: String,
     pub host: String,
     pub port: u16,
+    pub endpoint_url: String,
 
     pub router_table_name: String,
     pub message_table_name: String,
@@ -40,6 +41,7 @@ impl Default for Settings {
         Settings {
             scheme: "http".to_string(),
             host: "127.0.0.1".to_string(),
+            endpoint_url: "".to_string(),
             port: 8000,
             router_table_name: "router".to_string(),
             message_table_name: "message".to_string(),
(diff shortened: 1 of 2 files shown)


autoendpoint/src/settings.rs
```diff
@@ -17,6 +17,7 @@ pub struct Settings {
     pub scheme: String,
     pub host: String,
     pub port: u16,
+    pub endpoint_url: String,
```
pjenvey, line 20: minor nit: could also do `Option<String>` defaulting to `None`

pjenvey (approved):

autoendpoint/src/settings.rs
jrconlin: Yeah, I thought about that. I wasn't sure how well `Settings` would deal with that since I remember it being kind of a hassle, so I went the cheap route for now.

pjenvey: It works as you'd expect, see `statsd_host` in autoendpoint-rs  we use it elsewhere. No big deal though

PR #280, opened 2021-06-28
chore: library update for Jun 2021

jrconlin:


autopush-common/Cargo.toml
```diff
@@ -34,11 +34,11 @@ openssl = "0.10"
 rand = "0.8"
 regex = "1.4"
 # XXX: pin to < 0.10 until futures 0.3
-reqwest = "0.9.24"
+reqwest = "0.11"
```
AzureMarker, line 37: Looks like either the comment is outdated or the version shouldn't be upgraded?

jrconlin: Ah, thanks! Missed that comment, I'll make it a bit clearer for myself.

pjenvey (approved):

PR #281, opened 2021-07-23
Update to Cadence 0.26

56quarters:
Update to Cadence 0.26 which contains a breaking change and moves incr/decr methods to a new trait. Similar changes will have to be made in the various other services repos that use Cadence.

Signed-off-by: Nick Pillitteri <nick@56quarters.xyz>


jrconlin (approved):
Thanks!

pjenvey (approved):

PR #283, opened 2021-07-30
chore: Dep update for Jul 2021

jrconlin:


jrconlin:
Since I missed the prior update cycle.

autopush-common/Cargo.toml
```diff
@@ -24,7 +24,7 @@ futures = "0.1.29" # 0.1.30 requires hyper 0.13+
 futures-backoff = "0.1.0"
 hex = "0.4.2"
 httparse = "1.3"
-hyper = "^0.12.36" # 0.13+ requires too many changes
+hyper = "0.14"
```
pjenvey, line 27: sadly we still need this pinned to 0.12

pjenvey (changes requested):

autopush-common/Cargo.toml
jrconlin: Ah, I was hoping that since a bunch of the 0.13 changes had been regressed we could skip the version. Oh well, i'll update the comment too.

pjenvey (approved):

PR #284, opened 2021-08-14
 bug: fix local Authorization header

### LINKED ISSUE #282: Autopush-rs Endpoint fails to read `AUTHENTICATION` header
We're seeing https://sentry.prod.mozaws.net/operations/autopush-prod/issues/12011604/events/latest/ on test deploy of the autopush-rs endpoint.

The error being returned indicates that `get_token_from_auth_header` is failing, indicating that either the AUTHORIZATION header changed format or we're not reading it correctly. We need to test against `Mozilla/5.0 (Android 10; Mobile; rv:89.0) Gecko/89.0 Firefox/89.0` or later to verify that the header is correct and readable.

jrconlin:
So the problem isn't with Vapid. It's with how we were generating the
other authorization token, specifically differences with how python
and rust were handling things.

Python was converting the hex UAID string into a set of bytes, rust
was converting the 128bit UUID UAID into bytes. Combine that with some
fun regarding the base auth key string, and.. yeah.

I added a test to this to compare the auth key generated by python with
rust, broke the generate and validate functions into the
AuthorizationCheck impl and added a bunch of comments.

Closes #282

### DIFF
--- a/autoendpoint/src/extractors/authorization_check.rs
+++ b/autoendpoint/src/extractors/authorization_check.rs
@@ -20,24 +20,27 @@ use uuid::Uuid;
 pub struct AuthorizationCheck;
 
 impl AuthorizationCheck {
-
-    pub fn generate_token(auth_key: &str, user:&Uuid) -> Result<String, ErrorStack> {
+    pub fn generate_token(auth_key: &str, user: &Uuid) -> Result<String, ErrorStack> {
         sign_with_key(auth_key.as_bytes(), user.to_simple().to_string().as_bytes())
     }
 
-    pub fn validate_token(token: &str, uaid: &Uuid, auth_keys:&Vec<String>) -> Result<Self, ApiError> {
-            // Check the token against the expected token for each key
-            for key in auth_keys {
-                let expected_token = sign_with_key(key.as_bytes(), uaid.as_bytes())
-                    .map_err(ApiErrorKind::RegistrationSecretHash)?;
+    pub fn validate_token(
+        token: &str,
+        uaid: &Uuid,
+        auth_keys: &[String],
+    ) -> Result<Self, ApiError> {
+        // Check the token against the expected token for each key
+        for key in auth_keys {
+            let expected_token = sign_with_key(key.as_bytes(), uaid.as_bytes())
+                .map_err(ApiErrorKind::RegistrationSecretHash)?;
 
-                if expected_token.len() == token.len()
-                    && openssl::memcmp::eq(expected_token.as_bytes(), token.as_bytes())
-                {
-                    return Ok(Self);
-                }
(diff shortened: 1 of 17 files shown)


pjenvey (dismissed):
This looks good, but isn't it also an issue that `get_token_from_auth_header` only allows the `Bearer` scheme whereas [mobile sends `webpush`](https://github.com/mozilla-services/autopush-rs/issues/282#issuecomment-892183666)?

autoendpoint/src/extractors/authorization_check.rs
```diff
@@ -7,14 +7,45 @@ use actix_web::web::Data;
...
+use openssl::error::ErrorStack;
 use uuid::Uuid;
 
 /// Verifies the request authorization via the authorization header.
 ///
 /// The expected token is the HMAC-SHA256 hash of the UAID, signed with one of
 /// the available keys (allows for key rotation).
+/// NOTE: This is *ONLY* for internal calls that require authorization and should
+///      NOT be used by calls that are using VAPID authentication (e.g.
+///      subscription provider endpoints)
```
pjenvey, line 19: what do we mean by "local auth" exactly?

All the registration endpoints use this auth check.

jrconlin: There are two types of authentication: VAPID, which is meant for external, 3rd party subscription providers, and "local" User Agent authorization, which uses a bearer token provided as part of the registration process. 

The complication here is that both share the "Authentication" header. This call should only be used to check the latter Authentication form.

jrconlin:
Sigh. you're right. The fennec Android component does send the local, non-vapid Auth token `Authorization` header with a `webpush` prefix, not a `bearer` prefix. 

<expletives here>

pjenvey (approved):

autoendpoint/src/extractors/authorization_check.rs
```diff
@@ -7,14 +7,45 @@ use actix_web::web::Data;
...
+        uaid: &Uuid,
+        auth_keys: &[String],
+    ) -> Result<Self, ApiError> {
+        // Check the token against the expected token for each key
+        for key in auth_keys {
+            let expected_token =
+                sign_with_key(key.as_bytes(), uaid.to_simple().to_string().as_bytes())
+                    .map_err(ApiErrorKind::RegistrationSecretHash)?;
+
+            dbg!(&expected_token, &token);
```
AzureMarker, line 38: Temporary debugging code? This logs secrets.

jrconlin: Yep, this needs to go. It's not super critical since the local log is discarded and the token would already be compromised at that point, but it's needless chatter.

PR #291, opened 2021-10-19
feat: add status code and errno to reported metric for bridged errors

### LINKED ISSUE #288: Add metrics for FCM error responses
We currently do not record metrics for the 404, 410, and invalid token responses we get back from FCM. This means we are working in the dark when determining the sorts of faults we see with Send Tab. 

As a preliminary step, we should record a metric. We may also want to return this status back, however that may leak if a given endpoint was an android user.

jrconlin:
Closes #288

### DIFF
--- a/autoendpoint/src/routers/apns/router.rs
+++ b/autoendpoint/src/routers/apns/router.rs
@@ -10,7 +10,8 @@ use crate::routers::common::{
 use crate::routers::{Router, RouterError, RouterResponse};
 use a2::request::notification::LocalizedAlert;
 use a2::request::payload::{APSAlert, Payload, APS};
-use a2::{Endpoint, Error, NotificationOptions, Priority, Response};
+use a2::{self, Endpoint, NotificationOptions, Priority, Response};
+use actix_web::http::StatusCode;
 use async_trait::async_trait;
 use cadence::StatsdClient;
 use futures::{StreamExt, TryStreamExt};
@@ -42,7 +43,7 @@ trait ApnsClient: Send + Sync {
 
 #[async_trait]
 impl ApnsClient for a2::Client {
-    async fn send(&self, payload: Payload<'_>) -> Result<Response, Error> {
+    async fn send(&self, payload: Payload<'_>) -> Result<Response, a2::Error> {
         self.send(payload).await
     }
 }
@@ -124,9 +125,10 @@ impl ApnsRouter {
     async fn handle_error(&self, error: a2::Error, uaid: Uuid, channel: &str) -> ApiError {
         match &error {
             a2::Error::ResponseError(response) => {
+                let code = StatusCode::from_u16(response.code).unwrap_or(StatusCode::BAD_GATEWAY);
                 if response.code == 410 {
                     debug!("APNS recipient has been unregistered, removing user");
-                    incr_error_metric(&self.metrics, "apns", channel, "recipient_gone");
+                    incr_error_metric(&self.metrics, "apns", channel, "recipient_gone", code, None);
 
                     if let Err(e) = self.ddb.remove_user(uaid).await {
                         warn!("Error while removing user due to APNS 410: {}", e);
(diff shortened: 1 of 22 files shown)


jrconlin:
@pjenvey I don't suppose I could also get a review on this so I can land both for a new tag?

autoendpoint/src/routers/apns/router.rs
```diff
@@ -136,16 +138,30 @@ impl ApnsRouter {
                     return ApiError::from(ApnsError::Unregistered);
                 } else {
                     warn!("APNS error: {:?}", response.error);
-                    incr_error_metric(&self.metrics, "apns", channel, "server_error");
+                    incr_error_metric(&self.metrics, "apns", channel, "server_error", code, None);
```
pjenvey, line 141: I'm not sure what value `code` could become here but it looks like this `a2::Error` gets converted to an `ApnsError::ApnsUpstream` below making its code become a `BAD_GATEWAY` regardless.

autoendpoint/src/routers/apns/router.rs
```diff
@@ -125,9 +126,10 @@ impl ApnsRouter {
     async fn handle_error(&self, error: a2::Error, uaid: Uuid, channel: &str) -> ApiError {
         match &error {
             a2::Error::ResponseError(response) => {
+                let code = StatusCode::from_u16(response.code).unwrap_or(StatusCode::BAD_GATEWAY);
                 if response.code == 410 {
                     debug!("APNS recipient has been unregistered, removing user");
-                    incr_error_metric(&self.metrics, "apns", channel, "recipient_gone");
+                    incr_error_metric(&self.metrics, "apns", channel, "recipient_gone", code, None);
```
pjenvey, line 132: This becomes an `Unregistered` which does have an `errno` (106)

autoendpoint/src/routers/apns/router.rs
```diff
@@ -136,16 +138,30 @@ impl ApnsRouter {
...
             }
             a2::Error::ConnectionError => {
                 error!("APNS connection error");
-                incr_error_metric(&self.metrics, "apns", channel, "connection_unavailable");
+                incr_error_metric(
+                    &self.metrics,
+                    "apns",
+                    channel,
+                    "connection_unavailable",
+                    StatusCode::SERVICE_UNAVAILABLE,
```
pjenvey, line 151: Same w/ this one as above, it becomes `BAD_GATEWAY` w/ no `errno`.

pjenvey (changes requested):
I'm not against adding these new tags but I don't see them add much over what we already metric.

The `reason` tag is almost unique for all of these errors types (except `reason=unknown` which I haven't seen any records of currently in influx) So you can currently correlate this `reason` to an `error`/`errno` after digging into our code or docs, adding these just makes that a little easier.

We don't [trigger very many](https://earthangel-b40313e5.influxcloud.net/explore?orgId=1&left=%5B%22now-6h%22,%22now%22,%22svcops_aws%22,%7B%22policy%22:%22default%22,%22resultFormat%22:%22time_series%22,%22orderByTime%22:%22ASC%22,%22tags%22:%5B%7B%22key%22:%22environment%22,%22value%22:%22prod%22,%22operator%22:%22%3D%22%7D%5D,%22groupBy%22:%5B%7B%22type%22:%22time%22,%22params%22:%5B%22$__interval%22%5D%7D,%7B%22type%22:%22tag%22,%22params%22:%5B%22reason%22%5D%7D,%7B%22type%22:%22tag%22,%22params%22:%5B%22platform%22%5D%7D,%7B%22type%22:%22fill%22,%22params%22:%5B%22null%22%5D%7D%5D,%22select%22:%5B%5B%7B%22type%22:%22field%22,%22params%22:%5B%22value%22%5D%7D,%7B%22type%22:%22sum%22,%22params%22:%5B%5D%7D%5D%5D,%22measurement%22:%22autopush.notification.bridge.error%22%7D%5D) bridge errors either.

jrconlin:
Ah, part of the confusion here is that we're not capturing our error and response codes, we're capturing the ones that APNS gives us. We don't want these as sentry errors per-se because they're not fully actionable, but we do want to see trends. That's why a lot of these have no `errno`.  I realized that I got that confused as well and clarified.

autoendpoint/src/routers/apns/router.rs
```diff
@@ -125,27 +126,48 @@ impl ApnsRouter {
     async fn handle_error(&self, error: a2::Error, uaid: Uuid, channel: &str) -> ApiError {
         match &error {
             a2::Error::ResponseError(response) => {
+                // capture the APNs error as a metric response. This allows us to spot trends.
+                // While APNS can return a number of errors (see a2::response::ErrorReason) we
+                // shouldn't encounter many of those.
+                let reason = response
```
pjenvey, line 132: `ErrorReason::to_string` won't work for tags, [it returns descriptive values](https://docs.rs/a2/0.6.2/src/a2/response.rs.html#138) e.g. "The collapse identifier exceeds the maximum allowed size."

pjenvey (changes requested):

autoendpoint/src/routers/apns/router.rs
jrconlin: I'm kind of ok with that, since each string is unique, unless there's a tag value length limit we should avoid.

pjenvey: I believe the white space could be problematic. That enum's current `Debug` returns the variant name, I think that might be a better choice despite relying on a `Debug`. E.g. `format!("{:?}", a2::response::ErrorReason::BadCollapseId)` -> "BadCollapseId"

pjenvey (approved):

PR #296, opened 2022-02-25
chore: tag 1.59

jrconlin:


pjenvey (approved):
We'll need a new tag or retag, probably wanna squash this with a different commit message

PR #298, opened 2022-03-02
feat: Add postgres as a database source

jrconlin:
Work In Progress

Switch from DynamoDb to a more generic database as part of the GCP
transition.

General Work is being done before determining the final platform.

Issue: CONSRV-1535,


jrconlin:
Closing this PR. 
Support for Postgres is not required, since we are going with Bigtable. 

Later design improves the approach anyway.

PR #299, opened 2022-03-18
Bug/1707 ez vapid

jrconlin:
Covering two issues raised during the canary deploy of autopush-endpoint.

Due to rust's tighter typing requirements, we no longer accept quoted numerics in the VAPID JWT. The returned error would not give developers who are impacted by this change enough insight into what went wrong ("It was working before, what happened?") Added some text to the returned error message body to indicate what the cause of the new rejection is. (#CONSVC-1707 )

Many of the errors being logged to sentry from the new endpoint are not actionable. These are still errors, and require error handling, but do not need to be reported in detail. They can be converted to metrics. Added code to the sentry middleware to reduce the reporting load to sentry (#CONSVC-1708)


autoendpoint/src/extractors/subscription.rs
```diff
@@ -221,11 +221,27 @@ fn validate_vapid_jwt(vapid: &VapidHeaderWithKey, domain: &Url) -> ApiResult<()>
...
         &DecodingKey::from_ec_der(&public_key),
         &Validation::new(Algorithm::ES256),
-    )?;
+    ) {
+        Ok(v) => v,
+        Err(e) => match e.clone().into_kind() {
+            jsonwebtoken::errors::ErrorKind::Json(e) => {
+                if e.is_data() {
+                    return Err(VapidError::InvalidVapid(
+                        "A value in the vapid claims is either missing or incorrectly specified (e.g. \"exp\":\"12345\" or \"sub\":None). Please correct and retry.".to_owned(),
```
pjenvey, line 234: p
```suggestion
                        "A value in the vapid claims is either missing or incorrectly specified (e.g. \"exp\":\"12345\" or \"sub\":null). Please correct and retry.".to_owned(),
```

autoendpoint/src/middleware/sentry.rs
```diff
@@ -22,22 +23,46 @@ pub fn sentry_middleware(
         scope.add_event_processor(Box::new(move |event| process_event(event, &sentry_request)))
     });
 
+    let state = request
```
pjenvey, line 26: nit: It's probably not hurting anything as is but we might wanna avoid cloning the entire `ServerState` unless it's necessary. You could just pull out and clone solely the `metrics` instead.

autoendpoint/src/middleware/sentry.rs
```diff
@@ -22,22 +23,46 @@ pub fn sentry_middleware(
...
+                return Err(error);
             }
         };
 
         // Check for errors inside the response
         if let Some(error) = response.response().error() {
+            if let Some(api_err) = error.as_error::<ApiError>() {
+                if !api_err.kind.is_reportable() {
+                    debug!("Not reporting error (service error): {:?}", error);
+                    drop(_scope_guard);
```
pjenvey, line 62: I do wonder if both this and the other drop are actually necessary but fine with leaving it for now 🤷‍♂️

autoendpoint/src/error.rs
```diff
@@ -152,6 +152,58 @@ impl ApiErrorKind {
...
+            ApiErrorKind::Io(_) => "io",
+            ApiErrorKind::Metrics(_) => "metrics",
+            ApiErrorKind::Database(_) => "database",
+            ApiErrorKind::EndpointUrl(_) => "endpoint_url",
+            ApiErrorKind::RegistrationSecretHash(_) => "registration_secret_hash",
+        }
+    }
+
+    /// Don't report all errors to sentry
+    pub fn is_reportable(&self) -> bool {
```
pjenvey, line 191: We've discussed this before -- our old `reportable` naming is a bit confusing here and we should use something else. I called this `is_sentry_event` in contile.

autoendpoint/src/extractors/subscription.rs
```diff
@@ -221,11 +221,27 @@ fn validate_vapid_jwt(vapid: &VapidHeaderWithKey, domain: &Url) -> ApiResult<()>
...
     // NOTE: This will fail if `exp` is specified as a string instead of a numeric.
-    let token_data = jsonwebtoken::decode::<VapidClaims>(
+    let token_data = match jsonwebtoken::decode::<VapidClaims>(
         &vapid.token,
         &DecodingKey::from_ec_der(&public_key),
         &Validation::new(Algorithm::ES256),
-    )?;
+    ) {
+        Ok(v) => v,
+        Err(e) => match e.clone().into_kind() {
```
pjenvey, line 230:
```suggestion
        Err(e) => match e.kind() {
```

autoendpoint/src/error.rs
```diff
@@ -152,6 +152,58 @@ impl ApiErrorKind {
         }
     }
 
+    /// Specify the label to use for metrics reporting.
+    pub fn metric_label<'a>(&self) -> &'a str {
```
pjenvey, line 156: Can simplify this, it is and probably always will be `'static`

```suggestion
    pub fn metric_label(&self) -> &'static str {
```

pjenvey (changes requested):

autoendpoint/src/middleware/sentry.rs
jrconlin: I'll admit, I'm being a bit lazy here. `.cloned()` does a clone if `app_data()` returns Some. That saves a bit of complexity around dealing with `.app_data()` returning `None` for any reason.

autoendpoint/src/middleware/sentry.rs
jrconlin: Just figuring we should include it in the execute path.

autoendpoint/src/error.rs
jrconlin: Ah, right. Forgot about that. I'll change it.

pjenvey (approved):

PR #303, opened 2022-04-20
bug: add app_id to error message

### LINKED ISSUE #302: Display erroneous app_id in error message
While deploying the autopush-endpoint to canary we saw an usually high number of `No app ID found for user` errors. It's not a 100% failure rate, however, as about 50% of the traffic goes through without a hitch. 

We should record the "invalid" app-id as part of the the sentry error to determine if this is a string formatting error or some other bit of strangeness.

jrconlin:
This patch does two things in order to determine the spike in `InvalidAppId` messages:
1) Add the sent app_id to the InvalidAppId error message
2) strip quotes from the app_id (in case this is a json string
transposition error.)

Closes #302

### DIFF
--- a/autoendpoint/src/routers/apns/router.rs
+++ b/autoendpoint/src/routers/apns/router.rs
@@ -202,9 +202,7 @@ impl Router for ApnsRouter {
         router_input: &RouterDataInput,
         app_id: &str,
     ) -> Result<HashMap<String, Value>, RouterError> {
-        let quotes: &[_] = &['"', '\''];
-        let mapp = app_id.trim_matches(quotes);
-        if !self.clients.contains_key(mapp) {
+        if !self.clients.contains_key(app_id) {
             return Err(ApnsError::InvalidReleaseChannel.into());
         }
 
--- a/autoendpoint/src/routers/apns/router.rs
+++ b/autoendpoint/src/routers/apns/router.rs
@@ -202,7 +202,9 @@ impl Router for ApnsRouter {
         router_input: &RouterDataInput,
         app_id: &str,
     ) -> Result<HashMap<String, Value>, RouterError> {
-        if !self.clients.contains_key(app_id) {
+        let quotes: &[_] = &['"', '\''];
+        let mapp = app_id.trim_matches(quotes);
+        if !self.clients.contains_key(mapp) {
             return Err(ApnsError::InvalidReleaseChannel.into());
         }
 
--- a/autoendpoint/src/routers/fcm/client.rs
+++ b/autoendpoint/src/routers/fcm/client.rs
@@ -1,6 +1,6 @@
 use crate::routers::common::message_size_check;
 use crate::routers::fcm::error::FcmError;
-use crate::routers::fcm::settings::{FcmCredential, FcmSettings};
+use crate::routers::fcm::settings::{FcmServerCredential, FcmSettings};
 use crate::routers::RouterError;
 use reqwest::StatusCode;
 use serde::Deserialize;
@@ -21,16 +21,16 @@ pub struct FcmClient {
     endpoint: Url,
     timeout: Duration,
     max_data: usize,
-    auth: Option<DefaultAuthenticator>,
-    http: reqwest::Client,
-    credential: FcmCredential, // Only used for legacy GCM
+    fcm_authenticator: Option<DefaultAuthenticator>,
+    http_client: reqwest::Client,
+    server_credential: FcmServerCredential, // Only used for legacy GCM
 }
 
 impl FcmClient {
     /// Create an `FcmClient` using the provided credential
     pub async fn new(
         settings: &FcmSettings,
-        credential: FcmCredential,
+        server_credential: FcmServerCredential,
         http: reqwest::Client,
     ) -> std::io::Result<Self> {
         // `map`ping off of `serde_json::from_str` gets hairy and weird, requiring
@@ -38,12 +38,13 @@ impl FcmClient {
         // json detection does not. GCM keys are base64 values, so  `{` will never
         // appear in a GCM key. FCM keys are serialized JSON constructs.
         // These are both set in the settings and come from the `credentials` value.
-        let auth = if credential.credential.contains('{') {
+        let auth = if server_credential.server_access_token.contains('{') {
             trace!(
                 "Reading credential for {} from string...",
-                &credential.project_id
+                &server_credential.project_id
             );
-            let key_data = serde_json::from_str::<ServiceAccountKey>(&credential.credential)?;
+            let key_data =
+                serde_json::from_str::<ServiceAccountKey>(&server_credential.server_access_token)?;
             Some(
                 ServiceAccountAuthenticator::builder(key_data)
                     .build()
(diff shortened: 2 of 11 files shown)


autoendpoint/src/routers/fcm/router.rs
```diff
@@ -203,6 +225,13 @@ mod tests {
         map
     }
 
+    fn gcm_router_data() -> HashMap<String, serde_json::Value> {
+        let mut map = HashMap::new();
+        map.insert("auth".to_string(), serde_json::to_value(FCM_TOKEN).unwrap());
```
pjenvey, line 230: Per [gcm.py](https://github.com/mozilla-services/autopush/blob/5fd37e2/autopush/router/gcm.py#L69) we should expect auth/senderID to be nested under "creds".

autoendpoint/src/routers/fcm/router.rs
```diff
@@ -106,20 +108,36 @@ impl Router for FcmRouter {
...
+        // Newer "FCM" set the router data as "token", "app_id"
+        // Try reading as FCM and fall back to GCM.
+        let fcm_token = match router_data.get("token").and_then(Value::as_str) {
+            Some(v) => v,
+            None => router_data
+                .get("auth")
+                .and_then(Value::as_str)
+                .ok_or(FcmError::NoRegistrationToken)?,
+        };
+        let app_id = match router_data.get("app_id").and_then(Value::as_str) {
```
pjenvey, line 122: Are we sure these are the sole adjustments needed? The GCM vs FCM payload and Auth header look slightly different.

autoendpoint/src/routers/apns/router.rs
```diff
@@ -202,7 +202,9 @@ impl Router for ApnsRouter {
         router_input: &RouterDataInput,
         app_id: &str,
     ) -> Result<HashMap<String, Value>, RouterError> {
-        if !self.clients.contains_key(app_id) {
+        let quotes: &[_] = &['"', '\''];
+        let mapp = app_id.trim_matches(quotes);
```
pjenvey, line 206: I think we've ruled the quoting out completely so we might as well remove all these trim calls.

pjenvey (changes requested):

autoendpoint/src/routers/fcm/settings.rs
```diff
@@ -11,6 +11,24 @@ pub struct FcmSettings {
     pub min_ttl: usize,
     /// A JSON dict of `FcmCredential`s. This must be a `String` because
     /// environment variables cannot encode a `HashMap<String, FcmCredential>`
+    /// This contains both GCM and FCM credentials.
+    /// FCM (the more modern credential set) is specified as
+    ///
+    /// ```json
+    /// {"_project_id_":{"projectid": "_project_id_", "auth": "_key_"}, ...}
```
pjenvey, line 18:
```suggestion
    /// {"_project_id_":{"project_id": "_project_id_", "credential": "_key_"}, ...}
```

autoendpoint/src/routers/fcm/settings.rs
```diff
@@ -11,6 +11,24 @@ pub struct FcmSettings {
...
+    /// ```
+    ///
+    /// GCM follows the same pattern, where
+    ///
+    /// `_project_id_` == senderID and `_key_` == auth
+    /// e.g. for a FCM of "bar-project" and a GCM of "f00"
+    ///
+    /// ```json
+    /// {"bar-project":{"projectid": "bar-project-1234", "auth": "keys/bar-project-12345abcd.json"},
+    ///  "f00": {"projectid": "f00", "auth": "abcd0123457"},
```
pjenvey, line 28:
```suggestion
    /// {"bar-project":{"project_id": "bar-project-1234", "credential": "keys/bar-project-12345abcd.json"},
    ///  "f00": {"project_id": "f00", "credential": "abcd0123457"},
```

autoendpoint/src/routers/fcm/settings.rs
```diff
@@ -11,6 +11,24 @@ pub struct FcmSettings {
     pub min_ttl: usize,
     /// A JSON dict of `FcmCredential`s. This must be a `String` because
     /// environment variables cannot encode a `HashMap<String, FcmCredential>`
+    /// This contains both GCM and FCM credentials.
```
pjenvey, line 14: We should really mention how they're distinguished here as well (existence of "{" vs base64)) with the examples below reflecting that, but I'm fine deferring this to a separate issue though.

autoendpoint/src/routers/fcm/router.rs
```diff
@@ -70,6 +72,51 @@ impl FcmRouter {
...
+        let creds = router_data.get("creds").and_then(Value::as_object);
+        // I'm sure that there's a clever way to collapse these using `.map` or `.and_then`, but
+        // this is readable to me.
+        let auth = match router_data.get("token").and_then(Value::as_str) {
+            Some(v) => v.to_owned(),
+            None => {
+                if creds.is_none() {
+                    return Err(FcmError::NoRegistrationToken.into());
+                }
+                match creds
```
pjenvey, line 90: Is pulling from `creds.get("auth")` even necessary? The Python gcm.py router seems to always use `router_data.get("token")`.

pjenvey (changes requested):

autoendpoint/src/routers/fcm/router.rs
jrconlin: GCM stores [the server credential auth token](https://github.com/mozilla-services/autopush/blob/master/autopush/router/gcm.py#L70) as `auth`. This is different from the [registration id value](https://github.com/mozilla-services/autopush/blob/master/autopush/router/gcm.py#L110) stored in `token`

autoendpoint/src/routers/fcm/settings.rs
jrconlin: I thought I did that at line 27, where I have credentials for both, or do you mean something else?

jrconlin: ... and I found out that endpoint isn't optionally treating the credential value as a path identifier, like what the python version was doing.

autoendpoint/src/routers/fcm/router.rs
pjenvey: Right, it's stored as `auth` in `router_data`, but autopush python never actually reads from it. Instead the router reads `auth` from its own copy supplied from settings at startup (as we do in rust, reading FCMClient's `credential.credential`).

Whereas here we seem to be reading it as a fallback for `router_data.get("token")` which is `registration_id`, not `auth` (plus I don't think `router_data.get("token")` should ever be returning `None`)

Also this value's called `auth` in `send_gcm`, maybe call it `token` or `registration_id` instead?.

jrconlin: So, I think a BIG part of this is that the variables have terrible names. I'm going to try and fix that.

(_The following is also for my benefit..._)
What gets stored into the routing information is the `project_id` and the users `routing_token`. The `project_id` identifies which project to use, and is later matched with the `project_id` in the settings to get the corresponding `server_access_token` (which may be a value, a JSON block, or a file descriptor). FCM uses an `authenticator` function, and GCM needs to have access to the `server_credentials`, so both of those are stored into the `FcmClient` when it's initialized.

autoendpoint/src/routers/fcm/client.rs
```diff
@@ -18,42 +21,155 @@ pub struct FcmClient {
...
+        } else {
+            // check to see if this is a path to a file, and read in the credentials.
+            if Path::new(&server_credential.server_access_token).exists() {
+                warn!(
+                    "Reading credential for {} from file...",
+                    &server_credential.project_id
+                );
+                let mut file = File::open(&server_credential.server_access_token)?;
+                let mut content = String::new();
+                file.read_to_string(&mut content)?;
```
pjenvey, line 62: Rust added this helper at some point
```suggestion
                let content = fs::read_to_string(&server_credential.server_access_token)?;
```

pjenvey (commented):
Rest of it looks good

pjenvey (dismissed):

jrconlin:
Fuuuuuuuuuuuu...<expletive deleted>

pjenvey (approved):

PR #307, opened 2022-05-10
feat: Add tool to generate endpoints

jrconlin:
this tool should be useful in generating test endpoints for GCM on stage.

### DIFF
--- a/scripts/gendpoint.py
+++ b/scripts/gendpoint.py
@@ -1,0 +1,56 @@
+#! env python3
+
+import argparse
+import os
+
+from cryptography.fernet import Fernet
+
+
+def config(env_args: os._Environ) -> argparse.Namespace:
+    """Read the configuration from the args and environment."""
+    parser = argparse.ArgumentParser(
+        description="Generate Endpoint from environment and known UAID + CHID")
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


scripts/gendpoint.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def gendpoint(key: bytes, args: argparse.Namespace) -> str:
+    """Generate the endpoint"""
+    fernet = Fernet(key)
+    base = bytes(
+        list((bytes.fromhex(args.uaid.replace('-', '')) +
+              bytes.fromhex(args.chid.replace('-', '')))))
+    ekey = fernet.encrypt(base).strip(b'=').decode('utf-8')
+    return f"{args.root}/v1/{ekey}"
```
pjenvey, line 53: I guess v2 isn't needed for these? I can't recall when we use it vs v1.

pjenvey (dismissed):

scripts/gendpoint.py
jrconlin: Not really, since these would be 'data free' endpoints. 

Actually, thank you for reminding me, there are a few fixes I need to apply.

pjenvey (approved):

Issue #308, opened 2022-05-19
Resolve outstanding issues with GCM support


jrconlin:
There continue to be issues with GCM message delivery. We're seeing valid messages being rejected with an Authentication error. 

~If we cannot deliver a subscription because there is no user, subscription or valid authorization token to access the bridge system, we should remove that push subscription information.~

jrconlin:
GCM support was completely dropped by Google. This meant that documentation was removed, no tokens could be created, and there is no way to test. 
In addition, the error being returned is different than the prior Unauthorized error, which referred to the server credentials, not the user credentials. I initially thought the errors were from subscription providers which were ignoring 404/410 responses from autopush, but the actual error was that we were sending GCM traffic to the FCM path. I was able to find a copy of the original API spec, allowing me to see proper response samples and values, and adjust what the code does, however we still can't test.

PR #309, opened 2022-05-19
bug: Fix GCM handling

jrconlin:
This patch improves GCM handling, using recently discovered docs of the discontinued protocol. 

POST DEPLOY NOTE: 
There is currently an ignore in place for `Bridge authentication error` that will need to be removed.

Closes: #308

### DIFF
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -152,6 +152,16 @@ impl ApiErrorKind {
         }
     }
 
+    /// Should the User record be removed?
+    pub fn user_still_valid(&self) -> bool {
+        match self {
+            ApiErrorKind::NoUser | ApiErrorKind::NoSubscription | ApiErrorKind::InvalidToken => {
+                false
+            }
+            _ => true,
+        }
+    }
+
     /// Specify the label to use for metrics reporting.
     pub fn metric_label(&self) -> &'static str {
         match self {
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -152,11 +152,6 @@ impl ApiErrorKind {
         }
     }
 
-    /// Should the User record be removed?
-    pub fn user_still_valid(&self) -> bool {
-        !matches!(self, ApiErrorKind::InvalidToken)
-    }
-
     /// Specify the label to use for metrics reporting.
     pub fn metric_label(&self) -> &'static str {
         match self {
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -203,8 +203,8 @@ impl ApiErrorKind {
             self,
             // Ignore common webpush errors
             ApiErrorKind::NoTTL | ApiErrorKind::InvalidEncryption(_) |
-            // Ignore GCM router authentication errors otherwise
-            // we could blow through our quota.
+            // Do not report GCMAuthentication failures to Sentry since
+            // they're not really actionable and could flood our quota
             ApiErrorKind::Router(RouterError::GCMAuthentication) |
             // Ignore common VAPID erros
             ApiErrorKind::VapidError(_)
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -154,12 +154,7 @@ impl ApiErrorKind {
 
     /// Should the User record be removed?
     pub fn user_still_valid(&self) -> bool {
-        match self {
-            ApiErrorKind::NoUser | ApiErrorKind::NoSubscription | ApiErrorKind::InvalidToken => {
-                false
-            }
-            _ => true,
-        }
+        !matches!(self, ApiErrorKind::InvalidToken)
     }
 
     /// Specify the label to use for metrics reporting.
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -203,6 +203,9 @@ impl ApiErrorKind {
             self,
             // Ignore common webpush errors
             ApiErrorKind::NoTTL | ApiErrorKind::InvalidEncryption(_) |
+            // Ignore GCM router authentication errors otherwise
+            // we could blow through our quota.
+            ApiErrorKind::Router(RouterError::GCMAuthentication) |
             // Ignore common VAPID erros
             ApiErrorKind::VapidError(_)
             | ApiErrorKind::Jwt(_)
--- a/autoendpoint/src/error.rs
+++ b/autoendpoint/src/error.rs
@@ -198,9 +198,6 @@ impl ApiErrorKind {
             self,
             // Ignore common webpush errors
             ApiErrorKind::NoTTL | ApiErrorKind::InvalidEncryption(_) |
-            // Do not report GCMAuthentication failures to Sentry since
-            // they're not really actionable and could flood our quota
-            ApiErrorKind::Router(RouterError::GCMAuthentication) |
             // Ignore common VAPID erros
             ApiErrorKind::VapidError(_)
             | ApiErrorKind::Jwt(_)
--- a/autoendpoint/src/extractors/user.rs
+++ b/autoendpoint/src/extractors/user.rs
@@ -69,7 +69,7 @@ async fn validate_webpush_user(
 }
 
 /// Drop a user and increment associated metric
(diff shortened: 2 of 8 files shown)


autoendpoint/src/routers/fcm/client.rs
```diff
@@ -75,6 +76,10 @@ impl FcmClient {
                     server_credential.project_id
                 ))
                 .expect("Project ID is not URL-safe"),
+            gcm_endpoint: settings
+                .base_url
+                .join("gcm/send")
```
pjenvey, line 81:
```suggestion
                .join("fcm/send")
```

autoendpoint/src/routers/common.rs
```diff
@@ -57,6 +57,17 @@ pub async fn handle_error(
                 error.errno(),
             );
         }
+        RouterError::GCMAuthentication => {
+            // Don't record the GCM auth error.
+            incr_error_metric(
+                metrics,
+                platform,
+                app_id,
+                "gcm authentication",
```
pjenvey, line 66:
```suggestion
                "gcm_authentication",
```

autoendpoint/src/error.rs
```diff
@@ -193,6 +203,9 @@ impl ApiErrorKind {
             self,
             // Ignore common webpush errors
             ApiErrorKind::NoTTL | ApiErrorKind::InvalidEncryption(_) |
+            // Ignore GCM router authentication errors otherwise
+            // we could blow through our quota.
+            ApiErrorKind::Router(RouterError::GCMAuthentication) |
```
pjenvey, line 208: I think we should keep these Auth errors as a sentry event instead of making them a metric, seems to be more of an "internal" (autoendpoint is misconfigured) event. If you really want a metric though, we need to fix the tag value used below (per my code suggestion).

autoendpoint/src/routers/mod.rs
```diff
@@ -118,6 +121,7 @@ impl RouterError {
 
             RouterError::SaveDb(_) => StatusCode::SERVICE_UNAVAILABLE,
 
+            RouterError::GCMAuthentication | // Treat GCM auth failures as special, we reset the user.
```
pjenvey, line 124: I think this is outdated now, it isn't GONE
```suggestion
```

autoendpoint/src/routes/webpush.rs
```diff
@@ -12,15 +13,31 @@ use actix_web::HttpResponse;
...
+                // The user record is no longer valid. We should remove it so that the client
+                // can attempt to recover. We don't care if this succeeds or fails at this point.
+                if drop_user(
+                    notification.subscription.user.uaid,
+                    state.ddb.as_ref(),
+                    &state.metrics,
+                )
+                .await
+                .is_ok()
+                {};
```
pjenvey, line 36: the if .. {} is a weird read, alternatively:
```suggestion
                let _ = drop_user(
                    notification.subscription.user.uaid,
                    state.ddb.as_ref(),
                    &state.metrics,
                )
                .await;
```

pjenvey (changes requested):

autoendpoint/src/routers/common.rs
```diff
@@ -57,6 +57,17 @@ pub async fn handle_error(
                 error.errno(),
             );
         }
+        RouterError::GCMAuthentication => {
+            // Don't record the GCM auth error.
```
pjenvey, line 61: I think this comment is inaccurate, probably want a warn! line here like the others

autoendpoint/src/error.rs
```diff
@@ -152,6 +152,11 @@ impl ApiErrorKind {
         }
     }
 
+    /// Should the User record be removed?
+    pub fn user_still_valid(&self) -> bool {
+        !matches!(self, ApiErrorKind::InvalidToken)
```
pjenvey, line 157: `InvalidToken` is only produced by the extractors to signal the passed in token didn't parse. I think the only case we now should cover here is `RouterError::NotFound` but it's possible we don't need any of this at all (see my other comment)

autoendpoint/src/routes/webpush.rs
```diff
@@ -12,15 +13,29 @@ use actix_web::HttpResponse;
...
-    let response = router.route_notification(&notification).await?;
-
-    Ok(response.into())
+    match router.route_notification(&notification).await {
+        Ok(v) => Ok(v.into()),
+        Err(e) => {
+            if !e.kind.user_still_valid() {
+                // The user record is no longer valid. We should remove it so that the client
+                // can attempt to recover. We don't care if this succeeds or fails at this point.
+                let _ = drop_user(
```
pjenvey, line 29: I thought we already had a `drop_user` code path like this already but I couldn't find it last week, but I just found it: it's the `remove_user` call in `handle_error`, only triggered on `RouterError::NotFound`. The apns router has its own custom `handle_error` with its own `remove_user` call.

So it seems this change isn't necessary.

autoendpoint/src/routers/fcm/client.rs
```diff
@@ -138,18 +154,24 @@ impl FcmClient {
...
-                .map_err(FcmError::DeserializeResponse)?;
-
-            return Err(match (status, data.error) {
-                (StatusCode::UNAUTHORIZED, _) => RouterError::Authentication,
+        let data: GcmResponse = response
+            .json()
+            .await
+            .map_err(FcmError::DeserializeResponse)?;
+        if status.is_client_error() || status.is_server_error() || data.failure > 0 {
+            return Err(match (status, &data.results[0].error) {
```
pjenvey, line 162: `results[0]` could panic here if GCM ever sent us bad JSON (e.g. an empty array), let's guard against that.

autoendpoint/src/routers/fcm/client.rs
```diff
@@ -138,18 +154,24 @@ impl FcmClient {
...
+                (StatusCode::UNAUTHORIZED, _) => RouterError::GCMAuthentication,
                 (StatusCode::NOT_FOUND, _) => RouterError::NotFound,
```
pjenvey, line 164: I'm not sure GCM ever sends us `NOT_FOUND`? Totally fine keeping this here but might warrant a comment that we're not sure it's used.

pjenvey (changes requested):

autoendpoint/src/routers/fcm/client.rs
jrconlin: kinda/sorta tempted to say that if we get a panic here, it's indicative of a bigger problem we should be aware of, but yeah, I'm fine handling it.

autoendpoint/src/routers/fcm/client.rs
jrconlin: I have no idea if we get a `NOT_FOUND`, that's why I'm fine keeping it here.

autoendpoint/src/routes/webpush.rs
jrconlin: Yeah, I agree. I don't think we need the explicit `user_still_valid()` since we're now returning the correct error.

autoendpoint/src/error.rs
```diff
@@ -152,6 +152,11 @@ impl ApiErrorKind {
         }
     }
 
+    /// Should the User record be removed?
+    pub fn user_still_valid(&self) -> bool {
```
pjenvey, line 156: wanna kill this?

pjenvey (dismissed):

autoendpoint/src/error.rs
jrconlin: crap, must have snuck in during the refactor.

autoendpoint/src/extractors/user.rs
```diff
@@ -69,7 +69,7 @@ async fn validate_webpush_user(
 }
 
 /// Drop a user and increment associated metric
-async fn drop_user(uaid: Uuid, ddb: &dyn DbClient, metrics: &StatsdClient) -> ApiResult<()> {
+pub async fn drop_user(uaid: Uuid, ddb: &dyn DbClient, metrics: &StatsdClient) -> ApiResult<()> {
```
ddurst, line 72: Out of curiosity, why is `pub` added here?

pjenvey: It was needed on an earlier version of this PR, the `pub` can be removed now (though it's not terribly important)

pjenvey (approved):

PR #313, opened 2022-06-02
chore: Update for Jun 2022

jrconlin:


jrconlin:
Including QA since there's a slight alteration to the integration tests to allow for normalized environment variables.

pjenvey (dismissed):
This'll need a cloudops-deployment change for the env vars. You can replace the commit prefix "chore:" with "breaks:" and clog should put this in a Breaking Changes section to remind us about this.

You can also add "Broke: AUTOPUSH_ and AUTOEND_ env vars (now AUTOPUSH__ and AUTOEND__" at the end of the PR along with "Closes:" to additionally include that line in the changelog too.

pjenvey (approved):

Issue #318, opened 2022-06-27
Add jitter to retry for autoendpoint


jrconlin:
We're seeing a spike of ddb throttling.

![image](https://user-images.githubusercontent.com/35755/175995436-cd178855-52a0-4a2c-8ae7-a85307b86430.png)

The current theory is that this is due to autoendpoint hitting DDB more aggressively per client than the python version did. 

Add a jitter to the endpoints retry to more evenly distribute retrys.

jrconlin:
Note: This issue appears to be more of a problem for the endpoint than the connection server. To that end, I've only added the 'jitter' feature to endpoint.

For the modernization effort, which will use the same database method as endpoint, I'll add the 'jitter' as a common feature.

pjenvey:
AFAICT autoendpoint python doesn't retry any db calls, whereas the connection server retries many of its calls.

The rust connection server added retry logic on the db calls themselves, so with autoendpoint-rs "sharing" (it currently uses reimplementations of them that also include retry logic, but they'll be actually shared one day) it's now retrying whereas it didn't before.

pjenvey:
The connection db calls use `futures_backoff` whereas autoendpoint's using `again` crates for retry logic. These both work similarly and default to 1 second backoff delays, whereas autoendpoint overrides that default to use 100ms instead.

jrconlin:
Thanks. I was thinking that adding the jitter might help distribute retry load a bit more. I can't really see a clear correlation between traffic and throttling events although the general pattern is there (more events when there's more traffic in general). My dumb idea is that what might be happening is a bit of synchronicity between the retries and future traffic causing a sudden, sub-second spike up.

PR #319, opened 2022-06-27
bug: add jitter to retry

jrconlin:
Closes: #318

### DIFF
--- a/autoendpoint/src/db/retry.rs
+++ b/autoendpoint/src/db/retry.rs
@@ -1,12 +1,10 @@
 use again::RetryPolicy;
 use cadence::{CountedExt, StatsdClient};
-use rand::distributions::{Distribution, Uniform};
 use rusoto_core::RusotoError;
 use rusoto_dynamodb::{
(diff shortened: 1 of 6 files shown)


autoendpoint/src/db/retry.rs
```diff
@@ -48,6 +49,9 @@ pub fn retryable_describe_table_error(
 }
 
 /// Build an exponential retry policy
-pub fn retry_policy() -> RetryPolicy {
-    RetryPolicy::exponential(Duration::from_millis(100))
+pub fn retry_policy(jitter: u64) -> RetryPolicy {
```
pjenvey, line 52: Alternatively we could switch to `RetryPolicy::default()` which is `Exponential` at 1 second -- because this is what the connection server uses. I wonder if a jitter is even all that necessary with exponential backoff increasing the delay after every retry, but it wouldn't hurt -- though both retry libs also support jitter natively `RetryPolicy::default().with_jitter(true)`

PR #321, opened 2022-07-11
Fxa fixes

jrconlin:


autoendpoint/src/middleware/sentry.rs
```diff
@@ -35,6 +35,11 @@ pub fn sentry_middleware(
                 if let Some(api_err) = error.as_error::<ApiError>() {
                     // if it's not reportable, and we have access to the metrics, record it as a metric.
                     if !api_err.kind.is_sentry_event() {
+                        // The error (e.g. VapidErrorKind::InvalidKey(String)) might be too cardinal,
+                        // but we may need that information to debug a production issue. We can
+                        // add an info here, temporarily turn on info level debugging on a given server,
+                        // capture it, and then turn it off before we run out of money.
```
pjenvey, line 41: We really need to just add extras support, logged: https://mozilla-hub.atlassian.net/browse/CONSVC-1886

autoendpoint/src/error.rs
```diff
@@ -58,6 +58,9 @@ pub enum ApiErrorKind {
     #[error(transparent)]
     PayloadError(actix_web::Error),
 
+    #[error("Payload Error: {0}")]
+    PayloadErrorS(String),
```
pjenvey, line 62: Do we need this? I don't see it being used

autoendpoint/src/middleware/sentry.rs
jrconlin: Ah, good catch! I missed that.

autoendpoint/src/error.rs
jrconlin: Ah, good catch! I missed that.

autoendpoint/src/extractors/subscription.rs
```diff
@@ -292,7 +306,8 @@ mod tests {
             base64::STANDARD,
         )
         .unwrap();
-        let public_key = "BM3bVjW_wuZC54alIbqjTbaBNtthriVtdZlchOyOSdbVYeYQu2i5inJdft7jUWIAy4O9xHBbY196Gf-1odb8hds".to_owned();
+        // Specify a potentially invalid padding.
```
pjenvey, line 309: Might be worth it to make this a separate test case

pjenvey (approved):
Looks good but maybe the padding case could be a separate test case

PR #323, opened 2022-07-15
feat: allow for standard base64 private keys

jrconlin:
FxA sends in standard encoded base64 private key fields. The old Python
verison accepted them. We need to accept them as well.


autoendpoint/src/extractors/subscription.rs
```diff
@@ -190,6 +190,21 @@ fn version_1_validation(token: &[u8]) -> ApiResult<()> {
     Ok(())
 }
 
+/// Decode a public key string
+///
+/// NOTE: Some customers send a VAPID public key with incorrect padding and
+/// in standard base64 encoding. (Both of these violate the VAPID RFC)
+/// Prior python versions ignored these errors, so we should too.
+fn decode_public_key(public_key: &str) -> ApiResult<Vec<u8>> {
+    let encoding = if public_key.contains('/') || public_key.contains('+') {
```
pjenvey, line 199:
```suggestion
    let encoding = if public_key.contains(['/', '+']) {
```

pjenvey (changes requested):

pjenvey (approved):

PR #327, opened 2022-08-15
Bug/326 overflow

jrconlin:


autoendpoint/src/routers/apns/router.rs
```diff
@@ -335,6 +335,7 @@ mod tests {
     const DEVICE_TOKEN: &str = "test-token";
     const APNS_ID: &str = "deadbeef-4f5e-4403-be8f-35d0251655f5";
 
+    #[allow(clippy::type_complexity)]
```
jrconlin, line 338: We probably *should* make this a discrete type, but we only do this once, and I don't think it kills readability.

autoendpoint/src/routers/apns/router.rs
```diff
@@ -335,6 +335,7 @@ mod tests {
...
     /// A mock APNS client which allows one to supply a custom APNS response/error
     struct MockApnsClient {
         send_fn: Box<dyn Fn(Payload<'_>) -> Result<a2::Response, a2::Error> + Send + Sync>,
```
jrconlin, line 341: We probably should make this a discrete type, but we only do this once, and I don't think it kills readability.

pjenvey (approved):

PR #328, opened 2022-08-19
feat: Allow both string and integer VAPID exp values [CONSVC-1944]

jrconlin:
Several VAPID libraries use a string `exp` value. The current
jsonwebtoken library is a bit too strict about how it deals with this.

Closes: #CONSRV-1944

### DIFF
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -288,6 +288,9 @@ fn validate_vapid_jwt(
     } else if token_data.claims.exp.is_u64() {
         metrics.clone().incr("notification.auth.num_exp");
         token_data.claims.exp.as_u64().unwrap_or_default()
+    } else if token_data.claims.exp.is_f64() {
+        metrics.clone().incr("notification.auth.num_exp");
+        token_data.claims.exp.as_f64().unwrap_or_default().floor() as u64
     } else {
         warn!("Invalid exp value {:?}", &token_data.claims.exp);
         0
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -611,11 +611,7 @@ mod tests {
                 version_data: VapidVersionData::Version1,
             },
         };
-        let vv = validate_vapid_jwt(
-            &header,
-            &Url::from_str(domain).unwrap(),
-            &Metrics::noop(),
-        );
+        let vv = validate_vapid_jwt(&header, &Url::from_str(domain).unwrap(), &Metrics::noop());
         assert!(vv.is_ok());
         /*
         See https://www.rfc-editor.org/rfc/rfc8292.html#section-2.1
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -613,14 +613,12 @@ mod tests {
         };
         let vv = validate_vapid_jwt(
             &header,
-            &Url::from_str("http://example.org").unwrap(),
+            &Url::from_str(domain).unwrap(),
             &Metrics::noop(),
-        )
-        .unwrap_err()
-        .kind;
-        assert!(matches![
-            vv,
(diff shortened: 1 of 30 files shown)


autoendpoint/src/extractors/subscription.rs
```diff
@@ -263,8 +268,24 @@ fn validate_vapid_jwt(vapid: &VapidHeaderWithKey, domain: &Url) -> ApiResult<()>
         },
     };
 
-    // Check the signature and make sure the expiration is in the future, but not too far
-    if token_data.claims.exp > (sec_since_epoch() + ONE_DAY_IN_SECONDS) {
+    let exp = if token_data.claims.exp.is_string() {
+        u64::from_str(token_data.claims.exp.as_str().unwrap_or_default()).unwrap_or_default()
+    } else if token_data.claims.exp.is_u64() {
+        token_data.claims.exp.as_u64().unwrap_or_default()
```
pjenvey, line 274: It looks like `jsonwebtoken` also allows floats, so we probably should too?

https://github.com/Keats/jsonwebtoken/blob/2d2921f/src/validation.rs#L293

Their `fn numeric_type` there set as a `deserialize_with` is a nice way hook into serde to handle multiple input types. I suppose you could copy it over, adding a `visit_str` for this case but I'll admit I'm struggling to figure out why it's using `PhantomData` there (but it's also late Friday..)

autoendpoint/src/extractors/subscription.rs
```diff
@@ -263,8 +268,24 @@ fn validate_vapid_jwt(vapid: &VapidHeaderWithKey, domain: &Url) -> ApiResult<()>
...
+    } else {
+        warn!("Invalid exp value {:?}", &token_data.claims.exp);
+        0
+    };
+
+    if exp == 0 {
+        warn!("Invalid exp value {:?}", token_data.claims.exp);
+        return Err(VapidError::InvalidVapid(
+            "A value in the vapid claims is either missing or incorrectly specified (e.g. \"exp\":\"12345\" or \"sub\":null). Please correct and retry.".to_owned(),
```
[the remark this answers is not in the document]
jrconlin: Yeah, those are both errors that can trigger. Not sure if there's a better way to say that

autoendpoint/src/extractors/subscription.rs
jrconlin: I dunno, I would prefer that folk follow the spec, which is numeric UTC. Allowing string representation handles bad implementations, but those also stick to UTC. 

This lib is actually more forgiving than we used to be and allows for clock drift

pjenvey: The spec calls this NumericDate "A JSON numeric value" and "other than that non-integer values can be represented." so I think that accommodates floats. Blame JSON's number type.

pjenvey (changes requested):

autoendpoint/src/extractors/subscription.rs
```diff
@@ -236,19 +239,32 @@ fn version_2_validation(token: &[u8], vapid: Option<&VapidHeaderWithKey>) -> Api
...
+    metrics: &Metrics,
+) -> ApiResult<()> {
     let VapidHeaderWithKey { vapid, public_key } = vapid;
 
     let public_key = decode_public_key(public_key)?;
+    // disable automagic exp validation. (It's too strict, it requires a numeric exp, even though
+    // many vapid libraries don't provide it that way.)
+    let mut validation = jsonwebtoken::Validation::new(jsonwebtoken::Algorithm::ES256);
+    validation.validate_exp = false;
+    validation.required_spec_claims.remove("exp");
```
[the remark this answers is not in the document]
jrconlin: `exp` is still required. This removes the strict check that validation does which requires `exp` to be a numeric field. 

It just means that we need to do the validation after.

jrconlin:
closed in favor of https://github.com/mozilla-services/autopush-rs/pull/340

PR #329, opened 2022-08-23
feat: Add timeouts for Client::reqwest calls [CONSVC-3289]

jrconlin:
These are used for routing and the like.
Currently default for:

```
// Initial connection timeout
connection_timeout_millis: 1000
// Total request timeout
request_timeout_millis: 3000
```
Closes: #CONSVC-1632

### DIFF
--- a/autoendpoint/src/server.rs
+++ b/autoendpoint/src/server.rs
@@ -24,6 +24,7 @@ use actix_web::{
 use cadence::StatsdClient;
 use fernet::MultiFernet;
 use std::sync::Arc;
+use std::time::Duration;
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


pjenvey (approved):
👍 but unfortunately this doesn't help the APNS router because `a2` uses its own hyper client internally with no way to set timeouts

jrconlin:
Yep, working on a separate patch for a2 to include passing a custom ClientBuilder, which can have timeouts.

pjenvey:
@jrconlin Looks like it would be easier to move a2 to `reqwest` if the maintainer's are up for it. Only reason I can think they wouldn't is if it adds more dependency bloat to a2.

PR #330, opened 2022-09-01
 feat: Add additional metrics for message tracking [CONSVC-1660]

jrconlin:
* `notification.message.expired` -- In flight message has no TTL, endpoint is unavailable.
* `notification.message.stored` -- message stored for later transmission
* `notification.message.retrieved` -- message extracted from storage
* `ua.notification.sent` -- sent a notification to the websocket endpoint

Issue: CONSVC-1660

### DIFF
--- a/autoendpoint/src/routers/webpush.rs
+++ b/autoendpoint/src/routers/webpush.rs
@@ -73,11 +73,16 @@ impl Router for WebPushRouter {
         }
 
         if notification.headers.ttl == 0 {
+            let topic = notification.headers.topic.is_some().to_string();
             trace!(
                 "Notification has a TTL of zero and was not successfully \
                  delivered, dropping it"
             );
-            self.metrics.incr("notification.message.expired")?;
+            self.metrics
+                .incr_with_tags("notification.message.expired")
+                // TODO: include `internal` if meta is set.
+                .with_tag("topic", &topic)
+                .send();
             return Ok(self.make_delivered_response(notification));
         }
 
@@ -177,13 +182,6 @@ impl WebPushRouter {
             .save_message(notification.subscription.user.uaid, bundle.into())
             .await
             .map_err(|e| ApiErrorKind::Router(RouterError::SaveDb(e)).into())
-            .map(|_| {
-                self.metrics
-                    .incr_with_tags("notification.message.stored")
-                    .with_tag("topic", &notification.headers.topic.is_some().to_string())
-                    // TODO: include `internal` if meta is set.
-                    .send();
-            })
     }
 
     /// Remove the node ID from a user. This is done if the user is no longer
--- a/autoendpoint/src/routers/webpush.rs
+++ b/autoendpoint/src/routers/webpush.rs
@@ -77,6 +77,7 @@ impl Router for WebPushRouter {
                 "Notification has a TTL of zero and was not successfully \
                  delivered, dropping it"
             );
+            self.metrics.incr("notification.message.expired")?;
             return Ok(self.make_delivered_response(notification));
         }
 
@@ -166,6 +167,9 @@ impl WebPushRouter {
             )
             .await
             .map_err(|e| ApiErrorKind::Router(RouterError::SaveDb(e)).into())
+            .map(|_| {
+                self.metrics.incr("notification.message.stored").ok();
+            })
     }
 
     /// Remove the node ID from a user. This is done if the user is no longer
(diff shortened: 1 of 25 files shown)


autopush/src/client.rs
```diff
@@ -1344,6 +1344,10 @@ fn emit_metrics_for_send(metrics: &StatsdClient, notif: &Notification, source: &
     if notif.topic.is_some() {
         metrics.incr("ua.notification.topic").ok();
     }
+    metrics
+        .incr_with_tags("ua.notification.sent")
+        .with_tag("source", source)
+        .send();
```
pjenvey, line 1350: A thought: we could kill the `ua.notification.topic` above for a topic tag here like the other new metrics are doing

pjenvey (changes requested):

autoendpoint/src/routers/webpush.rs
```diff
@@ -166,6 +167,13 @@ impl WebPushRouter {
             )
             .await
             .map_err(|e| ApiErrorKind::Router(RouterError::SaveDb(e)).into())
+            .map(|_| {
+                self.metrics
+                    .incr_with_tags("notification.message.stored")
+                    .with_tag("topic", &notification.headers.topic.is_some().to_string())
+                    // TODO: include `internal` if meta is set.
+                    .send();
+            })
```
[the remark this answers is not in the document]
jrconlin: I kinda debated that a bit. Basically, the problem was the tag info, but perhaps I just need to pass that along as a parameter...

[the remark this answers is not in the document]
pjenvey: In this specific case it looked like the only tag info needed (topic) was available in the object passed to `save_message` (it's a slightly different `Notification` type, so it ends up being `notification.topic.is_some()` instead)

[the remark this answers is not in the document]
jrconlin: Also realized that part of my waffling on this was because I wanted to have a consistent layer where we record metrics. I think you're right that `DbClientImpl` is better than the calling layer, even if that means that the `DbClientImpl` may not have *all* the contexts that we may need or want. We should just add those to the `Notification` class when we need to.

autopush-common/src/db/mod.rs
```diff
@@ -329,12 +329,15 @@ impl DynamoStorage {
...
         };
-
+        let metrics = self.metrics.clone();
         retry_if(
             move || ddb.put_item(put_item.clone()),
             retryable_putitem_error,
         )
-        .and_then(|_| future::ok(()))
+        .and_then(move |_| {
+            metrics.incr("notification.db.stored").ok();
```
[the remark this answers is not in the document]
jrconlin: If we're going to move reporting down to `DbClientImpl`, I'm also going to move the `notification.message.deleted` to the same crate.

autoendpoint/src/routers/webpush.rs
```diff
@@ -117,6 +123,17 @@ impl Router for WebPushRouter {
...
                     trace!("Node has delivered the message");
+                    self.metrics
+                        .time_with_tags(
+                            "notification.total_request_time",
+                            (notification.timestamp - autopush_common::util::sec_since_epoch())
+                                * 1000,
+                        )
+                        .with_tag("platform", "websocket")
+                        .with_tag("app_id", "direct")
+                        .with_tag("stored", &notification.stored.to_string())
```
pjenvey, line 134: This is the only place one of the new `stored` fields is used, and I think it's inaccurate here? as the `store_notification` would have always been called by now and it doesn't modify this field on the `notification` object (only on the internal copy it makes).

I don't think we need or should probably have any of new `stored` fields -- especially on `DynamoDbNotification`, that type basically says the notification is stored (or will be stored shortly by the end of the request). It can become tricky/potentially buggy to keep accurate, feels like more the job of an ORM unless it's really necessary.

pjenvey (changes requested):

autoendpoint/src/routers/webpush.rs
jrconlin: It was originally added because I wanted to know if messages were getting dropped by storage. Adding the flag would at least indicate the source of the message being sent through. 

It's not super necessary and there might be ways to better address if we determine it's a point of loss.

autopush/src/client.rs
```diff
@@ -600,36 +603,37 @@ where
         }
         let now = ms_since_epoch();
         let elapsed = (now - webpush.connected_at) / 1_000;
-        let (ua_result, metrics_os, metrics_browser) = parse_user_agent(&user_agent);
+        let ua_info = UserAgentInfo::from(user_agent.as_ref());
```
pjenvey, line 606: Sorry one last thing, you shouldn't need to parse it again:

```suggestion
        let ua_info = webpush.ua_info.clone();
```

jrconlin: Thanks! I realized that we don't need to pass the user_agent string anymore, so that's also been removed.

pjenvey (dismissed):

pjenvey (approved):

PR #337, opened 2022-12-13
Update: modernize to rust 1.65

jrconlin:
* use workspace cargo dependencies to ensure crate/meta consistency
* update libraries and calls
* include `-A dead_code` to clippy to get around `future_state_machine`
* modern fmt/clippy recommendations

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,5 +5,43 @@ members = [
     "autoendpoint",
 ]
 
+[workspace.package]
+version = "1.65.1"
+authors = [
+  "Ben Bangert <ben@groovie.org>",
+  "JR Conlin <jrconlin@mozilla.com>",
+  "Phil Jenvey <pjenvey@underboss.org>",
+  "Alex Crichton <alex@alexcrichton.com>",
+  "Mark Drobnak <mdrobnak@mozilla.com>",
+]
+edition = "2021"
+
+[workspace.dependencies]
+# ideally, this would contain any crates that are shared between crates.
+# there are some lingering code interdependencies that prevent that, but it should
+# remain a goal for the overall refactor.
+base64 = "0.20"
+cadence = "0.29"
+config = "0.13"
+fernet = "0.2.0"
+hex = "0.4.2"
+lazy_static = "1.4"
+log = { version = "0.4", features = ["max_level_info", "release_max_level_info"] }
+openssl = "0.10"
+# Using debug-logs avoids https://github.com/getsentry/sentry-rust/issues/237
+sentry = { version = "0.19", features = ["debug-logs", "with_error_chain"] }  # Keep on 0.19 in order to work with our backend and use `with_error_chain`
+serde = "1.0"
+serde_derive = "1.0"
+serde_dynamodb = "0.4"  # 0.7+ requires carg 0.46+
+serde_json = "1.0"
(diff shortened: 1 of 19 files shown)


Cargo.toml
```diff
@@ -5,5 +5,43 @@ members = [
...
+authors = [
+  "Ben Bangert <ben@groovie.org>",
+  "JR Conlin <jrconlin@mozilla.com>",
+  "Phil Jenvey <pjenvey@underboss.org>",
+  "Alex Crichton <alex@alexcrichton.com>",
+  "Mark Drobnak <mdrobnak@mozilla.com>",
+]
+edition = "2021"
+
+[workspace.dependencies]
```
pjenvey, line 19: Cool, I hadn't even noticed this stuff in the 1.64 release notes. cc: @ethowitz

autoendpoint/src/tags.rs
```diff
@@ -92,6 +92,7 @@ impl Tags {
     }
 }
 
+/*
```
pjenvey, line 95: Why commented out, is this just not used in autoendpoint?

autoendpoint/src/extractors/subscription.rs
```diff
@@ -196,13 +196,21 @@ fn version_1_validation(token: &[u8]) -> ApiResult<()> {
 /// in standard base64 encoding. (Both of these violate the VAPID RFC)
 /// Prior python versions ignored these errors, so we should too.
 fn decode_public_key(public_key: &str) -> ApiResult<Vec<u8>> {
-    let encoding = if public_key.contains(['/', '+']) {
-        base64::STANDARD_NO_PAD
+    let engine = if public_key.contains(['/', '+']) {
+        base64::engine::fast_portable::FastPortable::from(
```
pjenvey, line 200: Let's take advantage of `FastPortable::from` being a const method and make all them constants (at least outside of tests, also have some in `endpoint.rs` and `notification.rs`), its [docs recommend it](https://docs.rs/base64/latest/base64/engine/fast_portable/struct.FastPortable.html#method.from) even though they're not very expensive to construct.

Cargo.toml
ethowitz: This is neat!

autoendpoint/src/tags.rs
jrconlin: It's commented out because the `slog::Key::from(key.clone())` no longer produces a string that is equivalent to `static. I'd like to eventually return the function, but for now, I've gone and converted the Keys to a formatted value.

autoendpoint/src/extractors/subscription.rs
jrconlin: Ah, cool! Didn't see that in the docs. I was mostly just focused on getting the stuff working.

autoendpoint/src/tags.rs
pjenvey: This really confused me after seeing the `slog::Key` docs state it implements `impl From<String> for Key` -- turns out it only really [uses the "dynamic" version of `Key`](https://github.com/slog-rs/slog/blob/master/src/key/mod.rs) when the `dynamic-keys` feature flag is enabled, which we're missing. So switching it back on should let us uncomment this again.

pjenvey (approved):

PR #340, opened 2023-01-19
feat: add metrics for vapid errors

pjenvey:
from the work for:
SYNC-3389

Co-authored-by: JR Conlin <jconlin+git@mozilla.com>

@jrconlin this is just the metric additions extracted from your #328

Context: When we determined autoendpoint-rs's stricter validation vs autoendpoint-python's wasn't the cause of the original issue that triggered [SYNC-3389](https://mozilla-hub.atlassian.net/browse/SYNC-3389), implementing laxer validation to match the python became optional to add. So at this point (especially a half year later) it doesn't seem worth adding, in particular because the jsonwebtoken workarounds are annoying/adding some complexity.

[SYNC-3389]: https://mozilla-hub.atlassian.net/browse/SYNC-3389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrconlin (approved):
🏳️

PR #344, opened 2023-01-26
feat: Add metrics to try and analyze UpdateItem bug.

jrconlin:
Includes cargo fix for 1.67 clippy strings.
see relevant changes in

autoendpoint::routers::webPush::WebPushRouter::route_notification

Issue: SYNC-3574


autoendpoint/src/routers/webpush.rs
```diff
@@ -66,8 +68,38 @@ impl Router for WebPushRouter {
                 }
                 Err(error) => {
                     // We should stop sending notifications to this node for this user
+                    if error.is_timeout() {
```
jrconlin, line 71: Start of additional metric fix.

I'm adding the node_id as a tag mostly because I want to see if there's some corruption that might be happening when we're pulling the record out of storage.

autopush/src/megaphone.rs
```diff
@@ -41,20 +40,20 @@ impl BroadcastRegistry {
...
     }
 
     fn lookup_id(&self, key: BroadcastKey) -> Option<String> {
-        self.table.get(key.0 as usize).cloned()
+        self.table.get(key as usize).cloned()
     }
 
     fn lookup_key(&self, broadcast_id: &str) -> Option<BroadcastKey> {
-        self.lookup.get(broadcast_id).cloned().map(BroadcastKey)
+        self.lookup.get(broadcast_id).cloned()
```
pjenvey, line 56: while you're here:

```suggestion
        self.lookup.get(broadcast_id).copied()
```

autoendpoint/src/routers/webpush.rs
```diff
@@ -66,8 +68,33 @@ impl Router for WebPushRouter {
...
+                            .incr_with_tags("error.node.connection")
+                            .with_tag("node_id", node_id)
+                            .try_send()
+                            .ok();
+                    }
                     debug!("Error while sending webpush notification: {}", error);
-                    self.remove_node_id(user, node_id.clone()).await?;
+                    if let Some(err) = self.remove_node_id(user, node_id.clone()).await.err() {
+                        match &err.kind {
+                            ApiErrorKind::Database(crate::db::error::DbError::UpdateItem(_)) => {
```
pjenvey, line 88: I think we can further narrow this to only `UpdateItemError::ConditionalCheckFailed` if we want that?

jrconlin:
I was mostly thinking that these metrics would capture the `node_id`, but after thinking about it last night, I agree that they might just introduce more noise than we want. I'll pull them.

autoendpoint/src/routers/webpush.rs
```diff
@@ -186,13 +186,42 @@ impl WebPushRouter {
...
+                rusoto_dynamodb::UpdateItemError::ConditionalCheckFailed(_),
+            )) = &err
+            {
+                /*
+                Most likely, the node_id recorded in the db does not match the
+                node_id contained in the notification. Track these as a metric
+                for now.
+                */
+                self.metrics
+                    .incr_with_tags("error.node.update")
```
pjenvey, line 214: On second thought this metric could instead become an optional tag added on to `updates.client.host_gone` emitted in this method, but they probably don't occur enough to where I care that much.

pjenvey (changes requested):

autoendpoint/src/routers/webpush.rs
jrconlin: No, I forgot to remove this when I reduced the call to DbClient. 

The Client should eat the error because it's the source. Ideally webpush should have no knowledge of the underlying storage mechanism.

PR #348, opened 2023-02-17
bug: make CORS default less restrictive.

jrconlin:
The default actix-cors is even more restrictive than it was before. We need to loosen it up again, so this time, be more explicit about what we're allowing. 

(I wonder if we should just use `.permissive()` since we handle most of the actual security internally?)

Fixes: Sync-3608


autoendpoint/src/server.rs
```diff
@@ -95,12 +95,23 @@ impl Server {
...
+            // These have a bad habit of being reset. Specify them explicitly.
+            let cors = Cors::default()
+                .allow_any_origin()
+                .allow_any_header()
+                .allowed_methods(vec![
+                    actix_web::http::Method::DELETE,
+                    actix_web::http::Method::GET,
+                    actix_web::http::Method::POST,
+                    actix_web::http::Method::PUT,
+                ])
```
linabutler, line 107: Do you want to expose the `Location` response header with `.expose_headers()`, also?

jrconlin: Not seeing a `Location` header, but yeah, there are a few X-Forward headers we should probably not disclose. 

sigh

Really didn't want to list out all the headers either.

jrconlin: Hmm... Looking at [allow_any_headers](https://docs.rs/actix-cors/latest/actix_cors/struct.Cors.html#method.allowed_headers), the call will echo the list of headers requested in  `Access-Control-Request-Headers` into `Access-Control-Allow-Headers`, so it's not returning all the headers that are sent, only the ones that are requested to be sent, and even then, it's only returning the names, not the values.

linabutler: Ohh, gotcha—I thought you might be returning a `Location` header in 201s for `POST` requests, sorry! Thanks for double-checking, @jrconlin! 😊

pjenvey: We do send a `Location` header (routers send them via `RouterResponse::success`) as well as `TTL`. The [python also sent](https://github.com/mozilla-services/autopush/blob/master/autopush/web/webpush.py#L497) `www-authenticate` alongside errors but the rust doesn't.

It's not clear to me that these need exposing, but I guess the Test Page might want access to them in theory which would require exposing them, I think?

pjenvey (approved):

PR #350, opened 2023-03-15
Update Autopush dependencies and libraries

jrconlin:
Before adding in the work for GCP, we need to update the various libraries and dependencies for Autopush.

This branch introduces the work in progress `autopush-connect`, but that code is not active.

_Changes include:_
* removal of some older RUSTSEC ignores
* remove now duplicated code from `autoendpoint`
* drop unneeded code from `autopush_common`
* updates to futures (0.3), actix (3 -> 4), tokio (0.2 -> 1.26) and others
* use `workspace` package management to consolidate versions for sub-crates
* Update to rust 1.68
* Added copious comments, more descriptive `trace!` and `debug!` messages
* Moved legacy `autopush-rs` dependencies to `autopush` crate. 
* Converted `autopush-common` to use `autoendpoint` compatible database structures. (_*NOTE*_ future versions will change the way that `message_table`, `router_table` and which databases to use.)

Closes: SYNC-3443,  SYNC-3444, SYNC-3446, SYNC-3453, SYNC-3454, SYNC-3450, SYNC-3452


autoconnect/autoconnect-web/src/dockerflow.rs
```diff
@@ -0,0 +1,58 @@
...
+        "errno" => 999,
+    );
+
+    thread::spawn(|| {
+        panic!("LogCheck");
+    });
+
+    HttpResponse::BadRequest()
+        .status(StatusCode::IM_A_TEAPOT)
+        .finish()
```
pjenvey, line 57: You can do `HttpResponse::build(StatusCode::IM_A_TEAPOT).finish()` instead but I think `new` does all of that for you.

```suggestion
    HttpResponse::new(StatusCode::IM_A_TEAPOT)
```

autoendpoint/src/server.rs
```diff
@@ -16,23 +28,15 @@ use crate::routes::registration::{
...
-use std::time::Duration;
 
 #[derive(Clone)]
 pub struct ServerState {
     /// Server Data
     pub metrics: Arc<StatsdClient>,
     pub settings: Settings,
     pub fernet: MultiFernet,
-    pub ddb: Box<dyn DbClient>,
+    // TODO: Convert this to the autopush_common dbCommandClient impl.
```
pjenvey, line 38: kill

autoendpoint/src/server.rs
```diff
@@ -47,11 +51,25 @@ impl Server {
...
+            } else {
+                settings.db_settings.clone()
+            },
+        };
+        let ddb: Box<dyn DbClient> = match StorageType::from_dsn(&db_settings.dsn) {
+            StorageType::DynamoDb => Box::new(
+                DdbClientImpl::new(metrics.clone(), &db_settings)
+                    .map_err(ApiErrorKind::Database)?,
+            ),
+            StorageType::INVALID => {
```
pjenvey, line 69: This will break the deployed autoendpoint because it doesn't set a `db_dsn`. I'm not sure how this will work as it is currently as setting `db_dsn` currently seems to imply the use of the emulator, as it causes main to set `AWS_LOCAL_DYNAMODB`.

I think we need some way of signaling: use dynamodb/no emulator stuff/use the autolookup from `rusoto_core::Region::default()`

autoendpoint/src/server.rs
```diff
@@ -96,15 +114,16 @@ impl Server {
 
         let server = HttpServer::new(move || {
             App::new()
-                .data(state.clone())
+                .app_data(web::Data::new(state.clone()))
                 // Middleware
                 .wrap(ErrorHandlers::new().handler(StatusCode::NOT_FOUND, ApiError::render_404))
-                .wrap_fn(sentry_middleware)
+                // .wrap_fn(sentry_middleware)
```
pjenvey, line 120: We really need to resolve the sentry middleware and the bytes limit before we can roll autoendpoint out. Should we log issues for this?

autopush-common/src/db/client.rs
```diff
@@ -0,0 +1,104 @@
...
+        timestamp: Option<u64>,
+        limit: usize,
+    ) -> DbResult<FetchMessageResponse>;
+
+    /// Delete a notification
+    async fn remove_message(&self, uaid: &Uuid, sort_key: &str) -> DbResult<()>;
+
+    /// record a Hello record
+    /// Each data store can handle this differently, thus it's best to hand things off to the engine.
+    async fn hello(
```
pjenvey, line 80: I'll probably delve into this later myself but I'd love it if we could separate out the autoconnect db calls from autoendpoint. Maybe the former would get another sub trait of the DbClient trait.

Some of how that's done might depend on whether we could say implement `hello` at a higher level, like on the trait itself, since it's maybe not db specific.

autopush-common/src/db/dynamodb/commands.rs
```diff
@@ -0,0 +1,505 @@
+use std::collections::HashSet;
```
pjenvey, line 1: Can this module be killed? Or at least some of it, as doesn't all or most of it now reside in `DbCllientImpl`?

autopush-common/src/db/dynamodb/retry.rs
```diff
@@ -48,5 +50,5 @@ pub fn retryable_describe_table_error(
 
 /// Build an exponential retry policy
 pub fn retry_policy() -> RetryPolicy {
-    RetryPolicy::default().with_jitter(true)
+    RetryPolicy::exponential(Duration::from_millis(100))
```
pjenvey, line 53: Can we revert this for now? If we're going to change it it really needs its own tracking issue/paper trail.

autopush-common/src/db/dynamodb/mod.rs
```diff
@@ -0,0 +1,678 @@
+use std::collections::HashSet;
```
pjenvey, line 1: Note to self: I have not reviewed this module yet..

autopush-common/src/db/mod.rs
```diff
@@ -1,594 +1,427 @@
...
-                hello_response.deferred_user_registration = Some(user);
-                future::Either::B(Box::new(future::ok(hello_response)))
-            }
-        })
-    }
+    ) -> Result<HelloResponse>;
 
-    pub fn register(
+    /// register a new channel for this UAID
+    async fn register(
```
pjenvey, line 114: Can we name this:

```suggestion
    async fn register_channel(
```

and `unregister` -> `unregister_channel`

autopush-common/src/errors.rs
```diff
@@ -99,8 +152,27 @@ pub enum ApcErrorKind {
...
+            }
+            _ => StatusCode::INTERNAL_SERVER_ERROR,
+        }
+    }
 }
 
 pub type Result<T> = std::result::Result<T, ApcError>;
 
-pub type MyFuture<T> = Box<dyn Future<Item = T, Error = ApcError>>;
+// pub type MyFuture<T> = Box<dyn Future<Item = T, Error = ApcError>>;
```
pjenvey, line 178: kill

autopush-common/src/util/mod.rs
```diff
@@ -1,43 +1,16 @@
...
-use std::time::Duration;
 
-use futures::future::{Either, Future, IntoFuture};
-use tokio_core::reactor::{Handle, Timeout};
+use base64::Engine;
 
-use crate::errors::*;
-
-mod send_all;
+// mod send_all; // kill?
```
pjenvey, line 7: kill!

autopush-common/src/util/mod.rs
```diff
@@ -1,43 +1,16 @@
...
 pub mod timing;
+pub mod user_agent;
 
-pub use self::send_all::MySendAll;
+// pub use self::send_all::MySendAll;
```
pjenvey, line 11: +kill

autopush/src/server/mod.rs
```diff
@@ -891,14 +907,41 @@ where
...
         try_ready!(self.send_ws_ping());
-        Ok(self.inner.poll_complete()?)
+        if self.inner.poll_complete().is_err() {
+            warn!("Error encountered with poll_complete");
+        }
+        Ok(Async::Ready(()))
     }
 
     fn close(&mut self) -> Poll<(), ApcError> {
+        /*
```
pjenvey, line 922: kill this comment

autopush/src/server/mod.rs
```diff
@@ -939,7 +982,7 @@ fn write_log_check(socket: WebpushIo) -> MyFuture<()> {
         StatusCode::IM_A_TEAPOT,
         json!({
                 "code": code,
-                "errno": 999,
+                "errno": 99,
```
pjenvey, line 985: ?

pjenvey (commented):
@jrconlin Sorry, I don't want to hold this branch up over every little thing but I also wanted to point out things as I went that we don't want to forget about. I think the most pressing thing is fixing autoendpoint so we can roll out a new version at some point.

Feel free to push back on some of these -- I think instead of new issues let's just have a "fix all TODOs" issue and make sure everything that needs a TODO comment has one.

autopush-common/src/db/client.rs
pjenvey: I guess there's somewhat of a WIP on that separate trait with that DbCommands trait..

jrconlin: I don't know. Having all the functions contained this way ensures that any future datastore must implement them all, rather than have to chase down multiple structs in various crates.

autopush-common/src/db/dynamodb/commands.rs
jrconlin: Agreed. The goal is to remove it, so I'll make sure all the functions are transitioned.

.circleci/config.yml
```diff
@@ -194,6 +194,34 @@ jobs:
 
 workflows:
   version: 2
+  build-test:
```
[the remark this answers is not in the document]
jrconlin: fair, it's a bit of left over from some local CircleCI testing.

autoendpoint/src/server.rs
jrconlin: I'll make `StorageType::from_dsn` highly preferential to DynamoDb for now, and add a comment.

autoendpoint/src/server.rs
jrconlin: Ah, I had forgotten that this was a TODO. Let me resolve some of that with the code duplication we talked about. I'll add comments as well.

autopush/src/server/mod.rs
```diff
@@ -891,14 +907,41 @@ where
...
+            Ok(futures::Async::NotReady) => return Ok(futures::Async::NotReady),
+            Err(e) => {
+                error!(&e);
+                return Err(e);
+            }
+        }
+        */
         try_ready!(self.poll_complete());
-        Ok(self.inner.close()?)
+        // Actix4 will return an error on a clean close. This can throw off
```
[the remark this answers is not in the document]
jrconlin: Removed the comment since there are other mitigation methods in play.

autopush/src/server/mod.rs
jrconlin: huh, possibly a typo or cp error. Thanks!

tests/test_integration_all_rust.py
```diff
@@ -630,15 +673,9 @@ def setup_module():
 
     os.environ["RUST_LOG"] = RUST_LOG
     connection_binary = get_rust_binary_path("autopush_rs")
-    if not os.environ.get("SKIP_CONNECTION"):
```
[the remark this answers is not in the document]
jrconlin: Yep. Seemed silly to keep both.

autoconnect/autoconnect-settings/src/options.rs
```diff
@@ -0,0 +1,133 @@
...
+        let router_url = settings.router_url();
+        let endpoint_url = settings.endpoint_url();
+
+        let db_settings = DbSettings {
+            dsn: settings.db_dsn.clone(),
+            db_settings: settings.db_settings.clone(),
+        };
+        let db_client = match StorageType::from_dsn(&db_settings.dsn) {
+            StorageType::DynamoDb => Box::new(DdbClientImpl::new(metrics.clone(), &db_settings)?),
+            StorageType::INVALID => panic!("Invalid Storage type. Check DB_DSN."),
```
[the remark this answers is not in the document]
jrconlin: Good call. I'll have it compose the Panic message and change the other values from `AUTOPUSH_`

autoendpoint/src/server.rs
```diff
@@ -1,38 +1,38 @@
...
+        get_channels_route, new_channel_route, register_uaid_route, unregister_channel_route,
+        unregister_user_route, update_token_route,
+    },
+    webpush::{delete_notification_route, webpush_route},
+};
+use crate::settings::Settings;
 
 #[derive(Clone)]
-pub struct ServerState {
+pub struct ServerOptions {
```
pjenvey, line 30: nit: I know the goal of this rename is to closer match autoconnect but IMHO I don't think `ServerOptions` is the right name for this (I think we originally got `ServerState` from actix-web docs, looks like they now commonly call this kind of thing `AppState` as there's one per every `App` instance created). E.g. a `reqwest` client isn't really an "option".

Also looking at autconnect-settings, I think ideally we would have only a `Settings` type without the second `ServerOptions` for deserializing things from `Settings` -- remember merino accomplished this with that fancy serde library to deserialize on the fly directly into its Settings objects: https://github.com/mozilla-services/merino/blob/main/merino-settings/src/lib.rs#L200

autopush-common/src/db/dynamodb/mod.rs
pjenvey: ..ok this is basically just a move from autoendpoint anyway that git didn't pickup.

autoendpoint/src/settings.rs
```diff
@@ -46,13 +52,15 @@ impl Default for Settings {
...
+            db_settings: "".to_owned(),
             router_table_name: "router".to_string(),
             message_table_name: "message".to_string(),
             /// max data is a bit hard to figure out, due to encryption. Using something
             /// like pywebpush, if you encode a block of 4096 bytes, you'll get a
             /// 4216 byte data block. Since we're going to be receiving this, we have to
-            /// presume base64 encoding, so we can bump things up to 5624 bytes max.
-            max_data_bytes: 5624,
+            /// presume base64 encoding, so we can bump things up to 5630 bytes max.
+            max_data_bytes: 5630,
```
pjenvey, line 63: Why's this changing in this PR?

pjenvey (commented):
Not going to bikeshed block this on the `ServerOptions` naming but I'm wondering why the `max_data_bytes` change here.

autoendpoint/src/settings.rs
pjenvey: For reference: bumping this because some envs seem to be exceeding the 5624 value somehow in the integration test's `test_big_message`.

pjenvey (approved):

PR #358, opened 2023-03-28
feat: reduce the number of errors reported to sentry.io

jrconlin:
Sentry only really wants actionable errors. We're currently logging a lot of inactionable ones which may cause some critical errors to be unaccounted for.

Closes #SYNC-3635


autoendpoint/src/error.rs
```diff
@@ -203,21 +212,32 @@ impl ApiErrorKind {
...
-            | ApiErrorKind::InvalidAuthentication
-            | ApiErrorKind::InvalidLocalAuth(_) |
-            // Ignore missing or invalid user errors
-            ApiErrorKind::NoUser | ApiErrorKind::NoSubscription |
-            // Ignore overflow errors
-            ApiErrorKind::Router(RouterError::TooMuchData(_)),
-        )
+        match self {
+            // ignore selected validation errors.
+            ApiErrorKind::Validation(msg) => !msg.to_string().contains(TOPIC_LENGTH_ERR),
```
pjenvey, line 217: It looks like we only have 3 distinct validators/`ValidationErrors` currently -- I see at least one of the others (Topic must be URL and Filename safe Base64 alphabet) also showing up a lot in [this event](https://mozilla.sentry.io/issues/3925602807/events/?project=6326250&referrer=issue-stream&statsPeriod=14d&stream_index=0) -- so maybe just ignore them all for now?

If we care more about these in the future we could always improve the metric labels to be more specific than just `"validation"` kind of like syncstorage does.

autoendpoint/src/routers/mod.rs
```diff
@@ -169,8 +169,27 @@ impl From<RouterError> for ApcErrorKind {
         match err {
             RouterError::Adm(e) => ApcErrorKind::EndpointError("Router:Adm", e.to_string()),
             RouterError::Apns(e) => ApcErrorKind::EndpointError("Router:APNS", e.to_string()),
-            RouterError::Fcm(e) => ApcErrorKind::EndpointError("Router:FCM", e.to_string()),
-            RouterError::TooMuchData(e) => ApcErrorKind::EndpointError("TooMucData", e.to_string()),
+            RouterError::Fcm(e) => {
```
pjenvey, line 172: This `From` impl currently isn't being used.

I'm still of the opinion that autoendpoint should have its own high level error type that's more specific to its crate. So IMO `ApcError` should only be used by autoendpoint as Errors triggered from the common crate (that are probably ultimately wrapped by one of its own errors) and thus converting from the inverse (endpoint specific to common) like this wouldn't be needed.

autopush-common/src/errors.rs
```diff
@@ -155,6 +155,8 @@ pub enum ApcErrorKind {
 
     #[error("Endpoint Error: [{0}] {1}")]
     EndpointError(&'static str, String),
+    #[error("Upstream Error: [{0}] {1}")]
+    UpstreamError(String, String),
```
pjenvey, line 159: Echoing my other comment I don't think this is needed since this is endpoint specific and not triggered by any common code.

pjenvey (changes requested):

autopush/src/server/mod.rs
```diff
@@ -579,7 +579,7 @@ impl Future for MegaphoneUpdater {
                             error!("📢Failed to get response, queue again {error:?}");
                             capture_message(
                                 &format!("Failed to get response, queue again {error:?}"),
-                                sentry::Level::Error,
+                                sentry::Level::Warning,
```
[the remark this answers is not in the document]
jrconlin: Yeah, i'll fix the slog macro to say `warn!`. I thought that there was inbound filtering that can be done on the sentry side to only accept certain levels? I was half thinking that we could use that to do some "on-the-fly" filtering if needed.

autoendpoint/src/error.rs
```diff
@@ -169,6 +173,11 @@ impl ApiErrorKind {
     pub fn metric_label(&self) -> Option<&'static str> {
         Some(match self {
             ApiErrorKind::PayloadError(_) => "payload_error",
+            ApiErrorKind::Router(RouterError::Fcm(FcmError::InvalidAppId(_)))
```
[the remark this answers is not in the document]
jrconlin: Agreed.

autoendpoint/src/routers/mod.rs
jrconlin: That's fine, I'll remove the error bridge stuff. I still would prefer to have a single error class to avoid some duplication, but I'm not going to die on that hill. 

I'm not going to do it in this PR, but we ought to change `ApiError` to be `ApeError` to mimic `ApcError`. `ApiError` is legacy as well and no longer refers to anything meaningful.

autoendpoint/src/routers/mod.rs
```diff
@@ -162,33 +161,37 @@ impl RouterError {
...
+        if !err.is_empty() {
+            return Some(err);
+        }
+        None
+    }
+
+    pub fn is_sentry_event(&self) -> bool {
+        match self {
+            RouterError::Adm(e) => !matches!(e, AdmError::InvalidProfile | AdmError::NoProfile),
+            RouterError::Apns(ApnsError::SizeLimit(_)) => false,
```
pjenvey, line 191: Did you want `ApnsError::Unregistered` in here too (it has a metric label)?

autoendpoint/src/extractors/notification_headers.rs
```diff
@@ -29,6 +29,8 @@ pub struct NotificationHeaders {
     #[validate(
         length(
             max = 32,
+            // can't refer directly because macro
+            // message = crate::error::TOPIC_LENGTH_ERR
```
pjenvey, line 33: This var went away in the last change
```suggestion
```

pjenvey (changes requested):

pjenvey (approved):

PR #359, opened 2023-03-29
refactor: reduce settings duplication w/ deserialize_with

pjenvey:
also kill the unused router ssl settings

refactor: options -> app_state
refactor: more db_client -> db

queued against #357


autoconnect/autoconnect-settings/src/app_state.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// The connected WebSocket clients
+    pub registry: Arc<ClientRegistry>,
+
+    pub settings: Arc<Settings>,
+    pub router_url: String,
+    pub endpoint_url: String,
+}
+
+impl AppState {
+    pub fn from_settings(settings: Settings) -> Result<Self> {
```
jrconlin, line 32: I keep wondering if this should be a `TryFrom`

autoconnect/autoconnect-settings/src/lib.rs
```diff
@@ -149,15 +149,13 @@ impl Settings {
...
+        Ok(s)
     }
 
     pub fn router_url(&self) -> String {
-        let router_scheme = if self.router_ssl_key.is_none() {
-            "http"
-        } else {
-            "https"
-        };
+        let router_scheme = "http";
```
jrconlin, line 158: ?? We're not going to talk to the routers via TLS?

jrconlin: https://mozilla-hub.atlassian.net/browse/SYNC-3676

jrconlin (dismissed):

pjenvey:
Sorry I force pushed this to rebase against master but also added one minor change -- no longer wrapping `AppState`'s `Settings` in an `Arc`

autoconnect/autoconnect-settings/src/app_state.rs
pjenvey: Same -- I found [this comment](https://www.reddit.com/r/rust/comments/k7z66j/comment/gev84pr/?utm_source=share&utm_medium=web2x&context=3) with a pretty reasonable take on this, looking at the intent (conversion vs construction): "for instance I can construct a db connection from a string saying that I can convert a string to a db connection would be… odd". `AppState` and e.g. `StorageType::from_dsn` that it calls are definitely  construction.

jrconlin (approved):

PR #362, opened 2023-04-04
feat: switch to latest release a2 library

jrconlin:
* This switches from our semi-vendored a2, back to the official release.
This should also clear the warning about the included `nom` library.

*Breaking Change*
This removes the unused `.aps` field from the router_data options. This
field was only used during testing. That test was also removed.


autoendpoint/src/routers/apns/router.rs
```diff
@@ -218,14 +199,6 @@ impl Router for ApnsRouter {
             serde_json::to_value(app_id).unwrap(),
         );
 
-        if let Some(aps) = &router_input.aps {
```
pjenvey, line 221: It doesn't look like aps was solely used for "testing" (at least unit testing). It comes in from `RouterDataInput` from new `registration`/`registration updates`. [firefox-ios may pass one in](https://github.com/mozilla-mobile/firefox-ios/blob/main/Push/PushClient.swift#L53) when its `experimentalMode` (is that dev mode?) is enabled.

It seems like it was ultimately added for dev use?
https://bugzilla.mozilla.org/show_bug.cgi?id=1364403
https://github.com/mozilla-services/autopush/issues/910#issuecomment-304953536

So we probably want to keep it or at least check w/ the ios team before removing it?

pjenvey:
We've identified some missing parts of Mark's original a2 branch that didn't get merged to a2 master likely due to merge conflicts. E.g. in [this PR](https://github.com/WalletConnect/a2/pull/64) the [Client::certificate_parts in here](https://github.com/mozilla-services/a2/commit/cad603059288477da5014f08a29ff7354e1e9cc4) wasn't added. So we should double check that all of Mark's original branch made it into a2 before moving off of it.

hackebrot:
Hi @jrconlin! 👋🏻 

I see you've added test eng as reviewers on this pull request, but neither @Trinaa or myself have any context about this project. Did you have any specific questions that we can help with?

hackebrot:
No worries, @jrconlin! Thanks. 👍🏻

jrconlin:
I compared the current version of mozilla-services/a2 against [Mark's PR](https://github.com/mozilla-services/a2/pull/3) and looks like the only thing missing was the `Client::certificate_parts` method. I added it to [this PR](https://github.com/mozilla-services/a2/pull/4/files). 

Once that PR lands, I'll redo this PR to include the new 0.8.0 version, and then submit a PR to `wallet/a2` to include the change as well. That should let us be able to release the new server with the new library.

jrconlin:
Switched to using our `mozilla-services/a2` git repo while we wait for https://github.com/WalletConnect/a2/pull/72 to land. 

Part of the delay was trying to debug the way that code wants to read in key/cert from the environment. I couldn't get the escape quite right. I would frequently get a Private Key error in the cert. Passing 
```
"AUTOEND__APNS__CHANNELS": "{\"dev\":{\"cert\": \"keys/test.crt\", \"key\":\"keys/test.key\"}}",
```
worked, even with the same values, so guessing all the coding/recording might have changed a value.

pjenvey:
> Part of the delay was trying to debug the way that code wants to read in key/cert from the environment. I couldn't get the escape quite right. I would frequently get a Private Key error in the cert. Passing
> 
> ```
> "AUTOEND__APNS__CHANNELS": "{\"dev\":{\"cert\": \"keys/test.crt\", \"key\":\"keys/test.key\"}}",
> ```

Maybe that's more of a shell problem, assuming you're running with `AUTOEND__APNS__CHANNELS="{\"dev\":{\"cert\": \"keys/test.crt\", \"key\":\"keys/test.key\"}}"`? With the outer double quotes conflicting with the inner ones

I just tried running it surrounded in single quotes instead (preserving everything inside) and it worked as I'd expect:

```
AUTOEND__APNS__CHANNELS='{"dev":{"cert": "keys/test.crt", "key":"keys/test.key"}}' AUTOEND__HUMAN_LOGS=true RUST_BACKTRACE=1 cargo run
...
May 25 23:48:08.863 DEBG Starting up...
settings.apns: ApnsSettings {
    channels: "{\"dev\":{\"cert\": \"keys/test.crt\", \"key\":\"keys/test.key\"}}",
    max_data: 4096,
}
May 25 23:48:08.864 DEBG initialized disabled sentry client due to disabled or invalid DSN
settings.apns.channels(): Ok(
    {
        "dev": ApnsChannel {
            cert: "keys/test.crt",
            key: "keys/test.key",
            topic: None,
            sandbox: false,
        },
    },
)
```

jrconlin:
Oh, it's absolutely a vscode problem, and I was being stubborn.

<rant>
In VS Code, you specify your run args as a JSON structure. This includes a set of environment variables specified as key values, that you have to properly escape so that the JSON to env and then the env to JSON parser can handle, so you wind up with stuff like:
```json
        {
            "type": "lldb",
            … ,
            "cwd": "${workspaceFolder}",
            "env": {
                "RUST_LOG": "autopush=DEBUG,autoconnect=DEBUG,autopush_common=DEBUG,autoendpoint=DEBUG,warn",
                "AUTOEND__PORT": "9160",
                //"AUTOEND__BIND_HOST": "0.0.0.0",
                "AUTOEND__HUMAN_LOGS": "1",
                "AUTOEND__CRYPTO_KEYS": "[…]",
                "AUTOEND__APNS__CHANNELS": "{\"dev\":{\"cert\":\"-----BEGIN CERTIFICATE-----\\n…=\\n-----END CERTIFICATE-----\", \"key\": \"-----BEGIN PRIVATE KEY-----\\n…----END PRIVATE KEY-----\"}}",
            }
        },
```
So yeah, LOTS of opportunities to screw things up, because control characters in JSON strings are big no-nos.

autoendpoint/Cargo.toml
```diff
@@ -45,14 +45,8 @@ tokio.workspace = true
...
 
-# Using a fork temporarily until these three PRs are merged:
-# - https://github.com/pimeys/a2/pull/49
-# - https://github.com/pimeys/a2/pull/48
-# - https://github.com/pimeys/a2/pull/47
-# The `autoendpoint` branch merges these three PRs together.
-# The version of a2 at the time of the fork is v0.5.3.
-a2 = { git = "https://github.com/mozilla-services/a2.git", branch = "autoendpoint" }
-# several of these libraries are pinned due to https://github.com/mozilla-services/autopush-rs/issues/249
+a2 = {version = "0.8", git = "https://github.com/mozilla-services/a2"}
```
pjenvey, line 48: We should probably pin this to a specific revision (even though we own this repo) -- actually, upstream quickly merged your pr 🎉 so let's pin to their repo instead.

autoendpoint/src/routers/apns/router.rs
```diff
@@ -251,15 +302,13 @@ impl Router for ApnsRouter {
             .get("rel_channel")
             .and_then(Value::as_str)
             .ok_or(ApnsError::NoReleaseChannel)?;
+        // XXX: We don't really use anything that is a numeric here, aside from
```
pjenvey, line 305: Not really an XXX: (broken/fixmeish) comment IMO

```suggestion
        // NOTE: We don't really use anything that is a numeric here, aside from
```

autoendpoint/src/routers/apns/router.rs
```diff
@@ -268,21 +317,136 @@ impl Router for ApnsRouter {
...
             .ok_or(ApnsError::InvalidReleaseChannel)?;
-        let payload = Payload {
-            aps,
-            data: message_data
-                .into_iter()
-                .map(|(k, v)| (k, Value::String(v)))
-                .collect(),
-            device_token: token,
-            options: NotificationOptions {
+        let mut aps = Self::default_aps();
```
pjenvey, line 320: Sorry, I probably missed some this the last time I reviewed it (is James back from PTO yet, can we just kill all this? 😄 ) -- looking at it now, I think we can at least move this large chunk into its own method, I've taken a [shot at that on this branch](https://github.com/mozilla-services/autopush-rs/compare/chore/main-a2...chore/main-a2-move-to-method).

pjenvey (changes requested):
Couple things:

- I forgot what we concluded when we last discussed this, but is there a way to add a test that ensures we fixed the panic? Can we just shoot a message to `Endpoint::Sandbox`?
- probably want to rename the branch to `feat/` at this point

autoendpoint/src/routers/apns/router.rs
jrconlin: I mean, sure? It's still a big ugly pile of code regardless of if we want to screw with the stack or not. Tossing it into a different function doesn't really help, and IDEs allow for block collapse if we just want to ignore it for now.

autoendpoint/Cargo.toml
jrconlin: ?? https://github.com/WalletConnect/a2/pull/72 is still pending review.

I'll pin to a specific commit, but will still keep the version, since I ought to do a release.

pjenvey: My fault I must have been looking at the wrong repo

autoendpoint/src/routers/apns/router.rs
jrconlin: Actually... It is a FIXME, specifically the third line:

```
// Once we're off of DynamoDB, we might want to kill the map.
```
That's a note to future us that we probably want to recheck if we need to do this, right?

autoendpoint/src/routers/apns/router.rs
pjenvey: It's just too many lines for one method, this block is > 100 by itself (clippy's pedantic `too_many_lines` lint is that). I assumed it was in here due to lifetime trickyness, hence my branch to solve that w/ the new method. I wouldn't justify e.g. an even larger say 500 line method either with IDE collapsing help -- so I'll argue this does help readability.

jrconlin: Yeah, fair. I broke it out into it's own function like you suggested. (changed the name and gave it a better comment) I was debating how much of that I could reduce using macros, but I figured that was more work for what was (hopefully) a temporary thing.

autoendpoint/src/routers/apns/router.rs
pjenvey: Ah, could do FIXME but XXX works

autoendpoint/Cargo.toml
```diff
@@ -45,7 +45,7 @@ tokio.workspace = true
 url.workspace = true
 uuid.workspace = true
 
-a2 = {version = "0.8", git = "https://github.com/mozilla-services/a2"}
+a2 = {version = "0.8", git = "https://github.com/mozilla-services/a2", branch="master"}
```
pjenvey, line 48: I was thinking more of a specific sha (thinking of reproducable builds) but then again we weren't doing this before by pointing to the autoendpoint branch, so never mind I guess.

pjenvey (dismissed):

pjenvey (approved):

PR #372, opened 2023-04-26
bug: Add back missing metrics

jrconlin:
A number of metrics were accidentally dropped during the code merge and refactor.

### DIFF
--- a/autoendpoint/src/routes/webpush.rs
+++ b/autoendpoint/src/routes/webpush.rs
@@ -7,13 +7,12 @@ use crate::extractors::routers::{RouterType, Routers};
 use crate::server::AppState;
 use actix_web::web::Data;
 use actix_web::HttpResponse;
-use cadence::CountedExt;
 
 /// Handle the `POST /wpush/{api_version}/{token}` and `POST /wpush/{token}` routes
 pub async fn webpush_route(
     notification: Notification,
     routers: Routers,
-    app_state: Data<AppState>,
+    _app_state: Data<AppState>,
 ) -> ApiResult<HttpResponse> {
     // TODO:
     sentry::configure_scope(|scope| {
--- a/autoendpoint/src/routes/webpush.rs
+++ b/autoendpoint/src/routes/webpush.rs
@@ -7,12 +7,13 @@ use crate::extractors::routers::{RouterType, Routers};
 use crate::server::AppState;
 use actix_web::web::Data;
 use actix_web::HttpResponse;
+use cadence::CountedExt;
 
 /// Handle the `POST /wpush/{api_version}/{token}` and `POST /wpush/{token}` routes
 pub async fn webpush_route(
     notification: Notification,
     routers: Routers,
-    _app_state: Data<AppState>,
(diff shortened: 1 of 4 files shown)


autoendpoint/src/routes/webpush.rs
```diff
@@ -25,7 +26,9 @@ pub async fn webpush_route(
         RouterType::from_str(&notification.subscription.user.router_type)
             .map_err(|_| ApiErrorKind::InvalidRouterType)?,
     );
-
+    if notification.has_topic() {
+        app_state.metrics.incr("ua.notification.topic")?;
+    }
```
pjenvey, line 31: We don't want this one, with the `ua.notification` prefix it was previously emitted by autoconnect. We consolidated it into `ua.notification.sent` in https://github.com/mozilla-services/autopush-rs/pull/330 (so I just struck it from the spreadsheet)

```suggestion
```

autopush-common/src/db/dynamodb/mod.rs
```diff
@@ -524,6 +525,13 @@ impl DbClient for DdbClientImpl {
             )
             .await?;
 
+        let metric = self.metrics.incr_with_tags("notification.message.stored");
+        if let Some(topic) = topic {
+            metric.with_tag("topic", topic.as_str()).send();
+        } else {
+            metric.send();
+        }
+
```
pjenvey, line 534: You want to render true/false like the autoconnect code does,
```suggestion
        self.metrics.incr_with_tags("notification.message.stored")
            .with_tag("topic", message.topic.is_some().to_string()).send()
 ```

File: autopush-common/src/db/dynamodb/mod.rs [hunk 3]
```diff
@@ -543,6 +551,9 @@ impl DbClient for DdbClientImpl {
                 retryable_delete_error(self.metrics.clone()),
             )
             .await?;
+        self.metrics
+            .incr_with_tags("notification.message.removed")
```
pjenvey, line 555: We call it `deleted` in the other impl, unless you want to rename it:
```suggestion
            .incr_with_tags("notification.message.deleted")
```

pjenvey (changes requested):

autopush-common/src/db/dynamodb/mod.rs
jrconlin: updated the spreadsheet to include `notification.message.deleted` 
thanks!

autopush-common/src/db/dynamodb/mod.rs
```diff
@@ -511,6 +511,7 @@ impl DbClient for DdbClientImpl {
     }
 
     async fn save_message(&self, uaid: &Uuid, message: Notification) -> DbResult<()> {
+        let topic = message.topic.clone().is_some().to_string();
```
pjenvey, line 514: nit:
```suggestion
        let topic = message.topic.is_some().to_string();
```

pjenvey (changes requested):

pjenvey (approved):

PR #379, opened 2023-05-17
feat: complete (mostly) the WebPushClient

pjenvey:
SYNC-3688


autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -102,6 +100,58 @@ impl UnidentifiedClient {
...
+                    user.set_last_connect();
+                    self.app_state.db.update_user(&user).await?;
+                    return Ok(GetOrCreateUser {
+                        user,
+                        existing_user: true,
+                        flags,
+                    });
+                }
+            }
+            // NOTE: when the client's specified a uaid but get_user returns
```
pjenvey, line 128: Wanted to point this out as one difference between the other state machine. Previously we only set the `defer_add_user` (previously called `deferred_registration`) for when clients did not specify a uaid.

Now we're doing it for any case where we would have created a brand new user record -- including when a user specifies a `uaid` and its lookup failed or its record was invalid.

pjenvey:
Added an integration test run against this to ci

autoconnect/autoconnect-common/src/test_support.rs
```diff
@@ -15,20 +15,37 @@ pub const HELLO: &str = r#"{"messageType": "hello", "use_webpush": true}"#;
 pub const HELLO_AGAIN: &str = r#"{"messageType": "hello", "use_webpush": true,
                                   "uaid": "deadbeef-0000-0000-deca-fbad00000000"}"#;
 
+pub const CURRENT_MONTH: &str = "message_2023_5";
```
jrconlin, line 18: I'm tempted to suggest that we either use the current, production value here (`message_2017_09` IIRC), or remove the table rotation code entirely since it's no longer useful and just adds complications.

autopush-common/src/db/dynamodb/mod.rs
```diff
@@ -510,6 +511,35 @@ impl DbClient for DdbClientImpl {
         })
     }
 
+    async fn increment_storage(&self, uaid: &Uuid, timestamp: u64) -> DbResult<()> {
```
jrconlin, line 514: TODO: jr to add `increment_storage` and `save_messages` to #364

jrconlin (approved):
minor notes, but nothing I'm going to block over.

Thanks!

autoconnect/autoconnect-common/src/test_support.rs
pjenvey: Changed this test value to match prod, of course we can't hardcode the result of `Db::message_month` to such a value though (in case you're suggesting that which you probably aren't).

pjenvey: With the metrics we saw yesterday I'll definitely remove the table migration code in a separate PR (mostly to document it in isolation). I think it will end up having us match autoendpoint's `validate_webpush_user` even more closely.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
```diff
@@ -82,35 +98,140 @@ impl WebPushClient {
...
+///
+/// Somewhat similar to autoendpoint's `validate_webpush_user` function. When a
+/// User record is invalid it will be dropped from the db and `None` will be
+/// returned.
+pub async fn process_existing_user(
+    app_state: &Arc<AppState>,
+    user: &User,
+) -> DbResult<Option<ClientFlags>> {
+    if user.current_month.is_none() {
+        app_state
```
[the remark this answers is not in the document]
pjenvey: Per our discussion, just to clarify: we aren't hitting `current_month.is_none()` We do see similar metrics but they're for legacy autoconnect dropping users due to their message month being outside our list of "valid" message tables.

We can see we still have more than one message month table (probably 1 or 2 behind prod's current month) because with users still in those months because they're continuing to trigger the message table rotation logic on legacy autoconnect.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -124,78 +127,161 @@ impl WebPushClient {
...
-    async fn nack(&mut self, _code: Option<i32>) -> Result<(), SMError> {
-        unimplemented!();
+    /// Negative Acknowledgement (a Client error occurred) of one or more Push
+    /// Notifications
+    fn nack(&mut self, code: Option<i32>) {
+        trace!("WebPushClient:nack");
+        // only metric codes expected from the client (or 0)
+        let code = code
+            .and_then(|code| (301..=303).contains(&code).then_some(code))
+            .unwrap_or(0);
```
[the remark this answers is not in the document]
pjenvey: I'd like to punt on this for now, it could be a separate issue (though it's not terribly important) just because I have no idea what the client might do if it begins trigger it.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
```diff
@@ -1,39 +1,292 @@
...
+    /// Read a chunk (max count 10 returned) of Notifications from storage
+    async fn do_check_storage(&self) -> Result<CheckStorageResponse, SMError> {
+        trace!("🗄️ WebPushClient::do_check_storage");
+        let timestamp = self.ack_state.unacked_stored_highest;
+        let resp = if self.flags.include_topic {
+            trace!("🗄️ WebPushClient::do_check_storage: fetch_messages");
+            self.app_state.db.fetch_messages(&self.uaid, 11).await?
+        } else {
+            Default::default()
+        };
```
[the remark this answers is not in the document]
pjenvey: I definitely meant to document the guts of check_storage more, I just added that as a TODO for now, but renamed the resp vars.

jrconlin (approved):

jrconlin:
FWIW: I pre-emptively merged this PR into #364. One thing that I spotted was that I don't know if we actually use the user `timestamp` field. (partly because of the way that BigTable keys work.) I'll dig into that a bit more next week.

pjenvey:
Which `timestamp`? `db::User` doesn't have one

jrconlin:
Sorry, the one we write to the database in the User meta record for DynamoDB. It's the field updated in `increment_storage`.

PR #381, opened 2023-05-25
feat: add broadcast (megaphone) support

pjenvey:
and setup logging for tests that's properly captured by cargo test

Closes SYNC-3689


autoconnect/autoconnect-ws/src/ping.rs
```diff
@@ -0,0 +1,109 @@
...
+#[derive(Debug)]
+pub struct PingManager {
+    /// Waiting to Ping or timeout recieving a Pong
+    waiting: Waiting,
+    ping_or_timeout: Interval,
+}
+
+impl PingManager {
+    pub async fn new(settings: &Settings) -> PingManager {
+        let mut ping = interval(settings.auto_ping_interval);
```
jrconlin, line 32: It feels a bit confusing to thave basically two different `ticks()` here.
I understand what they are, but a casual / fast read kind of blurs them.

I wonder if it might be useful to call this one `ping_clock`, since it's an `Interval`?

jrconlin (changes requested):

autoconnect/autoconnect-ws/src/ping.rs
pjenvey: Is this any clearer?

jrconlin (approved):

PR #389, opened 2023-06-13
feat: remove legacy table rotation

pjenvey:
and kill no longer used hello, etc and prototype client.rs

SYNC-3741


jrconlin (commented):
So, we're dropping users with bad `current_month` values, but ultimately keeping the value rather than drop the `current_month` value. We still do operations on that value. (e.g. the stuff we do in `autopush-common::db::dynamodb::DdbClientImpl::hello`)

(Also the commit message is [SYNC-3635](https://mozilla-hub.atlassian.net/browse/SYNC-3635) which appears to be unrelated to this PR.)

pjenvey:
> So, we're dropping users with bad `current_month` values, but ultimately keeping the value rather than drop the `current_month` value. We still do operations on that value. (e.g. the stuff we do in `autopush-common::db::dynamodb::DdbClientImpl::hello`)

For now, yes, we continue dropping when there's no `current_month` (very very legacy) or it doesn't match. My thought was to defer tweaking this further until BigTable lands. Looks like you have a `current_month` field in BigTable but I don't think it's necessary to have -- maybe the `Db::message_table()` would return an Option w/ BigTable returning `None` -- as it was really only used for table rotation (or dropping legacy users here).

`hello` is now unused, I was going to kill it later but I can do so in this PR if you want.

jrconlin (approved):

PR #394, opened 2023-06-29
fix: remove mozsvc-common

pjenvey:
its older get_ec2_instance_id isn't compat w/ modern tokio (and the crate's deprecated)

SYNC-3449


pjenvey:
This should fix that `thread 'main' panicked at 'Cannot drop a runtime in a context where blocking is not allowed.` error on startup when not using `human_logs=true`

PR #402, opened 2023-07-17
feat: Topic messages shouldn't have sortkey_timestamps

pjenvey:
- fix autoconnect's db initialization to match autoendpoint's (don't panic when no DB_DSN is set)

- don't unnecessarily remove expired messages in a background task during check storage

SYNC-3777
SYNC-3809

### DIFF
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
@@ -1,4 +1,3 @@
-use actix_web::rt;
 use cadence::{Counted, CountedExt};
 
 use autoconnect_common::protocol::{ServerMessage, ServerNotification};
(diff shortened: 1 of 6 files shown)


pjenvey:
FYI @jrconlin I put the naming change in a different branch against feat/big_table https://github.com/mozilla-services/autopush-rs/compare/feat/big_table...feat/big_table-chid_msgid

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
```diff
@@ -116,16 +115,24 @@ impl WebPushClient {
...
             }
-            // TODO: A batch remove_messages would be nicer
-            if n.sortkey_timestamp.is_none() {
-                self.spawn_remove_message(n.sort_key());
+            if msg.sortkey_timestamp.is_none() {
+                expired_topic_sort_keys.push(msg.sort_key());
             }
             false
         });
+        // TODO: A batch remove_messages would be nicer
```
jrconlin, line 129: technically, we should use the Topic Family to say that only one version of this should be stored or returned, but we can work on that later.

jrconlin (approved):

PR #407, opened 2023-07-24
Added comments to load test locustfile.py

jrconlin:
Added some comments per discussion with @Trinaa.

I marked things that are curious or inexplicable with "NOTE:" 

Please feel free to either ignore, improve, or whatever.


tests/load/locustfile.py
```diff
@@ -45,7 +65,22 @@ class ConnectionTaskSet(TaskSet):
 
     @task
     def test_basic(self):
+        """ Perform a "basic" transaction test.
+
+        Desktop Autopush clients use a websocket connection to exchange
+        JSON command and response messages. (See
+        [Autopush HTTP Endpoints for Notifications](https://mozilla-services.github.io/autopush-rs/http.html#push-service-http-api)
+        for details).
```
jrbenny35, line 73: Will this render in the file? This looks like a markdown style link.

tests/load/locustfile.py
```diff
@@ -96,23 +151,57 @@ def test_basic(self):
...
+            # NOTE: We are disconnecting from the Autoconnect server so that
+            # we can potentially send multiple topic messages. If we were still
+            # connected the server would deliver the messages as it received them.
+            # That is to be expected.
+
+        # NOTE: I believe that this is cruft from an earlier system. This condition
+        # should just be replaced with
+        # ```
+        # endpoint_url = res["pushEndpoint"]
+        # ```
```
jrbenny35, line 204: Do you think we need these notes here? Or would it be better off to track these externally so when we fix them, we won't have to edit this file again.

jrbenny35 (commented):
Hey JR, thanks for adding these comments! I think there are some comments that may be best held for a future update, I think we should file an issue for enhancing this in the short term as we will be looking to expand this suite in the next month.

tests/load/locustfile.py
jrconlin: I think it depends on whatever you're using to render the docs. I believe that Sphinx has a [markdown](https://www.sphinx-doc.org/en/master/usage/markdown.html) plug in. 

I tend to use markdown because I find it easier to read than RST.

tests/load/locustfile.py
jrconlin: I'm fine either way. I wanted to capture things like this in the code for now since I was going through it.

jrbenny35: I had a chat with Kat, I think I didn't understand the context of this PR. Thank you for going through and explaining all the steps, I learned some stuff for sure!

Trinaa:
@jrbenny35 I'm thinking to merge this PR so that the context is available in the document history? When we refactor we can edit/add/remove them accordingly.

pjenvey (commented):
Just an optional suggestion, LGTM otherwise

tests/load/locustfile.py
```diff
@@ -132,7 +227,12 @@ def test_basic_topic(self):
         assert (
             endpoint_res.status_code == 201
         ), f"status code was {endpoint_res.status_code}"
-        # connect and check notification
+
+        # NOTE: To properly test "topic" messages, we really ought to
```
pjenvey, line 231: This almost sounds like a TODO but `test_connect_stored` does this. Maybe reword or maybe mention it here?

jrconlin: `test_connect_stored` kind of does that.

What it does is send in 10 topic messages, then read from the queue 10 times looking to see if the response matches. Since the response is never ACK'd, it's never removed, and all 10 work correctly, but it feels like that's a bug rather than a feature.

pjenvey: Ah, my mistake

pjenvey (approved):

PR #417, opened 2023-08-03
feat: add VAPID quality tracking metric messages

jrconlin:
Closes: SYNC-3842


autoendpoint/src/extractors/subscription.rs
```diff
@@ -298,7 +326,12 @@ fn validate_vapid_jwt(
                 return Err(VapidError::InvalidVapid(e.to_string()).into());
             }
             _ => {
-                metrics.clone().incr("notification.auth.bad_vapid.other");
+                let mut tags = Tags::default();
+                tags.tags.insert("error".to_owned(), e.to_string());
```
pjenvey, line 330: This isn't a suitable tag value, lots of their [error variants produce strings with whitespace](https://docs.rs/jsonwebtoken/latest/src/jsonwebtoken/errors.rs.html#110)

autoendpoint/src/extractors/subscription.rs
```diff
@@ -93,7 +93,29 @@ impl FromRequest for Subscription {
...
+                        tags.tags
+                            .insert("error".to_owned(), e.as_metric().to_owned());
+                        metrics
+                            .clone()
+                            .incr_with_tags("notification.auth.error", Some(tags));
+                    })
+                    .unwrap_or_default();
+                // For now, record that we had a good (?) VAPID sub,
+                metrics.clone().incr("notification.auth.ok");
+                info!("VAPID sub: {:?}", sub)
```
pjenvey, line 117: A few things here:

- The lack of `sub` isn't really [an actual error](https://datatracker.ietf.org/doc/html/rfc8292#section-2.1), maybe call this metric something else? Could it be a tag on the existing `notification.auth` metric (moving all this into `parse_vapid`)?
- If it doesn't make sense as part of `parse_vapid` I think ideally this extra work for metrics shouldn't occur until `validate_vapid_jwt` succeeds in validating
- Is the new `notification.auth.ok` totally necessary? We emit `updates.vapid.draft{version}` for successfully validated vapid, we could get a similar number by subtracting the no sub metric from those 2 combined

autoendpoint/src/headers/vapid.rs
```diff
@@ -75,6 +76,29 @@ impl VapidHeader {
...
+
+    pub fn sub(&self) -> Result<String, VapidError> {
+        let data: HashMap<String, Value> = serde_json::from_str(&self.token).map_err(|e| {
+            warn!("🔐 Vapid: {:?}", e);
+            VapidError::SubInvalid
+        })?;
+
+        if let Some(sub_candiate) = data.get("sub") {
+            if let Some(sub) = sub_candiate.as_str() {
+                if !sub.starts_with("mailto:") {
```
pjenvey, line 88: Apparently `https` is [also valid](https://datatracker.ietf.org/doc/html/rfc8292#section-2.1)?

pjenvey (changes requested):

autoendpoint/src/extractors/subscription.rs
```diff
@@ -93,7 +93,29 @@ impl FromRequest for Subscription {
                 .map(|vapid| extract_public_key(vapid, &token_info))
                 .transpose()?;
 
-            trace!("Vapid: {:?}", &vapid);
+            trace!("raw vapid: {:?}", &vapid);
+
+            // Capturing the vapid sub right now will cause too much cardinality. Instead,
+            // let's just capture if we have a valid VAPID, as well as what sort of bad sub
+            // values we get.
+            if let Some(header) = vapid.clone() {
```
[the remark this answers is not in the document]
jrconlin: Thanks. I keep forgetting you can `ref` those now.

autoendpoint/src/extractors/subscription.rs
```diff
@@ -308,7 +337,34 @@ fn validate_vapid_jwt(
...
             }
             _ => {
-                metrics.clone().incr("notification.auth.bad_vapid.other");
+                // Attempt to match up the majority of ErrorKind variants.
+                // The third-party errors all defer to the source, so we can
+                // use that to differentiate for actual errors.
+                let mut tags = Tags::default();
+                let label = if e.source().is_none() {
+                    // These two have the most cardinality, so we need to handle
+                    // them separately.
```
pjenvey, line 346: I don't understand this comment and the other mention of cardinality a few lines below -- there's some sense of cardinality across all these ErrorKind variant since there's almost 20 of them, but the concern with these 2 variants and others is that their `to_string` includes whitespace which aren't valid for tag values

autoendpoint/src/extractors/subscription.rs
```diff
@@ -308,7 +337,34 @@ fn validate_vapid_jwt(
...
+                        jsonwebtoken::errors::ErrorKind::InvalidRsaKey(_) => {
+                            "InvalidRsaKey".to_owned()
+                        }
+                        jsonwebtoken::errors::ErrorKind::MissingRequiredClaim(_) => {
+                            "MissingRequiredClaim".to_owned()
+                        }
+                        // NOTE: It's possible that any new error introduced could
+                        // be a source of cardinality. We should handle these as they
+                        // come up.
+                        _ => e.to_string(),
```
pjenvey, line 357: There's still a few other variants that produce whitespace here such as `ErrorKind::Json`, can that and others still occur?

autoendpoint/src/extractors/subscription.rs
jrconlin: I'm guessing that this probably refers to a different line, since the reference now points to where we extract the raw claims string from the VAPID token.

pjenvey (approved):

PR #419, opened 2023-08-07
feat: Update loadtests user and tests

jrbenny35:
This commit updates the user class in accordance with the planned changes for adding a test shaper. It also updates the tasks for simple connection testing.

### DIFF
--- a/tests/load/locustfile.py
+++ b/tests/load/locustfile.py
@@ -1,31 +1,14 @@
-import base64
 import json
-import random
-import string
 import time
 import uuid
-from contextlib import closing
-from urllib.parse import urljoin, urlparse
+from typing import Any
 
-from locust import HttpUser, TaskSet, events, task
+from locust import FastHttpUser, between, events, task
 from websocket import create_connection
-from websocket._exceptions import WebSocketTimeoutException
-
-"""
-History:
-These tests pre-date the initial production release of Autopush and touch on
-a number of predicted scenarios. They were originally crafted using no framework,
-then ported to a number of more 'artesinal' style frameworks.
-
-Autopush provides a passive update system called "Megaphone"/"Broadcast". A client
-may either have registered endpoints to receive push notifications (active), or
-may simply connect up to the push server to receive Broadcast updates (passive).
-
-"""
 
 
 @events.init_command_line_parser.add_listener
-def _(parser):
+def _(parser: Any):
     parser.add_argument(
         "--websocket_url",
         type=str,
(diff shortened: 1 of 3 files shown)


Trinaa (changes requested):
While exploring this PR I created a child [branch](https://github.com/mozilla-services/autopush-rs/tree/sync-3830-kat), I think comparing with it will give the best examples for the requirements and suggestions made bellow.

**Required**
- [ ] Fix duplicate Request statistics reporting -> The `TimeEvent` fires statistics to Locust before validating the contents of the received message. If a problem is found in the message an exception will be raised reporting a second conflicting statistic (mixed messages). Optionally we can catch exceptions for better statistic organization.
- [ ] Evaluate message response status codes are '200'
- [ ] Rename `User` class to `AutopushUser` in order to be distinct from the Locust `User` class
- [ ] Remove `@task` decorator from `hello` function
- [ ] Define data structures (`list` & `dict`) with stored types

**Suggested**
- [ ] Simplify architecture by not using 'TaskSet' -> The TaskSet calling functions within an inherited user and the user having visibility on the TaskSet via the `tasks` parameter, creates a circular logic that I believe will be hard to maintain.
- [ ] Verify message contents using Pydantic models (picked this trick up from @hackebrot 🤖)
- [ ] Include `ws.revc()` as part of the response_time calculation. Is this a better reflection of round trip time?
- [ ] Include `response_length`, which is the 'size (bytes)' in Locust's statistics
- [ ] Add function docstrings

**Notes**
- Should the user fail to connect, I'm not sure if we need to kill the user explicitly?
- During initialization of this solution a warning is emitted image
- When executing a timed run with this solution, it gets stuck in a 'CLEANUP' state image

Trinaa:
> > While exploring this PR I created a child [branch](https://github.com/mozilla-services/autopush-rs/tree/sync-3830-kat), I think comparing with it will give the best examples for the requirements and suggestions made bellow.
> 
> This makes it quite hard to discuss the proposed changes. It also looks like you completely changed almost all of my code there, I am not sure what to make of that.

I apologize there are a lot of ideas and I didn't know how to convey an architectural alternative another way. When we created the [class diagram](https://miro.com/app/board/uXjVMx-kx9Q=/), we discussed dropping TaskSets.

> > **Required**
> > ```
> > * [ ]  Fix duplicate Request statistics reporting -> The `TimeEvent` fires statistics to Locust before validating the contents of the received message. If a problem is found in the message an exception will be raised reporting a second conflicting statistic (mixed messages). Optionally we can catch exceptions for better statistic organization.
> > ```
> Can you give an example of where this would happen? From my understanding the contents of the message don't matter, as the integration tests should cover that the message is correct. I am asserting within the `TimeEvent` context that a message is received and only firing the locust event after it exits the context.
> 
An example where this would occur is if the assertions fail, this would send a mixed signal.
By default in Locust, when using the FastHttpUser `client`, requests with status codes not in the 200 range are automatically classified as 'Failure'. Since we are using web socket, we need to evaluate explicitly. 
I think verifying the message schema instead of just the `messageType` would be a low lift and give a valuable signal (but that's a suggestion). 

> > ```
> > * [ ]  Remove `@task` decorator from `hello` function
> > ```
> 
> Is `hello` not a task?

I didn't think so. My understanding from @jrconlin's notes was that it needs to be sent upon connection, but not repeated... did I misunderstand?

jrconlin:
No, @jrbenny35, you are correct.

A "hello" is an initialization task. It transitions a connection from an Unidentified state to an Identified state. Connections that are Unidentified are quickly dropped, where as Identified connections can persist as long as the client desires. 

"hello" messages should not be sent after this state transition. It would be like me greeting you in the Hello middle of a conversation.

Trinaa (dismissed):

jrconlin (dismissed):
I'm a bit surprised that we're not testing more, here, but I'm guessing tests for things like backlog message delivery on reconnection, unregistration, ping presence checks, and other things are going into a different set of functions or a different script later?

jrconlin:
@Trinaa Even a passive connection should get regular ping messages. This ensures that the connection is still "live", as we can't always trust the socket report. (We hit that problem before, where a socket would report being "open" until we tried sending traffic across it and then it would fail and report being closed.)

If we're doing load testing we should probably make sure that the connections we say are live actually are. 

(Not critical for this stage of testing, but something to keep in mind at least.)

Trinaa:
> @Trinaa Even a passive connection should get regular ping messages. This ensures that the connection is still "live", as we can't always trust the socket report. (We hit that problem before, where a socket would report being "open" until we tried sending traffic across it and then it would fail and report being closed.)
> 
> If we're doing load testing we should probably make sure that the connections we say are live actually are.
> 
> (Not critical for this stage of testing, but something to keep in mind at least.)

@jrconlin Okay. Added to the scope of [SYNC-3846](https://mozilla-hub.atlassian.net/browse/SYNC-3846) and the [class diagram](https://miro.com/app/board/uXjVMx-kx9Q=/). Follow-up questions will probably be coming 😸

tests/load/locustfile.py
```diff
@@ -42,578 +25,99 @@ def _(parser):
...
-                header={"Origin": "http://localhost:1337"},
-                ssl=False,
+        with self._time_event(name="register"):
+            body = json.dumps(dict(messageType="register", channelID=chid))
+            self.ws.send(body)
+            res = json.loads(self.ws.recv())
+            assert res["messageType"] == (
+                "register",
+                f"Unexpected messageType. Expected: register Actual: {res['messageType']}",
             )
```
Trinaa, line 111: @jrbenny35 please apply this fix, Phil and I discovered it during calibration

```suggestion
            assert (
                 res["messageType"] == "register"
            ), f"Unexpected messageType. Expected: register Actual: {res['messageType']}"
```

jrbenny35: Looks like the linter messed this up huh

Trinaa: I think I messed it up actually. sry about that

Trinaa (approved):

PR #421, opened 2023-08-11
feat: set connected_at in User table whenever a bridge returns a successful handoff [CONSVC-1591, SYNC-3374]

jrconlin:
Closes: CONSVC-1591, SYNC-3374


autoendpoint/src/routes/registration.rs
```diff
@@ -157,11 +157,27 @@ pub async fn new_channel_route(
...
+/// and refresh the last_connect time. This will indicate how "lively" this
+/// device happens to be.
 pub async fn get_channels_route(
     _auth: AuthorizationCheck,
     path_args: RegistrationPathArgsWithUaid,
     app_state: Data<AppState>,
 ) -> ApiResult<HttpResponse> {
+    // this function should only called from mobile devices.
+    if let Some(user) = app_state.db.get_user(&path_args.uaid).await? {
+        app_state.db.update_user(&user).await?;
```
pjenvey, line 175: I don't think this will modify the user record at all as `update_user` simply writes what you give it.

Why are we updating `connected_at`? I thought we decided on the previous PR that we'd switch to updating the older `expiry` field instead of `last_connect` or `connected_at`.

We could even switch on DynamoDB's TTLs on the router table after running this on production for a while, but we'll probably just focus on the migration instead.

pjenvey (changes requested):

autoendpoint/src/routes/registration.rs
jrconlin: _Sigh_ I think that the more recent changes have invalidated this PR. 

I'll move it to draft, merge the various updates and resubmit.

(The old code had `update_user` updating `expiry`, which I think got dropped from [the newer DynamoDBUser](https://github.com/mozilla-services/autopush-rs/blob/58086e77a7f0ae6d84a6a7bbce246b90cfadbad4/autopush/src/db/models.rs#L68))

jrconlin:
Looking a bit more into this, I'm going to write up a proposal, because I think it's a lot more complicated now.

PR #422, opened 2023-08-11
test: calibrate load tests

Trinaa:
- setup recommended user count, spawn rate and resources required to trigger scaling in the autopush load scenario
- update README with instructions for distributed GCP execution and links for load test class diagram and load test result spreadsheet
- configure locust log files

[SYNC-3847](https://mozilla-hub.atlassian.net/browse/SYNC-3847)

[SYNC-3847]: https://mozilla-hub.atlassian.net/browse/SYNC-3847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/load/locustfile.py
```diff
@@ -89,7 +91,9 @@ def hello(self) -> None:
         with self._time_event(name="hello"):
             body = json.dumps(dict(messageType="hello", use_webpush=True))
             self.ws.send(body)
-            res = json.loads(self.ws.recv())
+            reply = self.ws.recv()
+            assert reply is not None, "No 'hello' response"
```
pjenvey, line 95: It doesn't look like we'll get `None` out of `recv()`, only an empty str.
```suggestion
            assert reply, "No 'hello' response"
```

tests/load/locustfile.py
```diff
@@ -104,7 +108,9 @@ def register(self) -> None:
         with self._time_event(name="register"):
             body = json.dumps(dict(messageType="register", channelID=chid))
             self.ws.send(body)
-            res = json.loads(self.ws.recv())
+            reply = self.ws.recv()
+            assert reply, "No 'register' response"
+            res = json.loads(reply)
```
jrbenny35, line 113: Wouldn't ```json.loads``` fail if the variable ```reply``` was empty?

jrbenny35 (changes requested):
Had one question

tests/load/locustfile.py
Trinaa: My understanding is if `reply` is empty the assertion will fail, raising an exception, and the code on line 113 should not be executed.

jrconlin: Yeah, sorry if it wasn't very explicit.

My hope is that since we can't examine the logs easily, we could use the assert errors to diagnose which of the messages are coming back as empty. These won't explain why those messages are coming back empty, but there are some assumptions we can make.

If the `hello` is coming back empty, then there was something wrong with the sent `hello` (conflicting UAID, bad value, timeout, etc.) and the server terminated the connection, which resulted in the empty packet being sent.

If the `register` came back empty, then there was a problem with the register command (e.g. the requested ChannelID was not lower case & hyphenated,  or the endpoint generation failed, etc.)

Trinaa: @jrconlin should I add some additional information to the assertion, perhaps print the channelID?

jrbenny35 (dismissed):

pjenvey (dismissed):

jrconlin (approved):

pjenvey (approved):

pjenvey (approved):

PR #425, opened 2023-08-14
bug: Detect logging hostname based off of specified platform 

jrconlin:
Improve logging to only try to get platform specific hostname when on that platform.

This introduces `AUTO{CONNECT,END,PUSH}__HOSTNAME` as an optional environment override for the hostname to use.

This introduces a new setting `server_platform` which is a string indicating the available platforms (currently 'aws', 'gcp', or 'unknown')

Closes: SYNC-3873


jrconlin:
No longer required. aging out.

PR #427, opened 2023-08-17
test: add a load test shape

Trinaa:
The shape class will distribute the swarming of Locust users to approximate a real-world traffic pattern

**Example:** 
100 users - 2 minutes - AutopushLoadTestShape
image

100 users - 2 minutes - Default Shape
image

[SYNC-3828](https://mozilla-hub.atlassian.net/browse/SYNC-3828)

[SYNC-3828]: https://mozilla-hub.atlassian.net/browse/SYNC-3828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/load/locustfiles/load.py
```diff
@@ -0,0 +1,81 @@
...
+    def _get_calculate_users_function(self) -> Callable[[int], int]:
+        a, b, c = numpy.polyfit(
+            [0, (self.MAX_RUN_TIME / 2), self.MAX_RUN_TIME], [0, self.MAX_USERS, 0], 2
+        )
+
+        def calculation(x: int) -> int:
+            return int(round((a * math.pow(x, 2)) + (b * x) + c))
+
+        return calculation
+
```
pjenvey, line 61: Looks like we can kill this?
```suggestion
```

tests/load/locustfiles/load.py
```diff
@@ -0,0 +1,81 @@
...
+        """Determined the number of active users given a run time.
+
+        Returns:
+            int: The number of users
+        """
+        return int(round((self.a * math.pow(run_time, 2)) + (self.b * run_time) + self.c))
+
+
+class AutopushLoadTestShape(LoadTestShape):
+    """A load test shape class for Autopush (Duration: 10 minutes, Users: 55500).
```
pjenvey, line 38:
```suggestion
    """A load test shape class for Autopush (Duration: 10 minutes, Users: 8330).
```

tests/load/locustfiles/load.py
```diff
@@ -0,0 +1,81 @@
...
+
+"""Load test shape module."""
+import math
+from typing import Callable, Type
+
+import numpy
+from locust import LoadTestShape, User
+from locustfile import AutopushUser
+
+TickTuple = tuple[int, float, list[Type[User]] | None]
```
pjenvey, line 13:
```suggestion
TickTuple = tuple[int, int, list[Type[User]]]
```

pjenvey: I don't think we need the Option here as tick returns this type or None?

pjenvey (changes requested):

tests/load/locustfiles/load.py
Trinaa: For reference this is the abstract method definition:
image

I think I'll remove the optional as you suggest, but change the `spawn_rate` to a float type to be more consistent with the base. What do you think?

tests/load/locustfiles/load.py
Trinaa: Oops 🤦🏻‍♀️! ty!

tests/load/locustfiles/load.py
```diff
@@ -0,0 +1,70 @@
...
+from locust import LoadTestShape, User
+from locustfile import AutopushUser
+
+TickTuple = tuple[int, float, list[Type[User]]]
+
+
+class QuadraticTrend:
+    """A class that defines a quadratic LoadTestShape trend."""
+
+    a: float
```
jrconlin, line 19: Could we use more descriptive variable names here?

tests/load/locustfiles/load.py
```diff
@@ -0,0 +1,70 @@
...
+        return int(round((self.a * math.pow(run_time, 2)) + (self.b * run_time) + self.c))
+
+
+class AutopushLoadTestShape(LoadTestShape):
+    """A load test shape class for Autopush (Duration: 10 minutes, Users: 83300).
+    Note: The Shape class assumes that the workers can support the generated spawn rates.
+    """
+
+    MAX_RUN_TIME: int = 600  # 10 minutes
+    MAX_USERS: int = 83300
```
jrconlin, line 43: This feels a bit like a magic number. Can you add a comment as to why you picked this or if it's just an arbitrary amount?

Trinaa: It is indeed a bit magic like, I share your sentiment. Comments are added, what do you think?

tests/load/locustfiles/load.py
Trinaa: I'm willing to change them, but actually am at a loss for what to name them, since that's how they appear in the quadratic function, do you have a suggestion? I thought to add a comment, is that better?

jrconlin: That's fair. 
I've [seen them referred](https://en.wikipedia.org/wiki/Quadratic_equation) to as the Quadradic, Linear and Constant coefficients, but those may not really help define them much to someone going through the code.
We can keep them _a_, _b_, and _c_, for now.

tests/load/locustfiles/load.py
jrconlin: If someone else six months from now decides to change this value to `79481` (because those were his lucky lottery numbers), what impact would that have?

Trinaa:
I executed the shape on GCP and noticed the class-picker env var was missing. Otherwise all seemed well. Results
image

tests/load/locustfiles/load.py
Trinaa: The max users can be seen as the peak of the parabola created by the trend. Is that helpful information to write down? 

A little bit I would have to trust that whoever maintains these numbers reads a bit about Locust. No matter what you are asked to input a 'Magic' user number.

jrconlin: Ah, Yes! that would be helpful. Maybe we rename that as `peak_user_load` and add a note that the app will ramp up and down to that number, and that for now the number is arbitrarily chosen. 

Sorry to be a pain about all of this, but I've been caught by things like that in the past and I've learned to treat myself as the idiot that I will become.

pjenvey (approved):

PR #430, opened 2023-08-28
Updates for Aug 2023

jrconlin:
Includes updates for Rust 1.72

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ defaults:
 jobs:
   audit:
     docker:
-      - image: rust:1.70
+      - image: rust:1.72
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ defaults:
 jobs:
   audit:
     docker:
-      - image: rust:1.70
+      - image: rust:1.72
```
pjenvey, line 27: Can you please also add in this PR a force of this same version to the rustup command. It's a bug that we're having it fmt/clippy with the latest stable it pulls down.

You want to pass `--default-toolchain 1.72` after `sh -s -- `, and let's add a comment here to note that these should both match. Unfortunately there doesn't seem to be a way to declare the version number once for reuse in another command via e.g. yaml anchors.

pjenvey (changes requested):

pjenvey (approved):

PR #433, opened 2023-08-31
feat: Add send notification task to locust load test file

jrbenny35:
This adds a send notification task to locust. It also updates the register task to handle errors a bit better.

[SYNC-3846](https://mozilla-hub.atlassian.net/jira/software/c/projects/SYNC/issues/SYNC-3846)

Note: I didn't add the unregister function as it seems like we don't need that currently, but it can be added if needed.

[SYNC-3846]: https://mozilla-hub.atlassian.net/browse/SYNC-3846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pjenvey:
Rebasing against master will fix the ci failures

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +172,56 @@ def register(self) -> None:
...
             ValidationError: If the register message schema is not as expected
         """
+        if not self.uaid:
+            raise RescheduleTask()
+
         chid: str = str(uuid.uuid4())
+        control: bool = True
+
+        with self._time_event(name="send_notification") as timer:
+            while control:
```
Trinaa, line 182: This while loop can be removed.

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +172,56 @@ def register(self) -> None:
...
+                try:
+                    body = json.dumps(dict(messageType="register", channelID=chid))
+                    self.ws.send(body)
+                    reply = self.ws.recv()
+                except ssl.SSLError:
+                    # if there is an error, disconnect and retry
+                    self.disconnect()
+                    self.connect()
+                    self.hello()
```
Trinaa, line 191: This is also doing flow-of-control through a try-catch, which should be avoided.

In the case there is an error instead of a retry, we should report the error in the performance metrics.
It could be we need to stop the user after.

tests/load/locustfiles/locustfile.py
```diff
@@ -123,6 +139,31 @@ def hello(self) -> None:
...
+        before sending an acknowledgement.
+        """
+        with self._time_event(name="acknowledge"):
+            channel_id = list(self.channels.keys())[-1]
+            body = json.dumps(
+                dict(messageType="ack", use_webpush=True, updates=dict(channelID=channel_id))
+            )
+
+            self.connect()
+            self.hello()
```
Trinaa, line 157: In the case where we are re-connecting. I think we want to re-use the `uaid`? I think we need to add support in the hello() method for this.

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +172,56 @@ def register(self) -> None:
...
-            res: RegisterMessage = RegisterMessage(**json.loads(reply))
-            assert res.status == 200, f"Unexpected status. Expected: 200 Actual: {res.status}"
-            assert res.channelID == chid, f"Channel ID did not match, received {res.channelID}"
-            timer.response_length = len(reply.encode("utf-8"))
-        self.channels[chid] = res.pushEndpoint
+        Raises:
+            AssertionError: If the server does not respond correctly (400, 500, etc)
+        """
+        if not self.channels:
+            raise RescheduleTask()
```
Trinaa, line 213: returning is preferable to raising ResheduleTask()

At the moment how do we end up in a state where `channels` is Falsy? We register a channel on start and stop the user if it doesn't succeed. We also don't have an unregister method yet. Do we need this check at this point in time?

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +172,56 @@ def register(self) -> None:
...
 
-    @task
-    def do_nothing(self) -> None:
-        pass
+        self.disconnect()
+
+        with self._time_event(name="send_notification") as timer:
```
Trinaa, line 217: We shouldn't use the `_time_event` context when using the `client`. Instead use a with statement with the `client` and I think 'catch_response' must be set to True.

```
with self.client.post(..., catch_response=True, ... ) as response:
```

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +172,56 @@ def register(self) -> None:
...
+            endpoint_url = random.choice(list(self.channels.items()))[-1]
+            endpoint_res = self.client.post(
+                url=endpoint_url,
+                name="Endpoint Notification",
+                data=self.encrypted_data,
+                headers=self.headers,
+            )
+            assert endpoint_res.status_code == 201, f"status code was {endpoint_res.status_code}"
```
Trinaa, line 225: Because we are using the `self.client` an assertion isn't the appropriate choice for verification. Instead we change the response status. Something like:

```suggestion
            if response.status_code != 201:
                response.failure(f"{response.status_code=}, expected 201, {response.text=}")
                return
```

Trinaa (changes requested):
Sorry @jrbenny35 I'm asking for a lot of changes. If it's easier we can go through them together in pair.

Also, have you had a chance to run the test in GCP?
Before merging this change we'll need to evaluate the impact and adjust the Users and Spawn Rate if needed (ref: [Previous Results](https://docs.google.com/spreadsheets/d/1-nF8zR98IiLZlwP0ISwOB8cWRs74tJGM1PpgDSWdayE/edit#gid=0)) Let me know if you'd like to do that together.

tests/load/locustfiles/locustfile.py
jrbenny35: Check if user already has UAID and reuse it

tests/load/locustfiles/locustfile.py
```diff
@@ -123,6 +139,31 @@ def hello(self) -> None:
...
+            body = json.dumps(
+                dict(messageType="ack", use_webpush=True, updates=dict(channelID=channel_id))
+            )
+
+            self.connect()
+            self.hello()
+
+            try:
+                self.ws.send(body)
+                self.ws.recv()
```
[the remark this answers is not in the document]
jrbenny35: Check Status code and response length

tests/load/locustfiles/locustfile.py
```diff
@@ -123,6 +139,31 @@ def hello(self) -> None:
...
+            except ssl.SSLError:
+                self.ws.recv()
+                self.ws.send(body)
```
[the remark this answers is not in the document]
jrbenny35: Remove try/catch and try and receive the notification

tests/load/locustfiles/locustfile.py
jrbenny35: Should bail is failure

tests/load/locustfiles/locustfile.py
jrbenny35: Client already supports time tracking

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +172,56 @@ def register(self) -> None:
             AssertionError: If the user fails to register a channel
             ValidationError: If the register message schema is not as expected
         """
+        if not self.uaid:
+            raise RescheduleTask()
+
         chid: str = str(uuid.uuid4())
+        control: bool = True
+
+        with self._time_event(name="send_notification") as timer:
```
Trinaa, line 181:
```suggestion
        with self._time_event(name="register") as timer:
```

Trinaa:
@jrbenny35 I noticed that 'hello' can have 'channelIDs' as part of it's payload. Perhaps the absence of this information on re-connection is causing some of the odd behaviour.

tests/load/locustfiles/locustfile.py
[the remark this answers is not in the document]
Trinaa: Further investigation reveals that the response can be modelled as follows:

```
class AckMessage(BaseModel):
    """Autopush 'ack' response message."""
    data: str
    headers: dict[str, str]
    messageType: Literal["notification"]
    channelID: str
    version: str
```

There is no status in this case, so I suggest that we verify using a combination of 'channelID' and 'data'. They should match the values that were posted.

tests/load/locustfiles/locustfile.py
```diff
@@ -123,6 +139,31 @@ def hello(self) -> None:
...
+        Send an 'ack' message to push.
+
+        After sending a notification, the client must also send an 'ack' to the server to
+        confirm receipt. If there is a pending notification, this will try and receive it
+        before sending an acknowledgement.
+        """
+        with self._time_event(name="acknowledge"):
+            channel_id = list(self.channels.keys())[-1]
+            body = json.dumps(
+                dict(messageType="ack", use_webpush=True, updates=dict(channelID=channel_id))
```
pjenvey, line 153: Only hello needs `use_webpush`
```suggestion
                dict(messageType="ack", updates=dict(channelID=channel_id))
```

pjenvey (changes requested):
Also I noticed we're passing `ssl=False` to `create_connection`. The websocket connection definitely forces SSL so I believe this is an invalid argument that the library's ignoring (closest I see is its `sslopt` argument)

tests/load/locustfiles/locustfile.py
```diff
@@ -123,6 +139,31 @@ def hello(self) -> None:
...
+        """
+        print(f"notification")
+        with self._time_event(name="acknowledge") as timer:
+            reply = self.ws.recv()
+            notification = NotificationMessage(**json.loads(reply))
+            body = json.dumps(
+                dict(
+                    messageType="ack",
+                    use_webpush=True,
+                    updates=dict(channelID=notification.channelID, version=notification.version)
```
pjenvey, line 158: The register failing is due to the `ack` being malformed, causing autopush to not be in a happy state when the register was issued:
```suggestion
                    messageType="ack",
                    updates=[dict(channelID=notification.channelID, version=notification.version)]
```

tests/load/locustfiles/locustfile.py
```diff
@@ -123,6 +146,29 @@ def hello(self) -> None:
...
+            notification = NotificationMessage(**json.loads(reply))
+            body = json.dumps(
+                dict(
+                    messageType="ack",
+                    updates=[dict(channelID=notification.channelID, version=notification.version)],
+                )
+            )
+            self.ws.send(body)
+            timer.response_length = len(reply.encode("utf-8"))
+        time.sleep(1)
```
pjenvey, line 169: Why do we sleep here, is it necessary?

pjenvey:
Here's a [branch](https://github.com/mozilla-services/autopush-rs/compare/sync-3846-2...sync-3846-2-direct-param) that makes the disconnect/reconnect optional via a parameter. The disconnect causes the notification to be written to DynamoDB: I think it's useful to have both options.

A mixture of "both" would be nice to have -- ideally via task weighting -- but I'm not quite sure how a parameter could easily drive that.

tests/load/locustfiles/locustfile.py
```diff
@@ -66,13 +108,25 @@ def __exit__(self, *args) -> None:
...
+                [
+                    random.choice(string.ascii_letters + string.digits)
+                    for i in range(0, random.randrange(1024, 4096, 2) - 8)
+                ]
+            )
+            + "=="
+        )
+
+    def wait_time(self):
+        return self.environment.autopush_wait_time(self)
```
Trinaa, line 129: Something is not evaluating correctly here, so that i'm not able to execute the load tests. Error message:
image

I've kept the default input:
image

tests/load/locustfiles/models.py
```diff
@@ -24,5 +24,15 @@ class RegisterMessage(BaseModel):
 
     messageType: Literal["register"]
     channelID: str
-    status: int
+    status: int = 200
```
Trinaa, line 27: Instead of an assignment, I think we probably want a literal. Should the register message return without a status an assignment would add a 200, which would give us a false positive.

```suggestion
    status: Literal[200]
```

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +219,47 @@ def register(self) -> None:
...
+
+        Raises:
+            AssertionError: If the server does not respond correctly (400, 500, etc)
+        """
+        stored = self.environment.parsed_options.notification_type == NotificationType.STORED
+        if stored:
+            self.disconnect()
+
+        channel_id = random.choice(list(self.channels.keys()))
+        endpoint_url = self.channels[channel_id]
```
Trinaa, line 248: suggestion

```suggestion
        channel_id, endpoint_url = random.choice(list(self.channels.items()))

```

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +219,47 @@ def register(self) -> None:
...
             timer.response_length = len(reply.encode("utf-8"))
         self.channels[chid] = res.pushEndpoint
 
-    @task
-    def do_nothing(self) -> None:
-        pass
+    @task(weight=95)
+    def send_notification(self):
+        """
+        Sends a notification to a registered enpoint
```
Trinaa, line 238: nit
```suggestion
        Sends a notification to a registered endpoint
```

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +219,47 @@ def register(self) -> None:
...
+            AssertionError: If the server does not respond correctly (400, 500, etc)
+        """
+        stored = self.environment.parsed_options.notification_type == NotificationType.STORED
+        if stored:
+            self.disconnect()
+
+        channel_id = random.choice(list(self.channels.keys()))
+        endpoint_url = self.channels[channel_id]
+
+        self.disconnect()
```
Trinaa, line 250: I think this is a left over from a merge. 
```suggestion
```

tests/load/locustfiles/locustfile.py
```diff
@@ -131,19 +219,47 @@ def register(self) -> None:
...
+            headers=self.headers,
+        ) as response:
+            if response.status_code != 201:
+                response.failure(f"{response.status_code=}, expected 201, {response.text=}")
+                return
+
+        if stored:
+            self.connect()
+            self.hello()
+        self.ack()
```
Trinaa, line 265: In the case where we send 'stored' messages, a failure to post would result in the user becoming disconnected because we return on line 260.

I solved this locally by extracting the post logic into another method, and checking for success before calling self.ack().

```suggestion
        notification_success: bool = self.post_notification(endpoint_url)

        if stored:
            self.connect()
            self.hello()

        if not notification_success:
            return

        self.ack()
```

```
    def post_notification(self, endpoint_url) -> bool:
        """
        Send a notification to Push

        Args:
            endpoint_url: A channel destination endpoint url
        Returns:
            bool: Flag indicating if the post notification request was successful
        """
        with self.client.post(
                url=endpoint_url,
                name="notification",
                data=self.encrypted_data,
                headers=self.headers,
        ) as response:
            if response.status_code != 201:
                response.failure(f"{response.status_code=}, expected 201, {response.text=}")
                return False
        return True
```

Trinaa (changes requested):

tests/load/locustfiles/models.py
jrbenny35: Ah, I didn't mean to include that. Will remove it

tests/load/locustfiles/locustfile.py
jrbenny35: Could we just ```else```  and not return? 
```suggestion
        with self.client.post(
            url=endpoint_url,
            name="Endpoint Notification",
            data=self.encrypted_data,
            headers=self.headers,
        ) as response:
            if response.status_code == 201:
                 if stored:
                     self.connect()
                     self.hello()
                 self.ack()
             else:   
                 response.failure(f"{response.status_code=}, expected 201, {response.text=}")
```

Trinaa: I think for the else to work it would have to look like this right 😬?
```
if response.status_code == 201:
    if stored:
        self.connect()
        self.hello()
    self.ack()
else:
    response.failure(f"{response.status_code=}, expected 201, {response.text=}")
    if stored:
        self.connect()
        self.hello()
```

jrbenny35: Oh so we need to run ```connect``` and ```hello``` even on a failure? If the endpoint doesn't return a 201 do we need to reconnect or should we just hope the user (through locust) tries to send another notification and when that passes we reconnect and send the ```ack```

Trinaa: Leaving the users connected is part of the load we are trying to re-create is my understanding and in future when we have more tasks, what if the next task chosen isn't 'send_notification', we would be leaving the user in an unexpected state.

tests/load/locustfiles/locustfile.py
Trinaa: This is resolved in https://github.com/mozilla-services/autopush-rs/pull/438

Trinaa:
Hi @jrbenny35 

Changes from our pair session today --> https://github.com/mozilla-services/autopush-rs/pull/438

jrbenny35:
Merged in @Trinaa's branch. https://github.com/mozilla-services/autopush-rs/pull/438/files

tests/load/locustfiles/locustfile.py
```diff
@@ -96,54 +141,149 @@ def connect(self) -> None:
...
-            assert res.status == 200, f"Unexpected status. Expected: 200 Actual: {res.status}"
-            timer.response_length = len(reply.encode("utf-8"))
-        self.uaid = res.uaid
 
-    def register(self) -> None:
+        with self._time_event(name="hello - recv") as timer:
+            response: str = self.ws.recv()
+            assert response, "No 'hello' response"
+            message: HelloMessage = HelloMessage(**json.loads(response))
+            assert message.status == 200, f"Status Error. Expected: 200 Actual: {message.status}"
```
Trinaa, line 177: This can be removed, since the HelloMessage model now enforces it.
```suggestion
```

Trinaa (commented):
There has been a lot of churn on this PR and I thank you @jrbenny35 for coordinating all the debugging that was going on. I'm going to suggest we clean-up a bit and stop here for the scope of [SYNC-3846](https://mozilla-hub.atlassian.net/browse/SYNC-3846). We can continue with the following tasks:

* [SYNC-3914](https://mozilla-hub.atlassian.net/browse/SYNC-3914) - This bug was discovered during the implementation of this task and the solution looks extensive enough to warrant an individual ticket and PR.
* [SYNC-3915](https://mozilla-hub.atlassian.net/browse/SYNC-3915) - This work was originally scoped for this PR, but is blocked by [SYNC-3914](https://mozilla-hub.atlassian.net/browse/SYNC-3914), so I think it's best to continue after the bug is resolved.
* [SYNC-3916](https://mozilla-hub.atlassian.net/browse/SYNC-3916) - As @pjenvey suggested, I think we should add this functionality in a follow-up epic.

What do you think?

PS: Should we squash these commits before merging?

jrbenny35:
Bit of a scary squash but I think I got it

tests/load/locustfiles/locustfile.py
```diff
@@ -96,54 +134,132 @@ def connect(self) -> None:
...
+        before sending an acknowledgement.
+
+        Raises:
+            AssertionError: If the notification message response is empty
+            ValidationError: If the notification message schema is not as expected
+        """
+        ack_sleep: int = self.environment.parsed_options.ack_sleep
+
+        if ack_sleep:
+            time.sleep(ack_sleep)
```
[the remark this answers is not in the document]
jrbenny35: Ah. that's what I get for not running it before a push

Trinaa (dismissed):
🚀 Thanks for the hard work @jrbenny35!

tests/load/locustfiles/locustfile.py
```diff
@@ -96,54 +134,127 @@ def connect(self) -> None:
...
-            timer.response_length = len(reply.encode("utf-8"))
-        self.channels[chid] = res.pushEndpoint
-
-    @task
-    def do_nothing(self) -> None:
-        pass
+
+        with self._time_event(name="register - recv") as timer:
+            response: str = self.ws.recv()
+            assert response, "No 'notification' response"
```
pjenvey, line 243:
```suggestion
            assert response, "No 'register' response"
```

tests/load/locustfiles/locustfile.py
```diff
@@ -96,54 +134,127 @@ def connect(self) -> None:
         self.ws = create_connection(
             self.environment.parsed_options.websocket_url,
             header={"Origin": "http://localhost:1337"},
-            ssl=False,
+            timeout=30,  # timeout defaults to None
```
pjenvey, line 137: I wonder if this is too high -- it might be preferable to fail faster when things aren't working as expected.

pjenvey (changes requested):

tests/load/locustfiles/locustfile.py
Trinaa: oh! that's my bad, at one point I thought the timeout had to be greater than our wait_time interval. What's a good amount 5?

pjenvey: It's very short in our integration tests (1) but 5 works

pjenvey (approved):

Trinaa (approved):

PR #435, opened 2023-09-05
feat: emit a tag in autoconnect's metrics

pjenvey:
to distinguish its metrics from the legacy version

and consolidate autoendpoint's metrics builder w/ autopush_common's

[SYNC-3893](https://mozilla-hub.atlassian.net/browse/SYNC-3893)

[SYNC-3893]: https://mozilla-hub.atlassian.net/browse/SYNC-3893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrconlin (approved):
I'm a bit torn about having these both use the legacy mark. I can see having them on the same chart might be a bit easier to compare, but I think it might be hard to distinguish them cleanly.

PR #439, opened 2023-09-21
fix: AutopushUser fails with WebSocketConnectionClosedException

Trinaa:
This PR manages closing WebSocket connections by using [WebSocketApp](https://websocket-client.readthedocs.io/en/latest/app.html#websocket._app.WebSocketApp), which seems to be the recommendation when needing [long-lived](https://github.com/websocket-client/websocket-client#long-lived-connection) WebSocket connections.

[SYNC-3914](https://mozilla-hub.atlassian.net/browse/SYNC-3914)

[SYNC-3914]: https://mozilla-hub.atlassian.net/browse/SYNC-3914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/load/locustfiles/locustfile.py
+++ b/tests/load/locustfiles/locustfile.py
@@ -6,18 +6,40 @@
 
 import base64
 import json
+import logging
 import random
 import string
 import time
 import uuid
-from typing import Any
+from json import JSONDecodeError
+from logging import Logger
+from typing import Any, TypeAlias
 
+import gevent
+import websocket
 from args import parse_wait_time
 from exceptions import ZeroStatusRequestError
+from gevent import Greenlet
 from locust import FastHttpUser, events, task
-from models import HelloMessage, NotificationMessage, RegisterMessage
+from models import (
+    HelloMessage,
+    HelloRecord,
+    NotificationMessage,
+    NotificationRecord,
+    RegisterMessage,
+    RegisterRecord,
+)
 from pydantic import ValidationError
-from websocket import create_connection
+from websocket import WebSocketApp, WebSocketConnectionClosedException
+
+Message: TypeAlias = HelloMessage | NotificationMessage | RegisterMessage
+Record: TypeAlias = HelloRecord | NotificationRecord | RegisterRecord
+
+# Set to 'True' to view the verbose connection information for the web socket
+websocket.enableTrace(False)
+websocket.setdefaulttimeout(5)
+
+logger: Logger = logging.getLogger("AutopushUser")
 
 
 @events.init_command_line_parser.add_listener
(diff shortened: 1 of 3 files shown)


tests/load/locustfiles/locustfile.py
```diff
@@ -31,119 +48,392 @@ def _(parser: Any):
...
-            context=self.user.context(),
-        )
+@events.test_start.add_listener
+def _(environment, **kwargs):
+    environment.autopush_wait_time = parse_wait_time(environment.parsed_options.wait_time)
 
 
 class AutopushUser(FastHttpUser):
-    wait_time = between(30, 35)
+    REST_HEADERS: dict[str, str] = {"TTL": "60", "Content-Encoding": "aes128gcm", "Topic": "aaaa"}
```
jrconlin, line 66: A "Topic" header means that subsequent subscription updates will replace the undelivered update. (e.g. if you have a pending 'aaaa' topic update, the new 'aaaa' update will replace it. This means that the UA will receive only the latest 'aaaa' update, instead of the default behavior, which is to receive ALL pending updates.)

Is this intended?

(Even if it is, might be worth adding a comment describing this behavior for future folk that might not be aware of it.)

tests/load/locustfiles/locustfile.py
```diff
@@ -31,119 +48,392 @@ def _(parser: Any):
...
-    def hello(self) -> None:
+        Args:
+            endpoint_url: A channel destination endpoint url
+        Raises:
+            ZeroStatusRequestError: In the event that Locust experiences a network issue while
+                                    sending a notification.
         """
-        Send a 'hello' message to Autopush.
+        message_type: str = "notification"
+        data: str = "TestData" + "".join(
```
jrconlin, line 193: I think this was covered in the discussion, but leading off the data with `TestData` followed by random garbage allows you to easily differentiate the payload from the incoming base64 encoded data segment (which can  also look like random garbage).

Again, not sure if you might want to add some comment here noting that since it was a big point of confusion yesterday.

tests/load/locustfiles/locustfile.py
```diff
@@ -31,119 +48,392 @@ def _(parser: Any):
...
+            decoded_data: str = base64.urlsafe_b64decode(message.data + "===").decode("utf8")
+
+            record: MessageRecord = next(
+                (
+                    r
+                    for r in self.message_records
+                    if r.message_type == message.messageType and r.data == decoded_data
+                ),
+                None,
+            )
```
jrconlin, line 283: Does it make sense to "pop" a recv'd message out of `message_records`? Otherwise, if the server just returned the very first message it got, wouldn't this still return back that things were successful?

(Alternately, since this is a load test and not an integration or unit test, if you wanted to save a bit of memory, maybe have the data always be the same and validate that the received notification data matches expected?)

tests/load/locustfiles/locustfile.py
```diff
@@ -31,119 +48,392 @@ def _(parser: Any):
...
+            message = HelloMessage(**json.loads(data))
+
+            # If a user is not subscribed to any channels upon re-connection a new
+            # ID will be assigned
+            self.uaid = message.uaid
+
+            # Get last 'hello' message record
+            record: MessageRecord = next(
+                (
+                    self.message_records[i]
```
jrconlin, line 236: I'm a bit curious about this `message_records` bucket. It contains all messages and data regardless of the content? (e.g. `hello` responses mixed in with `notification` messages, etc.) We then do a scan of this bucket looking for a match, right? 

I'm wondering if we should divvy these up a bit so that notifications are distinct from commands that way we're not potentially missing if we cross streams at some point.

tests/load/locustfiles/locustfile.py
Trinaa: I think this might explain why my numbers don't add up during test runs! TY!

I actually don't know the original intent behind having the topic here, but my gut reaction from your description is that for load test purposes we should remove this.

tests/load/locustfiles/locustfile.py
Trinaa: Yes! I will add a comment.

tests/load/locustfiles/locustfile.py
Trinaa: Originally I was poping messages from message_records. However WebSocket connection issues before or during 'ack' would happen (async yay!) and then when the message was recent, we had no matching message in message_records. I've opted to store all the messages we sent and am identifying them with the 'decoded_data", which i'm counting on being unique. 

Verifying that message data is correct is nice, but what I'm really after is measuring the response time.

You are right about there being a memory impact though.

tests/load/locustfiles/locustfile.py
Trinaa: We could do this, my original implementation had 3 queues.
The hello use case is wasteful for sure because I don't have a unique identifier and have to assume the last instance is correct.

message_records should be a a small data structure 30-40 entries max.

jrbenny35 (changes requested):
Looks great Kat, thanks. I was wondering if we could combine some of the ```send_xyz``` methods into one. They look quite similar. Also, the ```recv_xyz``` methods also look like they could be combined. Any reason you didn't want to do that?

jrbenny35 (dismissed):
Looks like that clean up the code quite a bit too

tests/load/locustfiles/locustfile.py
```diff
@@ -32,6 +32,9 @@
 from pydantic import ValidationError
 from websocket import WebSocketApp, WebSocketConnectionClosedException
 
+Message: TypeAlias = HelloMessage | NotificationMessage | RegisterMessage | None
+Record: TypeAlias = HelloRecord | NotificationRecord | RegisterRecord | None
```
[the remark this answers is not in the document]
Trinaa: The [doc](https://docs.python.org/3/library/typing.html#type-aliases) can explain better than me. 

Here I'm declaring a type alias so that I don't have to write `HelloMessage | NotificationMessage | RegisterMessage | None` each time, instead I can just say `Message`.

tests/load/locustfiles/locustfile.py
```diff
@@ -222,129 +220,60 @@ def post_notification(self, endpoint_url: str) -> None:
...
+                        (r for r in self.notification_records if r.data == decode_data), None
+                    )
+                case "register":
+                    message = RegisterMessage(**message_dict)
+                    record = next(
+                        (r for r in self.register_records if r.channel_id == message.channelID),
+                        None,
+                    )
+                case _:
+                    exception = f"Unexpected data was received. Data: {data}"
```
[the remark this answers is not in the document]
Trinaa: Thanks for pointing it out and making me rethink things

tests/load/locustfiles/locustfile.py
```diff
@@ -50,211 +72,282 @@ def _(environment, **kwargs):
...
+        """Called when received data from a WebSocket.
 
-        Connections must say hello after connecting to the server, otherwise the connection is
-        quickly dropped.
+        Args:
+            ws: WebSocket class object
+            data: utf-8 data received from the server
+        """
+        message: Message = self.recv(data)
+        if isinstance(message, HelloMessage):
```
pjenvey, line 122: FYI (we don't need to change this) Python's match statement can replace multiple `isinstance` checks like this. It looks like
```suggestion
        match message:
            case HelloMessage():
```

tests/load/locustfiles/locustfile.py
```diff
@@ -50,211 +72,282 @@ def _(environment, **kwargs):
...
-                use_webpush=True,
-                uaid=self.uaid,
-                channelIDs=list(self.channels.keys()),
-            )
+        logger.error(str(error))
+
+        # WebSocket closures are expected
+        if isinstance(error, WebSocketConnectionClosedException):
+            self.ws = None  # WebSocket is not available for task use
+            self.connect()  # Reconnect
```
pjenvey, line 147: I'm sure you've tested this reconnecting on WS conns closing case. However it's unclear to me how recreating the `WebSocketApp`/`run_forever` loop works with the previous instance of it spawned.

Apparently `WebSocketApp` supports reconnecting on its own but this lead me down a rabbit hole of its notion of a "dispatcher" which I'd rather not deal with. I guess this current method of reconnecting works ok? :shipit:

Trinaa: Summary from Slack discussion:
It seems that we need to set the `reconnect` Arg to a value higher than 0, in order for the `run_forever()` to apply.
It doesn't seem like we need a dispatcher for our use case.

tests/load/locustfiles/locustfile.py
Trinaa: mypy is raising the following errors when I make the switch:
image
I'm inclined to find the solution another day, unless you feel strongly about the match-case?

pjenvey (approved):

Trinaa:
@pjenvey I'll need a re-approval when you have a chance.
When verifying the reconnect change in GCP, I found 5 seconds to be too long, leading to WebSocket connection errors cluttering the UI. I reduced the value to 1 second and am now suppressing the UI feedback for connection errors in favour of logging.

pjenvey (approved):

Issue #367, opened 2023-04-21
Cleanup autoendpoint bridge RouterError metrics


data-sync-user:
In master we’ve recently began reporting statsd style metrics for bridge errors in 2 ways:

* A couple of `handle_error` callbacks done by the various `Router::route_notification` calls. This emits the `autoendpoint.notification.bridge.error` metric for all `RouterError`s, with `reason`, etc. tags depending on the error variant
* More recently via the more general mechanism shared across our actix-web based services: from a middleware. This generally emits different individual metrics per error variant

We’ve begun using the latter because we’ve: a) maybe forgotten about `handle_error` b) `handle_error` isn’t called on all `RouterError`'s we produce: it’s called in more specific places where we have more context about the error that’s added as metrics tags (mostly just `app_id` but also `platform` somewhat).

We might be able to remove `handle_error` completely by adding some more machinery to the general mechanism such as:

* Adding a syncstorage-rs’s `Taggable` interface, so we could `req.add_tag("app_id", app_id)` in those specific places
* Letting `Error`'s write their own arbitrary tags to their metrics. All `RouterErrors` would return the same `autoendpoint.notification.bridge.error` metric label but could provide a new `add_tags` method to add their specific `reason`, `error`, and `errno` tags from the middleware

The goals here:

* Ideally keep the tags emitted by `handle_error` the same
* We probably want to hide all `RouterError`'s from sentry (emitting metrics instead) but we still want the option of easily changing this to emit certain ones to sentry for diagnosing issues
* Ideally keeping the sentry vs metric configuration all in the same place (like the general mechanism attempts to do) as to avoid bad configurations (mistakenly hiding errors from both sentry and metrics)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3695)

PR #445, opened 2023-09-26
bug: Add better error messaging for GCM/FCM processing

jrconlin:
Attempt to determine the cause of the JSON parse fail for GCM messages

Issue: SYNC-3931

### DIFF
--- a/autoendpoint/src/routers/fcm/client.rs
+++ b/autoendpoint/src/routers/fcm/client.rs
@@ -238,10 +238,17 @@ impl FcmClient {
         // Handle error
         let status = response.status();
         if status.is_client_error() || status.is_server_error() {
-            let data: FcmResponse = response
-                .json()
+            let raw_data = response
+                .bytes()
                 .await
                 .map_err(FcmError::DeserializeResponse)?;
+            if raw_data.is_empty() {
+                return Err(FcmError::EmptyResponse.into());
+            }
+            let data: FcmResponse = serde_json::from_slice(&raw_data).map_err(|e| {
+                let s = String::from_utf8(raw_data.to_vec()).unwrap_or_else(|e| e.to_string());
+                FcmError::InvalidResponse(e, s)
+            })?;
 
             // we only ever send one.
             return Err(match (status, data.error) {
--- a/autoendpoint/src/routers/fcm/client.rs
+++ b/autoendpoint/src/routers/fcm/client.rs
@@ -159,7 +159,7 @@ impl FcmClient {
             .await
             .map_err(FcmError::DeserializeResponse)?;
         if raw_data.is_empty() {
-            return Err(FcmError::EmptyResponse.into());
+            return Err(FcmError::EmptyResponse(status).into());
         }
         let data: GcmResponse = serde_json::from_slice(&raw_data).map_err(|e| {
             let s = String::from_utf8(raw_data.to_vec()).unwrap_or_else(|e| e.to_string());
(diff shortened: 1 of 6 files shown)


autoendpoint/src/routers/fcm/client.rs
```diff
@@ -238,10 +238,17 @@ impl FcmClient {
...
-                .json()
+            let raw_data = response
+                .bytes()
                 .await
                 .map_err(FcmError::DeserializeResponse)?;
+            if raw_data.is_empty() {
+                return Err(FcmError::EmptyResponse.into());
+            }
+            let data: FcmResponse = serde_json::from_slice(&raw_data).map_err(|e| {
+                let s = String::from_utf8(raw_data.to_vec()).unwrap_or_else(|e| e.to_string());
```
pjenvey, line 249: We probably shouldn't even bother including this string in the error until we're actually using it.

autoendpoint/src/routers/fcm/client.rs
```diff
@@ -238,10 +238,17 @@ impl FcmClient {
...
         let status = response.status();
         if status.is_client_error() || status.is_server_error() {
-            let data: FcmResponse = response
-                .json()
+            let raw_data = response
+                .bytes()
                 .await
                 .map_err(FcmError::DeserializeResponse)?;
+            if raw_data.is_empty() {
+                return Err(FcmError::EmptyResponse.into());
```
pjenvey, line 246: The response `status` code would be nice to know. I think you could put it in both of these new error variants and include it in their to_string (for now, since we can't easily put it in `extras`) because I doubt we'll be seeing more than 1 status code anyway.

pjenvey (changes requested):

autoendpoint/src/routers/fcm/client.rs
```diff
@@ -238,10 +238,17 @@ impl FcmClient {
         // Handle error
         let status = response.status();
         if status.is_client_error() || status.is_server_error() {
-            let data: FcmResponse = response
-                .json()
+            let raw_data = response
```
pjenvey, line 241: Per the metrics, we want this change in `send_gcm`, not `send` (but it might as well happen in both places)

autoendpoint/src/routers/fcm/client.rs
jrconlin: Well, I was thinking that capturing it here would at least give us some clue. 
I mean, I'm fine logging it instead, but we don't really capture logs either.
I also suspect that this string probably isn't different between calls, so it might be less cardinal than we think?

(e.g. "Go away. This doesn't exist anymore" or whatever)

autoendpoint/src/routers/fcm/client.rs
jrconlin: ah, doh! thanks forgot about the `send_gcm` function.

autoendpoint/src/routers/fcm/client.rs
jrconlin: Ok, also added `warn!()` messages. I figure Ops could enable them on a server and collect the logs. 

(hopefully by setting `RUST_LOG=autoendpoint=warn,error` for one of the boxes (if that's possible))

pjenvey: Right, my point was these weren't going to output anywhere AFAICT, since they're not in the Error's `to_string`.

I created [a branch](https://github.com/mozilla-services/autopush-rs/compare/bug/sync-3931_fcm...bug/sync-3931_fcm-trait-extras?expand=1) with an alternative to the new log messages. Turns out it's not very difficult to get extras/tags out from the `Error` itself now that we have a trait for this purpose -- and this is part of the work we ultimately want for https://github.com/mozilla-services/autopush-rs/issues/367

jrconlin:
Merged in that branch (swears at github)

pjenvey (approved):

PR #476, opened 2023-10-18
bug: Add a `RETRY-AFTER` header to 503 messages

jrconlin:
Closes #SYNC-3963


autoconnect/autoconnect-web/src/error.rs
```diff
@@ -25,7 +29,11 @@ impl ResponseError for ApiError {
 
     fn error_response(&self) -> HttpResponse {
         let code = self.status_code();
-        HttpResponse::build(code).json(json!({
+        let mut resp = HttpResponseBuilder::new(code);
```
pjenvey, line 32: We definitely want this in autoendpoint. I'm not sure there's a really compelling reason for it in autoconnect but I suppose it wouldn't hurt to have it here as well for its main websocket route "/".

pjenvey (changes requested):

pjenvey:
I think #478 handled all we need here?

autoconnect doesn't emit any 410s AFAICT so the cache header isn't really needed.

I don't think `Retry-After` makes much sense for autoconnect either, the only public route is the websocket connection. If there's a 503 while establishing the connection we don't generally want clients backing off since the whole point is to have the connection open all the time.

Hopefully the service is resilient enough to handle lots of establishing of connections (granted there could be a theoretical some constant thundering herd of immediate connects/disconnects situation issue but seems unlikely especially across a large enough number of our clients) -- which is pretty much what happens during redeploys. Plus it's too late to send the header after the WS upgrade anyway.

Also I rediscovered old backoff logic in the client triggered via a special websocket close code:

https://searchfox.org/mozilla-central/rev/4ab72574054da8d350689c0488e8ef74106e8b3f/dom/push/PushServiceWebSocket.sys.mjs#22

Further down memory lane we used to use this for a client bug (but the rust version never did this):
https://github.com/mozilla-services/autopush/blob/9f6c9fb/autopush/websocket.py#L1101

PR #478, opened 2023-10-24
bug: Convert WSError(Timeout for Pong) to metric

jrconlin:
Closes: SYNC-3971

### DIFF
--- a/autoconnect/autoconnect-web/src/error.rs
+++ b/autoconnect/autoconnect-web/src/error.rs
@@ -25,8 +25,7 @@ impl ResponseError for ApiError {
 
     fn error_response(&self) -> HttpResponse {
         let code = self.status_code();
-        let mut builder = autopush_common::errors::build_error(code);
-        builder.json(json!({
+        HttpResponse::build(code).json(json!({
             "code": code.as_u16(),
             "errno": self.errno(),
             "error": self.to_string(),
--- a/autoconnect/autoconnect-web/src/error.rs
+++ b/autoconnect/autoconnect-web/src/error.rs
@@ -25,7 +25,8 @@ impl ResponseError for ApiError {
 
     fn error_response(&self) -> HttpResponse {
         let code = self.status_code();
-        HttpResponse::build(code).json(json!({
+        let mut builder = autopush_common::errors::build_error(code);
+        builder.json(json!({
             "code": code.as_u16(),
             "errno": self.errno(),
             "error": self.to_string(),
@@ -41,7 +42,7 @@ impl ReportableError for ApiError {
     fn is_sentry_event(&self) -> bool {
         match self {
             // Ignore failing upgrade to WebSocket
-            ApiError::Actix(e) => e.as_error() != Some(&HandshakeError::NoWebsocketUpgrade),
+            ApiError::Actix(e) => e.as_error::<HandshakeError>().is_some(),
             _ => true,
         }
     }
--- a/autoconnect/autoconnect-web/src/error.rs
+++ b/autoconnect/autoconnect-web/src/error.rs
@@ -42,7 +42,7 @@ impl ReportableError for ApiError {
     fn is_sentry_event(&self) -> bool {
         match self {
             // Ignore failing upgrade to WebSocket
-            ApiError::Actix(e) => e.as_error::<HandshakeError>().is_some(),
+            ApiError::Actix(e) => e.as_error::<HandshakeError>().is_none(),
             _ => true,
         }
     }
--- a/autoconnect/autoconnect-ws/src/error.rs
+++ b/autoconnect/autoconnect-ws/src/error.rs
@@ -61,6 +61,7 @@ impl ReportableError for WSError {
     fn is_sentry_event(&self) -> bool {
         match &self.kind {
             WSErrorKind::SM(e) => e.is_sentry_event(),
+            WSErrorKind::Protocol(_) | WSErrorKind::RegistryDisconnected => true,
             _ => false,
         }
     }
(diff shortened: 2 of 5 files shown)


autoconnect/autoconnect-ws/src/error.rs
```diff
@@ -61,14 +61,18 @@ impl ReportableError for WSError {
...
+                WSErrorKind::PongTimeout | WSErrorKind::Json(_) | WSErrorKind::SessionClosed(_)
+            ),
         }
     }
 
     fn metric_label(&self) -> Option<&'static str> {
         match &self.kind {
             WSErrorKind::SM(e) => e.metric_label(),
             // TODO:
+            WSErrorKind::PongTimeout => Some("pong_timeout"),
```
pjenvey, line 75: Let's not even metric this for a couple reasons: legacy autopush didn't, they're not terribly important and we're collecting these error events in the session's single `info!` log message, where its`disconnect_reason` field would be set to "Timeout waiting for Pong".

Apparently this is even more verbose than legacy autopush which treats a number of common errors like `PongTimeout` as being no Error, leading to an empty string for `disconnect_reason`: https://github.com/mozilla-services/autopush-rs/blob/f8e6525/autopush/src/client.rs#L583

Maybe autoconnect should even eventually do the same? Basically consider timeouts/io errors and the like as not really errors for a long lived websocket connection.

```suggestion
```

pjenvey (changes requested):

autoconnect/autoconnect-ws/src/error.rs
jrconlin: I wasn't sure if there were more errors we should block, but it looks like that's the only one so far.

pjenvey (changes requested):
Sorry, let's add one more ^

pjenvey (changes requested):
and let's also hide all handhshake errors as a few of those started showing up in sentry:

```diff
diff --git a/autoconnect/autoconnect-web/src/error.rs b/autoconnect/autoconnect-web/src/error.rs
index 09929a7..c775c5e 100644
--- a/autoconnect/autoconnect-web/src/error.rs
+++ b/autoconnect/autoconnect-web/src/error.rs
@@ -40,8 +40,8 @@ impl ReportableError for ApiError {
 
     fn is_sentry_event(&self) -> bool {
         match self {
-            // Ignore failing upgrade to WebSocket
-            ApiError::Actix(e) => e.as_error() != Some(&HandshakeError::NoWebsocketUpgrade),
+            // Ignore various WebSocket handshake errors
+            ApiError::Actix(e) => e.as_error::<HandshakeError>().is_some(),
             _ => true,
         }
     }
```

autoconnect/autoconnect-ws/src/error.rs
jrconlin: Looking at this, I'm thinking that we should match rather than exclude, so I am saying it's a sentry event if we match:

WSErrorKind::SM(_) 
| WSErrorKind::Protocol(_)
| WSErrorKind::RegistryDisconnected

We're ignoring everything else. 

Sound good?

autoconnect/autoconnect-ws/src/error.rs
```diff
@@ -61,14 +61,13 @@ impl ReportableError for WSError {
     fn is_sentry_event(&self) -> bool {
         match &self.kind {
             WSErrorKind::SM(e) => e.is_sentry_event(),
-            e => !matches!(e, WSErrorKind::Json(_) | WSErrorKind::SessionClosed(_)),
+            _ => false,
```
pjenvey, line 65: weren't we going with https://github.com/mozilla-services/autopush-rs/pull/478#discussion_r1372239508?

autoconnect/autoconnect-web/src/error.rs
```diff
@@ -25,7 +25,8 @@ impl ResponseError for ApiError {
 
     fn error_response(&self) -> HttpResponse {
         let code = self.status_code();
-        HttpResponse::build(code).json(json!({
+        let mut builder = autopush_common::errors::build_error(code);
```
pjenvey, line 28: Can we keep these changes exclusive to #476? I briefly looked at it yesterday and had some thoughts on how much should even apply to autoconnect

autoendpoint/src/error.rs
```diff
@@ -296,8 +299,14 @@ impl ResponseError for ApiError {
     fn error_response(&self) -> HttpResponse {
         let mut builder = HttpResponse::build(self.kind.status());
 
-        if self.status_code() == 410 {
-            builder.insert_header(("Cache-Control", "max-age=86400"));
+        match self.status_code() {
+            StatusCode::GONE => {
+                builder.insert_header((header::CACHE_CONTROL, "max-age=86400"));
```
pjenvey, line 304: not blocking on this but FYI there's also this fancy way of doing it:

```suggestion
                builder.insert_header(CacheControl(vec![CacheDirective::MaxAge(86400)]));
```
```
use actix_web::{
    http::header::{CacheControl, CacheDirective}
};
```

autoendpoint/src/error.rs
```diff
@@ -10,7 +10,9 @@ use actix_web::{
     middleware::ErrorHandlerResponse,
     HttpResponse, Result,
 };
-use backtrace::Backtrace; // Sentry uses the backtrace crate, not std::backtrace.
+use backtrace::Backtrace;
+use reqwest::header;
+// Sentry uses the backtrace crate, not std::backtrace.
```
pjenvey, line 15: this comment got out of place, needs to move up 3 lines

autopush-common/src/errors.rs
```diff
@@ -9,7 +9,8 @@ use actix_web::{
     dev::ServiceResponse, http::StatusCode, middleware::ErrorHandlerResponse, HttpResponse,
     HttpResponseBuilder, ResponseError,
 };
-use backtrace::Backtrace; // Sentry 0.29 uses the backtrace crate, not std::backtrace
+use backtrace::Backtrace;
+// Sentry 0.29 uses the backtrace crate, not std::backtrace
```
pjenvey, line 13: same as the other file

pjenvey (changes requested):

pjenvey (approved):

PR #483, opened 2023-10-26
fix: capture sentry events for the unidentified state 

pjenvey:
and kill db settings defaults

SYNC-3975
SYNC-3976


pjenvey:
wrong branch

PR #484, opened 2023-10-26
fix: capture sentry events for the unidentified state

pjenvey:
and kill db settings defaults

SYNC-3975
SYNC-3976

### DIFF
--- a/autoconnect/autoconnect-ws/src/error.rs
+++ b/autoconnect/autoconnect-ws/src/error.rs
@@ -3,8 +3,8 @@ use std::fmt;
 use actix_ws::CloseCode;
 use backtrace::Backtrace;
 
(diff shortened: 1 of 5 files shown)


autoconnect/autoconnect-ws/src/error.rs
```diff
@@ -51,6 +51,18 @@ impl WSError {
     pub fn close_description(&self) -> &str {
         self.kind.as_ref()
     }
+
+    /// Emit an event for this Error to Sentry if set to
```
jrconlin, line 55: `set to`... ?

jrconlin (changes requested):

jrconlin (approved):

PR #491, opened 2023-10-30
fix: split router endpoints into their own app

pjenvey:
which binds solely to the router port

SYNC-3979

### DIFF
--- a/autoconnect/autoconnect-web/src/lib.rs
+++ b/autoconnect/autoconnect-web/src/lib.rs
@@ -17,6 +17,9 @@ use actix_web::web;
 #[macro_export]
 macro_rules! build_app {
     ($app_state: expr) => {
+        build_app!($app_state, $crate::config)
+    };
+    ($app_state: expr, $config: expr) => {
         actix_web::App::new()
             .app_data(actix_web::web::Data::new($app_state.clone()))
             .wrap(actix_web::middleware::ErrorHandlers::new().handler(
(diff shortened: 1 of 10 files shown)


autoconnect/src/main.rs
```diff
@@ -79,9 +82,24 @@ async fn main() -> Result<()> {
         "Starting autoconnect on port {} (router_port: {})",
         port, router_port
     );
-    HttpServer::new(move || build_app!(app_state))
-        .bind(("0.0.0.0", port))?
-        .bind(("0.0.0.0", router_port))?
+
+    let app_state2 = app_state.clone();
```
jrconlin, line 86: Even though it's essentially a temporary value, kinda wondering if we should name it `router_state` or something a bit more intentional.

jrconlin (dismissed):

autoconnect/autoconnect-web/src/lib.rs
```diff
@@ -17,6 +17,9 @@ use actix_web::web;
 #[macro_export]
 macro_rules! build_app {
     ($app_state: expr) => {
```
jrconlin, line 19: This kinda feels 'clever' to me, instead of just updating the `test` instance of `build_app!` to include `crate::config()` 
We're already modifying a bunch of similar code already.

pjenvey: That's exactly why I did it, so I didn't have to change it anywhere else: but good call, it's only 2 places

pjenvey:
I also audited any actix-ws use of `app_info`. Obviously it only references anything from `HttpRequest` (the likely place where app_info would be used) during the WS handshake.

The handshake's handled by `actix-http::ws`, which only uses `RequestHead` to, as one would imagine, look at headers.

jrconlin (approved):

PR #495, opened 2023-10-31
feat: handle already connected users

pjenvey:
SYNC-3980

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -511,8 +511,10 @@ impl DbClient for BigTableClientImpl {
     /// BigTable doesn't really have the concept of an "update". You simply write the data and
     /// the individual cells create a new version. Depending on the garbage collection rules for
     /// the family, these can either persist or be automatically deleted.
-    async fn update_user(&self, user: &User) -> DbResult<()> {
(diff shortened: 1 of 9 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -511,8 +511,10 @@ impl DbClient for BigTableClientImpl {
     /// BigTable doesn't really have the concept of an "update". You simply write the data and
     /// the individual cells create a new version. Depending on the garbage collection rules for
     /// the family, these can either persist or be automatically deleted.
-    async fn update_user(&self, user: &User) -> DbResult<()> {
-        self.add_user(user).await
+    async fn update_user(&self, user: &User) -> DbResult<bool> {
+        self.add_user(user).await?;
+        // TODO: is a conditional check possible?
```
jrconlin, line 516: Will add this as a TODO:
Bigtable wants to overwrite, so we may need to do a pre-emptive read first. I can contain that to the `add_user()` function.

jrconlin (approved):
I was debating about making the autoendpoint conditional errors more explicit but I think the traceback should suffice.

jrconlin (approved):
I was debating about making the autoendpoint conditional errors more explicit but I think the traceback should suffice.

PR #500, opened 2023-11-01
fix: add ReportableError::reportable_source

pjenvey:
for event_from_error to workaround its issues w/ downcast_ref

SYNC-3981

### DIFF
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/error.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/error.rs
@@ -112,7 +112,7 @@ pub async fn __test_sm_reqwest_error() -> SMError {
         .https_only(true)
         .build()
         .unwrap()
-        .get("http://foo.com")
+        .get("http://example.com")
         .send()
         .await
         .unwrap_err();
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/error.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/error.rs
@@ -103,3 +103,18 @@ pub enum SMErrorKind {
     #[error("Client sent too many pings too often")]
     ExcessivePing,
 }
+
+#[cfg(debug_assertions)]
+/// Return a [SMErrorKind::Reqwest] [SMError] for tests
+pub async fn __test_sm_reqwest_error() -> SMError {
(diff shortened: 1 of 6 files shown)


autoconnect/autoconnect-ws/autoconnect-ws-sm/src/error.rs
```diff
@@ -103,3 +103,18 @@ pub enum SMErrorKind {
...
+
+#[cfg(debug_assertions)]
+/// Return a [SMErrorKind::Reqwest] [SMError] for tests
+pub async fn __test_sm_reqwest_error() -> SMError {
+    // An easily constructed reqwest::Error
+    let e = reqwest::Client::builder()
+        .https_only(true)
+        .build()
+        .unwrap()
+        .get("http://foo.com")
```
jrconlin, line 115: Let's use `example.com` here. (Yeah, I know, but it's good sanitary practice.)

autopush-common/src/errors.rs
```diff
@@ -181,6 +181,12 @@ impl ApcErrorKind {
 
 /// Interface for reporting our Error types to Sentry or as metrics
 pub trait ReportableError: std::error::Error {
+    /// Like [Error::source] but the lower-level [ReportableError] source of
```
jrconlin, line 184:
```suggestion
    /// Like [Error::source] but selectively return the lower-level [ReportableError] source of one of the inner components of
```

autopush-common/src/sentry.rs
```diff
@@ -17,21 +17,25 @@ pub fn client_options() -> sentry::ClientOptions {
 ///
 /// `std::error::Error` doesn't support backtraces, thus `sentry::event_from_error`
 /// doesn't either. This function works against `ReportableError` instead to
-/// access its backtrace.
-pub fn event_from_error<E>(err: &E) -> sentry::protocol::Event<'static>
-where
-    E: ReportableError + 'static,
-{
-    let mut exceptions = vec![exception_from_error_with_backtrace(err)];
+/// access it and its `reportable_source's` backtraces.
```
jrconlin, line 20:
```suggestion
/// access it and extract  backtraces from its chain of `reportable_source`s.
```

jrconlin (changes requested):
Feel free to ignore the comment nits if you like.

jrconlin (approved):

PR #503, opened 2023-11-03
bug: Add proper update_user logic for Bigtable 

jrconlin:
Closes: SYNC-3982


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -505,9 +546,53 @@ impl DbClient for BigTableClientImpl {
...
+        type_filter.set_value_regex_filter(user.router_type.as_bytes().to_vec());
+        filter_set.push(type_filter);
+
+        // then check to make sure that the last connected_at time is before this one.
+        let mut connected_filter = RowFilter::default();
+        connected_filter.set_family_name_regex_filter(ROUTER_FAMILY.to_owned());
+        connected_filter
+            .set_column_qualifier_regex_filter("connected_at".to_owned().as_bytes().to_vec());
+        let mut val_range = ValueRange::default();
+        val_range.set_start_value_closed(user.connected_at.to_be_bytes().to_vec());
```
pjenvey, line 586: _the new `connected_at` time is later than the existing `connected_at`_

doesn't that mean a:

```suggestion
        val_range.set_end_value_closed(user.connected_at.to_be_bytes().to_vec());
```
A test for this would be great (I'm not sure anything like the integration test actually test this) but I'm ok adding one after this PR if it's too involved.

pjenvey (changes requested):

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: Sigh. I remember this. It's completely counter intuitive, and I'll add a comment to address it.
What the filter is doing is ensuring that the value that we will provide is ranged to be inclusive of the provided timestamp, not that the existing value is below that timestamp. Honestly, I shouldn't use this filter. 

I'm not sure if there's a way to specify "Only modify if this condition is present".

jrconlin: Turns out, there is. Thanks @pjenvey.

pjenvey (changes requested):
With the changes on [this branch](https://github.com/mozilla-services/autopush-rs/compare/bug/SYNC-3982-bt_reg...bug/SYNC-3982-bt_reg-pjenvey2) we can continue using be bytes for `connected_at` like the other numeric types

pjenvey (approved):

PR #506, opened 2023-11-03
feat: Add metrics for user creation/deletion.

jrconlin:
There were a few questions raised about possible UAID churn occuring with autoconnect. Adding some metrics

Add new metrics for:
* `ua.expiration`  - a UAID has been reset for
  * `reason`:(`reset_uaid`|`too_many_messages`|`no_current_mo`|`invalid_current_mo`)
  * added `autoconnect`:`true` for consistency
* `ua.existing_user` - A user was found in the database
* `ua.new_user` - no record was found, assinging a new UAID.
  * `reassined`: Was a UAID offered by the UA?

Closes: SYNC-3994

### DIFF
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
@@ -183,7 +183,7 @@ impl WebPushClient {
         // Log out the final stats message
         info!("Session";
             "uaid_hash" => self.uaid.as_simple().to_string(),
-            "uaid_reset" => self.flags.reset_uaid,
+            "uaid_reset" => self.flags.old_record_version,
             "existing_uaid" => stats.existing_uaid,
             "connection_type" => "webpush",
             "ua_name" => &ua_info.browser_name,
@@ -321,7 +321,7 @@ pub async fn process_existing_user(
 
     let flags = ClientFlags {
         check_storage: true,
-        reset_uaid: user
+        old_record_version: user
             .record_version
             .map_or(true, |rec_ver| rec_ver < USER_RECORD_VERSION),
         ..Default::default()
@@ -338,7 +338,7 @@ pub struct ClientFlags {
     /// Whether this client needs to check storage for messages
     check_storage: bool,
     /// Flags the need to drop the user record
-    pub(crate) reset_uaid: bool,
+    old_record_version: bool,
 }
 
 impl Default for ClientFlags {
@@ -347,7 +347,7 @@ impl Default for ClientFlags {
             include_topic: true,
             increment_storage: false,
             check_storage: false,
-            reset_uaid: false,
+            old_record_version: false,
         }
     }
 }
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
@@ -298,7 +298,7 @@ pub async fn process_existing_user(
                 .incr_with_tags("ua.expiration")
                 .with_tag("errno", "104")
                 .with_tag("autoconnect", "true")
-                .with_tag("reason", "no_current_mo")
+                .with_tag("reason", "no_current_month")
                 .send();
             app_state.db.remove_user(&user.uaid).await?;
             return Ok(None);
@@ -314,7 +314,7 @@ pub async fn process_existing_user(
                 .incr_with_tags("ua.expiration")
                 .with_tag("errno", "105")
                 .with_tag("autoconnect", "true")
-                .with_tag("reason", "invalid_current_mo")
+                .with_tag("reason", "invalid_current_month")
                 .send();
             app_state.db.remove_user(&user.uaid).await?;
             return Ok(None);
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
@@ -340,7 +340,7 @@ pub struct ClientFlags {
     /// Whether this client needs to check storage for messages
     check_storage: bool,
     /// Flags the need to drop the user record
-    reset_uaid: bool,
+    pub(crate) reset_uaid: bool,
 }
 
 impl Default for ClientFlags {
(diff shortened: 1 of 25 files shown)


autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
```diff
@@ -297,6 +297,8 @@ pub async fn process_existing_user(
                 .metrics
                 .incr_with_tags("ua.expiration")
                 .with_tag("errno", "104")
+                .with_tag("autoconnect", "true")
+                .with_tag("reason", "no_current_mo")
```
pjenvey, line 301: None of the `autoconnect: true` are necessary, that's already emitted for all metrics. Might as well name these `_month`.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -124,6 +124,11 @@ impl UnidentifiedClient {
                         let _ = self.app_state.metrics.incr("ua.already_connected");
                         return Err(SMErrorKind::AlreadyConnected.into());
                     }
+                    self.app_state
+                        .metrics
+                        .incr_with_tags("ua.existing_user")
+                        .with_tag("autoconnect", "true")
+                        .send();
```
pjenvey, line 131: Let's just add a tag onto `ua.command.hello` instead of adding both of these

pjenvey (changes requested):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
jrconlin: Yeah, truncated these partly because that's about where they truncate on my display, and I felt it provided enough detail.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
```diff
@@ -297,6 +297,8 @@ pub async fn process_existing_user(
                 .metrics
                 .incr_with_tags("ua.expiration")
                 .with_tag("code", "104")
+                .with_tag("autoconnect", "true")
```
pjenvey, line 300: etc

```suggestion
```

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -67,19 +67,27 @@ impl UnidentifiedClient {
...
-        let _ = self.app_state.metrics.incr("ua.command.hello");
+        self.app_state
+            .metrics
+            .incr_with_tags("ua.command.hello")
+            .with_tag("existing_user", &original_uaid.is_some().to_string())
+            .with_tag(
+                "reassigned",
+                &(original_uaid.unwrap_or(uaid) != uaid).to_string(),
+            )
+            .send();
```
pjenvey, line 90: With all reassigned being new users I think one tag should cover it all instead, e.g. `uaid`: `existing` | `new` | `reassigned`

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -67,19 +67,27 @@ impl UnidentifiedClient {
...
         let uaid = user.uaid;
         debug!(
             "💬UnidentifiedClient::on_client_msg Hello! uaid: {} existing_user: {}",
             uaid, existing_user,
         );
-        let _ = self.app_state.metrics.incr("ua.command.hello");
+        self.app_state
+            .metrics
+            .incr_with_tags("ua.command.hello")
+            .with_tag("existing_user", &original_uaid.is_some().to_string())
```
pjenvey, line 85:
```suggestion
            .with_tag("existing_user", &existing_user.to_string())
```
Otherwise this is subtlety changing the meaning of `existing_user` here.

pjenvey (changes requested):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
jrconlin: I think you meant to say that all reassigned users were existing users, but yeah, I'll collapse these into one tag.

pjenvey: Nope: they would be new (`existing_user` would be false), their new uaids generated from `Default::default()` in `get_or_create_user`'s fallback case

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -79,7 +79,19 @@ impl UnidentifiedClient {
...
             uaid, existing_user,
         );
-        let _ = self.app_state.metrics.incr("ua.command.hello");
+        self.app_state
+            .metrics
+            .incr_with_tags("ua.command.hello")
+            .with_tag("uaid", {
+                if existing_user {
+                    "existing"
+                } else if flags.reset_uaid {
```
pjenvey, line 88: We want what you had previously for reassigned (comparing the uaids), `reset_uaid` is something else. It's only triggered when that old `record_version` field in the db is bad/invalid, which is probably never at this point.

Technically this flag is a form of a "reassign" but it's delayed: the flag triggers the user to be dropped after it checks/acks all its messages from storage. So I'd argue its "reassign" doesn't occur in this hello'd session w/ the flag set but instead during the following session after the user drop/disconnection.

Have we ever used/bumped `record_version` (it's currently at `1`, was it ever `0` or non-existent?) or has it always been theoretical?

pjenvey (changes requested):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
jrconlin: IIRC, `record_version` is there in case we ever decided to radically change the record. There was a change we did a long time ago, which may have introduced this field.

autoconnect/src/main.rs
```diff
@@ -83,6 +83,8 @@ async fn main() -> Result<()> {
         port, router_port
     );
 
+    actix_web::App::new();
+
```
pjenvey, line 87: Leftover from something else?

```suggestion
```

pjenvey (changes requested):

autoconnect/src/main.rs
jrconlin: no idea where that came from and had to pull to see it. 

Should be gone now.

pjenvey (approved):

PR #510, opened 2023-11-06
feat: demote `retry` message, dump vapid claims.

jrconlin:
This patch demotes the `retryable` `error!` message since it's not actionable.

It also introduces a dump of the `VAPID` claims, extracting the `sub` from the header for easy parsing.

Closes: SYNC-3996

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.69.7"
+version = "1.69.1"
 authors = [
   "Ben Bangert <ben@groovie.org>",
   "JR Conlin <jrconlin@mozilla.com>",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.69.1"
+version = "1.69.7"
 authors = [
   "Ben Bangert <ben@groovie.org>",
   "JR Conlin <jrconlin@mozilla.com>",
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -8,6 +8,7 @@ use autopush_common::{
     tags::Tags,
     util::{b64_decode_std, b64_decode_url, sec_since_epoch},
 };
+use base64::Engine;
 use cadence::{CountedExt, StatsdClient};
 use futures::{future::LocalBoxFuture, FutureExt};
 use jsonwebtoken::{Algorithm, DecodingKey, Validation};
@@ -305,9 +306,7 @@ fn validate_vapid_jwt(
     };
 
     // Dump the claims.
-    let claims = vapid.token.split('.').collect::<Vec<&str>>();
-    if let Some(claims_str) = claims.get(1) {
-        use base64::Engine;
+    if let Some(claims_str) = vapid.token.split('.').next() {
         info!(
             "Vapid";
             "sub" => &token_data.claims.sub,
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -8,7 +8,6 @@ use autopush_common::{
     tags::Tags,
     util::{b64_decode_std, b64_decode_url, sec_since_epoch},
 };
-use base64::Engine;
 use cadence::{CountedExt, StatsdClient};
 use futures::{future::LocalBoxFuture, FutureExt};
 use jsonwebtoken::{Algorithm, DecodingKey, Validation};
@@ -306,7 +305,10 @@ fn validate_vapid_jwt(
     };
 
     // Dump the claims.
+    // Note, this can produce a LOT of log messages if this feature is enabled.
+    #[cfg(log_vapid)]
     if let Some(claims_str) = vapid.token.split('.').next() {
+        use base64::Engine;
         info!(
             "Vapid";
             "sub" => &token_data.claims.sub,
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::fmt;
 use std::str::FromStr;
 
 use actix_web::{dev::Payload, web::Data, FromRequest, HttpRequest};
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -13,7 +13,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.69.1"
+version = "1.69.7"
```
jrconlin, line 16: Not sure why this isn't getting updated to reflect releases. 
If we go to a different cadence, we should maybe revisit this as a release date designation?

autoendpoint/src/extractors/subscription.rs
```diff
@@ -54,6 +55,15 @@ impl Default for VapidClaims {
     }
 }
 
+impl fmt::Display for VapidClaims {
```
jrconlin, line 58: Technically not needed for this release, and I'm fine pulling it, but it's still a useful thing to have for debugging and development.

autoendpoint/src/extractors/subscription.rs
```diff
@@ -294,6 +304,22 @@ fn validate_vapid_jwt(
...
+        use base64::Engine;
+        info!(
+            "Vapid";
+            "sub" => &token_data.claims.sub,
+            "claims" => String::from_utf8(
+                base64::engine::general_purpose::URL_SAFE_NO_PAD
+                    .decode(claims_str)
+                    .unwrap_or_default()
+            )
+            .unwrap_or("UNKNOWN".to_owned())
```
jrconlin, line 319: Technically, it's possible for the decode to fail due to `base64` pendency, but if that were the case, we probably can discard the content anyway.

Cargo.toml
pjenvey: I just haven't merged the release/1.69 branch yet because we've been kinda rapid firing releases. Leave this out and I'll merge it after the next release

autoendpoint/src/extractors/subscription.rs
pjenvey: Since this isn't even used by the new log statement (right?) it's fine.

autoendpoint/src/extractors/subscription.rs
```diff
@@ -294,6 +304,22 @@ fn validate_vapid_jwt(
         },
     };
 
+    // Dump the claims.
+    let claims = vapid.token.split('.').collect::<Vec<&str>>();
```
pjenvey, line 308: nit: this doesn't need to collect, I think you can just look at `next()`

autoendpoint/src/extractors/subscription.rs
```diff
@@ -294,6 +304,22 @@ fn validate_vapid_jwt(
...
+    if let Some(claims_str) = claims.get(1) {
+        use base64::Engine;
+        info!(
```
pjenvey, line 311: What's the goal of this one? This won't emit to the logs currently if that was the intention because autoendpoint only logs at the warn level.

I know we've talked about emitting more such data in the future but I'm not sure exactly what and where to emit it yet -- as is this seems like it would too chatty (if it were a `warn!`).

autoendpoint/src/extractors/subscription.rs
```diff
@@ -294,6 +304,22 @@ fn validate_vapid_jwt(
         },
     };
 
+    // Dump the claims.
+    let claims = vapid.token.split('.').collect::<Vec<&str>>();
+    if let Some(claims_str) = claims.get(1) {
+        use base64::Engine;
```
pjenvey, line 310: should move this to the top

pjenvey (changes requested):

autoendpoint/src/extractors/subscription.rs
jrconlin: No, it's not used by the new log statement. I did use it for some internal things while I was working on stuff and realized it'd be useful to leave.

autoendpoint/src/extractors/subscription.rs
jrconlin: Huh. Doing a BigQuery search on `...autopush_prod_log.stdout` shows that we're capturing log messages with `INFO` severity, (`jsonPayload.severity = 6.0`) for the Session info. I presume I'm looking at the wrong ones?

autoendpoint/src/extractors/subscription.rs
jrconlin: Yeah, I was wondering about that. 
This is the only location where we're doing base64, so I wasn't sure if defining things here would let us scope this as a feature.

autoendpoint/src/extractors/subscription.rs
jrconlin: As for the goal of this, I view it as the same sort of data we collect for the session info. We'd dump it to stdout and query it using Bigquery.

pjenvey: autoconnect is set to info so you may only be seeing those (or maybe our [severity is not trustworthy?](https://github.com/mozilla-services/slog-mozlog-json/pull/7/files))

jrconlin: Perhaps, but I got those values [from the Google docs](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity)

That said, the values we're seeing in the GCP logs do not appear to be the same ones we're generating here (The ones in the logs appear to be floats, these are integers), so maybe the docs are wrong and some experimentation is in order.

pjenvey: But per the metric emitted a little later in this method we've had about 2k vapid notifications per _second_ emitted around the time that I wrote this which isn't even peak.

We can merge this as is since it won't output to logs yet, but it probably makes more sense as a `debug!` for now.

jrconlin: Fair. I'll demote to debug as well for this, just to keep things safer.

jrconlin: Meh, if we're not using this, throw it behind a feature flag.

autoendpoint/src/extractors/subscription.rs
pjenvey: I tend to keep them at the top anyway, just duplicating the cfg(feature) there as well

pjenvey (approved):

autoendpoint/src/extractors/subscription.rs
jrconlin: Yeah, for larger things, I totally see that. This is really self contained since it's the only location we're using it, and it's one less line to chase if we decide to pull it entirely later.

PR #515, opened 2023-11-09
feat: add an actix_max_connections setting

pjenvey:
and print the available_parallelism on startup

SYNC-4005

### DIFF
--- a/autoconnect/autoconnect-settings/src/lib.rs
+++ b/autoconnect/autoconnect-settings/src/lib.rs
@@ -61,8 +61,6 @@ pub struct Settings {
     /// How long to wait for a response Pong before being timed out and connection drop
     #[serde(deserialize_with = "deserialize_f64_to_duration")]
     pub auto_ping_timeout: Duration,
-    /// Max number of websocket connections to allow
-    pub max_connections: u32,
     /// How long to wait for the initial connection handshake.
     #[serde(deserialize_with = "deserialize_u32_to_duration")]
     pub open_handshake_timeout: Duration,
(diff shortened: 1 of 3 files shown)


autoconnect/autoconnect-settings/src/lib.rs
```diff
@@ -99,10 +97,11 @@ pub struct Settings {
...
     pub msg_limit: u32,
-    /// Maximum number of pending notifications for individual UserAgent handlers.
-    /// (if a given [autoconnect-common::RegisteredClient] receives more than this number, the calling
-    /// thread will lock.)
-    pub max_pending_notification_queue: u32,
+    /// Sets the maximum number of concurrent connections per actix-web worker.
+    ///
+    /// All socket listeners will stop accepting connections when this limit is
+    /// reached for each worker.
+    pub actix_max_connections: Option<usize>,
```
jrconlin, line 104: Let's add a setting and support for actix server [workers](https://docs.rs/actix-server/latest/actix_server/struct.ServerBuilder.html#method.workers) as well, since I don't know if the kubernetes core limit is sufficiently differentiated from whatever rust is using to read the core count. (`grep processor /proc/cpuinfo | wc -l`?)

At least that way, we can know for sure.

jrconlin (changes requested):

autoconnect/autoconnect-settings/src/lib.rs
pjenvey: it's all based on `available_parallelism` so we can look at the log statement when this goes out. Added workers

jrconlin (approved):

PR #518, opened 2023-11-14
chore: lint & make integration tests in autopush-rs

Trinaa:
- move the integration test code to the tests\integration folder
- convert the integration test dependency manager to poetry (merge with load test pyproject.toml)
- lint integration test code
- update make to lint and execute integration tests
- update CI to lint integration test code

[SYNC-3833](https://mozilla-hub.atlassian.net/browse/SYNC-3833)

[SYNC-3833]: https://mozilla-hub.atlassian.net/browse/SYNC-3833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrbenny35:
Is there any way to split this? +2000 changes is a lot. Guess we could just trust the CI build and our local testing

tests/.flake8
```diff
@@ -2,4 +2,5 @@
 exclude =
     .git,
     __pycache__,
-max-line-length = 99
\ No newline at end of file
+ignore = E203
```
jrconlin, line 5: Might be nice to include that this ignores whitespace before `:` on list.

tests/integration/db.py
```diff
@@ -70,19 +58,73 @@ class ItemNotFound(Exception):
...
+
+    def __getattr__(self, name):
+        return getattr(self._table, name)
+
+
 # Autopush utils:
 def generate_hash(key: bytes, payload: bytes) -> str:
     """Generate a HMAC for the uaid using the secret
 
     :returns: HMAC hash and the nonce used as a tuple (nonce, hash).
```
jrconlin, line 121: Heh, I thought that function docstrings end with a blank line, but I see neither [PEP8](https://peps.python.org/pep-0008/#documentation-strings) and [PEP257](https://peps.python.org/pep-0257/#multi-line-docstrings) do that. 

TIL.

tests/integration/db.py
```diff
@@ -133,21 +172,19 @@ def get_month(delta=0):
...
 
 
 def create_message_table_ddb(
-    tablename,  # type: str
-    read_throughput=5,  # type: int
-    write_throughput=5,  # type: int
-    boto_resource=None,  # type: DynamoDBResource
-):
-    # type: (...) -> Any  # noqa
+    boto_resource: DynamoDBResource,
```
jrconlin, line 183: Not a huge fan of changing the order of parameters, but this is a limited use library. 

Still, it's a good idea to keep order and only append new args to the end of a given function declaration.

tests/integration/test_integration_all_rust.py
```diff
@@ -9,96 +9,105 @@
...
+    "autopush",
+    "autopush_common",
+]
 log_string = [f"{x}=trace" for x in modules]
-RUST_LOG = ",".join(log_string)+",error"
+RUST_LOG = ",".join(log_string) + ",error"
 
 
-def get_free_port() -> int:
+def get_free_port() -> Any:
```
jrconlin, line 96: Are there instances where the `port` is not an integer?

jrconlin (approved):
Some nits and a few questions, but otherwise 👍🏻

tests/integration/test_integration_all_rust.py
```diff
@@ -214,23 +220,22 @@ def __getattribute__(self, name: str):
...
             url = "ws://localhost:{}/".format(connection_port)
         self.ws = websocket.create_connection(url, header=self.headers)
         return self.ws.connected
 
-    def hello(
-        self, uaid: Optional[str] = None, services: Optional[list[str]] = None
-    ):
+    def hello(self, uaid: str | None = None, services: list[str] | None = None):
+        if not self.ws:
+            raise Exception("WebSocket client not available as expected")
```
pjenvey, line 232: Is this check necessary? I ask because there's a number of other methods using `self.ws` that could probably also use it, if so -- then it's probably worth a refactor into an `ensure_ws` method.

pjenvey:
> Is there any way to split this? +2000 changes is a lot. Guess we could just trust the CI build and our local testing

It's really not that large, this large diff stat is mostly due to git not recognizing the `poetry.lock` path change as a rename.

tests/integration/db.py
Trinaa: I 100% agree, but made a judgment, very much with the understanding that the impact was contained. 

**Background**
Previously, the `boto_resource` was declared as an Optional parameter in many places which defaulted to `None` in the method declaration. This causes mypy to complain, i think rightly so, when we proceed to use `boto_resource` without checking if it is `None`. 

Option 1 was to add null checks, which added unnecessary complexity in my opinion. So I opted for Option 2, to remove the optionality for `boto_resource`, but then I ran into the limitation that non-default method arguments need to be listed before defaulted arguments. I propagated this change in order across methods to be consistent, but perhaps that's too much.

tests/integration/test_integration_all_rust.py
Trinaa: sigh*
image
mypy gymnastics see https://github.com/mozilla-services/autopush-rs/pull/518/commits/12ae9f015398b0ba897e8a7aedf52d223e62ea57

tests/integration/test_integration_all_rust.py
Trinaa: So I added the check to appease mypy, because "technically" self.ws can evaluate to None and we operate with an assumption that it won't be None.

I like the ensure_ws idea. Like this right?
image
Unfortunately mypy doesn't accept this though
image

pjenvey: Ok, totally disregard my comment. This is just silly mypy implementation details that aren't worth bothering with for now

tests/integration/db.py
Trinaa: I admit that some linting magic must have removed those lines.

jrconlin (approved):

pjenvey (approved):

PR #519, opened 2023-11-17
chore: report test results in CI 

Trinaa:
- Test results are output in junit-xml format and uploaded to CircleCI
- Fix the following error when installing python dependencies
  image
- Remove the python caching from the test CI job
    - The paths set by the `save_cache` step need to be updated to the following, however the performance gains are negligible, so that I recommend removing the steps.
    ```
          paths:
            - /root/.cache/pip
            - /root/.cache/pypoetry
    ```

[SYNC-3934](https://mozilla-hub.atlassian.net/browse/SYNC-3934)

[SYNC-3934]: https://mozilla-hub.atlassian.net/browse/SYNC-3934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,9 @@ jobs:
           name: Restoring Rust cache
           key: rust-v1-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
       - run:
+          name: Create Workspace
+          command: mkdir -p workspace
+      - run:
           name: Set up system
           command: |
             apt update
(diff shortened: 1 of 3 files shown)


jrbenny35 (approved):

.circleci/config.yml
```diff
@@ -128,15 +128,15 @@ jobs:
...
           name: Integration tests (Autoconnect)
           command: make integration-test
-      - save_cache:
-          name: Save Python cache
-          key: python-v1-{{ checksum "tests/poetry.lock" }}-{{ checksum "tests/integration/test_integration_all_rust.py"}}
-          paths:
-            - /home/circleci/.local/bin/
-            - /home/circleci/.local/lib/
+          environment:
+            TEST_RESULTS_DIR: workspace/test-results
```
pjenvey, line 137: Maybe these are here intentionally to be explicit but I'll note they match the default, so they're not necessary

pjenvey (approved):

PR #521, opened 2023-11-21
bug: remove unneeded clone operation

jrconlin:
Closes SYNC-3836

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/row.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/row.rs
@@ -15,15 +15,14 @@ pub struct Row {
 
 impl Row {
     /// Return all cells for a given column
-    pub fn get_cells(&self, column: &str) -> Option<Vec<Cell>> {
-        self.cells.get(column).cloned()
+    pub fn get_cells(&mut self, column: &str) -> Option<Vec<Cell>> {
+        self.cells.remove_entry(column).map(|(_, cell)| cell)
     }
 
     /// get only the "top" cell value. Ignore other values.
-    /// TODO: https://mozilla-hub.atlassian.net/browse/SYNC-3836
     pub fn get_cell(&mut self, column: &str) -> Option<Cell> {
-        if let Some(cells) = self.cells.get(column) {
-            return cells.last().cloned();
+        if let Some((_, mut cells)) = self.cells.remove_entry(column) {
+            return cells.pop();
         }
         None
     }
--- a/autopush-common/src/db/bigtable/bigtable_client/row.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/row.rs
@@ -15,12 +15,12 @@ pub struct Row {
 
 impl Row {
     /// Return all cells for a given column
-    pub fn get_cells(&mut self, column: &str) -> Option<Vec<Cell>> {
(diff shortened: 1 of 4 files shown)


autopush-common/src/db/bigtable/bigtable_client/row.rs
```diff
@@ -15,15 +15,14 @@ pub struct Row {
 
 impl Row {
     /// Return all cells for a given column
-    pub fn get_cells(&self, column: &str) -> Option<Vec<Cell>> {
-        self.cells.get(column).cloned()
+    pub fn get_cells(&mut self, column: &str) -> Option<Vec<Cell>> {
+        self.cells.remove_entry(column).map(|(_, cell)| cell)
```
pjenvey, line 19:
```suggestion
        self.cells.remove(column)
```

autopush-common/src/db/bigtable/bigtable_client/row.rs
```diff
@@ -15,15 +15,14 @@ pub struct Row {
...
     }
 
     /// get only the "top" cell value. Ignore other values.
-    /// TODO: https://mozilla-hub.atlassian.net/browse/SYNC-3836
     pub fn get_cell(&mut self, column: &str) -> Option<Cell> {
-        if let Some(cells) = self.cells.get(column) {
-            return cells.last().cloned();
+        if let Some((_, mut cells)) = self.cells.remove_entry(column) {
+            return cells.pop();
```
pjenvey, line 25:
```suggestion
        if let Some(mut cells) = self.cells.remove(column) {
            return cells.pop();
```

autopush-common/src/db/bigtable/bigtable_client/row.rs
```diff
@@ -15,15 +15,14 @@ pub struct Row {
 
 impl Row {
     /// Return all cells for a given column
-    pub fn get_cells(&self, column: &str) -> Option<Vec<Cell>> {
-        self.cells.get(column).cloned()
+    pub fn get_cells(&mut self, column: &str) -> Option<Vec<Cell>> {
```
pjenvey, line 18: I think these methods should be renamed to `remove_cells/cell` to be clearer to the caller

pjenvey (changes requested):

autopush-common/src/db/bigtable/bigtable_client/row.rs
jrconlin: Perhaps, but the function is to return the content of the cell, not just delete them.

pjenvey: But it can be a leaky abstraction potentially requiring the caller to care: because `get` method suggests it could be called repeatedly w/ the same result (idempotent). I'd also argue it should care at this level where a significant amount of data is shuffling around, thus the ownership of it vs where it's cloned should be clear.

In fact a good example of this is the grpc generated structs, offering a `get_`  method returning a reference and a `take_` version that similarly "removes" it via `mem::replace` or `take`. Renaming these methods to `take_cells/cell` instead is probably even clearer to the caller than `remove` since it follows that same pattern, how does that sound?

jrconlin: Yep, I'm fine with renaming these `take_*`

pjenvey (approved):

PR #522, opened 2023-11-21
feat: remove `use_webpush` as a required flag

jrconlin:
Closes SYNC-3845


autoconnect/autoconnect-common/src/protocol.rs
```diff
@@ -95,8 +93,6 @@ pub enum ServerMessage {
     Hello {
         uaid: String,
         status: u32,
-        #[serde(skip_serializing_if = "Option::is_none")]
-        use_webpush: Option<bool>,
```
pjenvey, line 99: Unfortunately we need to continue sending it in the response as Desktop expects it:

https://searchfox.org/mozilla-central/rev/d6b24b15cd275874c114f2772cf6fdb0d2045847/dom/push/PushServiceWebSocket.sys.mjs#608

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -277,7 +272,6 @@ mod tests {
         let msg = ClientMessage::Hello {
             uaid: Some("".to_owned()),
             channel_ids: None,
-            use_webpush: Some(true),
```
pjenvey, line 280: It's maybe worth keeping an additional test that still sends `use_webpush` via `json!` even though the integration test continues to sending it.

pjenvey (changes requested):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -262,12 +264,12 @@ mod tests {
...
-            channel_ids: None,
-            use_webpush: Some(true),
-            broadcasts: None,
-        };
+        // Ensure that we do not need to pass the "use_webpush" flag.
+        // (yes, this could just be passing the string, but I want to be
+        // very explicit here.)
+        let json = serde_json::json!({"messageType":"hello"});
+        let raw = json.to_string();
+        let msg = ClientMessage::from_str(&raw).unwrap();
```
pjenvey, line 272: This seems unnecessary with the type itself enforcing this -- it's very clear from the explicit construction of the struct variant in the previous code of what's actually being passed

autopush/src/client.rs
```diff
@@ -518,7 +515,6 @@ where
                     Ok(ServerMessage::Hello {
                         uaid: uaid.as_simple().to_string(),
                         status: 200,
-                        use_webpush: Some(true),
```
pjenvey, line 521: This needs to be put back like we just did for autoconnect

autopush/src/server/protocol.rs
```diff
@@ -95,8 +93,6 @@ pub enum ServerMessage {
     Hello {
         uaid: String,
         status: u32,
-        #[serde(skip_serializing_if = "Option::is_none")]
-        use_webpush: Option<bool>,
```
pjenvey, line 99: Same as above

tests/integration/test_integration_all_rust.py
```diff
@@ -909,7 +909,8 @@ def test_hello_echo(self):
         yield client.connect()
         result = yield client.hello()
         assert result != {}
-        assert result["use_webpush"] is True
```
pjenvey, line 912: These are all asserting the field exists in the responses so I think we want to put all these back (basically revert the whole file) as well

pjenvey (changes requested):

autopush/src/client.rs
jrconlin: If we are currently 100% on autoconnect, shouldn't we drop the code related to autopush entirely? 

I'm fine reverting this change here, but it's more for minor historical reasons.

pjenvey: We're definitely dropping it at [some point](https://mozilla-hub.atlassian.net/browse/SYNC-4036), I'd like to wait a little bit before doing so just to be on the safe side.

I'm also fine w/ 100% reverting everything in the autopush crate on this PR and pretty much not touching anything in there going forward.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
jrconlin: Right, but using the struct variant requires that fields be present, correct? Passing JSON like this would ensure that default values are being set by the constructor rather than being defined at initialization. This is essentially an even more stripped down version of the structure that's handed in with the above function where we specify all the fields as per the old format. Here, we don't include the previously required `use_webpush` as well as skip the `uaid` and rely on default management for the rest of the fields. 
This should be a properly accepted structure, correct? Shouldn't we test for that?

pjenvey (approved):

PR #523, opened 2023-11-23
chore: replace AUTOPUSH_ENDPOINT_URL and AUTOPUSH_SERVER_URL with LOCUST_HOST in load tests

Trinaa:
- Replace AUTOPUSH_ENDPOINT_URL and AUTOPUSH_SERVER_URL with LOCUST_HOST in load tests
- Update Python dependencies
- Update stage server URL in load tests
- Update Kubernetes definitions to mitigate cluster creation errors
- Re-calibrate load tests after GCP migration

Closes [SYNC-3974](https://mozilla-hub.atlassian.net/browse/SYNC-3974), Closes [SYNC-4034](https://mozilla-hub.atlassian.net/browse/SYNC-4034)

[SYNC-4034]: https://mozilla-hub.atlassian.net/browse/SYNC-4034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SYNC-3974]: https://mozilla-hub.atlassian.net/browse/SYNC-3974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,7 @@ POETRY := poetry --directory $(TESTS_DIR)
 DOCKER_COMPOSE := docker compose
 PYPROJECT_TOML := $(TESTS_DIR)/pyproject.toml
 FLAKE8_CONFIG := $(TESTS_DIR)/.flake8
-STAGE_SERVER_URL := "wss://autoconnect.stage.mozaws.net"
-STAGE_ENDPOINT_URL := "https://updates-autopush.stage.mozaws.net"
+LOCUST_HOST := "wss://autoconnect.stage.mozaws.net"
 
 .PHONY: ddb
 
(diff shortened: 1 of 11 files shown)


pjenvey (commented):
The issue I have with this change is that locust's special `host` argument is frustrating, disallowing websocket urls so we can't use it for that.

We currently don't even use the `endpoint_url` argument here as we use the one supplied by autoconnect when registering for a subscription. However it could be handy to support overriding the base url of those endpoints. This was something our older load tester supported: we needed it for testing an alternative endpoint server we had prototyped but it's also came in handy in other rare situations.

I played around with using a [`host` attribute](https://docs.locust.io/en/stable/writing-a-locustfile.html?highlight=host#host-attribute) with a default value so we could make it act as an optional `endpoint_url` override but this doesn't work: setting it to a value in one run then an empty value in a subsequent run didn't work correctly (locust seems to be overriding it on the class itself and only doing it when `environment.host` is not None).

Digging into this a little further I noticed only `FastHttpUser` disallows websocket urls for probably not a good reason, so I'm hoping upstream agrees and we can [get that problem fixed](https://github.com/locustio/locust/pull/2493).

Trinaa:
🌟 KUDOS @pjenvey this is a great improvement!

Next Steps: 
1. Wait for the next Locust release (expected early next week)
2. Change this PR to include the removal of AUTOPUSH_WEBSOCKER_URL in favour of LOCUST_HOST

Changing the status of this PR to draft until then.

pjenvey (approved):

PR #530, opened 2023-11-30
feat: Report BigTableErrors with extra data

jrconlin:
Closes SYNC-3707

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/error.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/error.rs
@@ -1,26 +1,76 @@
+use backtrace::Backtrace;
 use thiserror::Error;
+
+use crate::errors::ReportableError;
 
 #[derive(Debug, Error)]
 pub enum BigTableError {
-    #[error("Invalid Row Response: {0}")]
+    #[error("Invalid Row Response")]
     InvalidRowResponse(String),
 
-    #[error("Invalid Chunk: {0}")]
+    #[error("Invalid Chunk")]
     InvalidChunk(String),
 
-    #[error("BigTable read error {0}")]
+    #[error("BigTable read error")]
     Read(String),
 
-    #[error("BigTable write error {0}")]
+    #[error("BigTable write error")]
     Write(String),
 
-    #[error("BigTable Admin Error {0}")]
-    Admin(String),
+    #[error("BigTable Admin Error")]
+    Admin(String, Option<String>),
 
     #[error("Bigtable Recycle request")]
     Recycle,
 
     /// General Pool builder errors.
-    #[error("Pool Error {0}")]
+    #[error("Pool Error")]
     Pool(String),
+}
+
+impl ReportableError for BigTableError {
+    fn reportable_source(&self) -> Option<&(dyn ReportableError + 'static)> {
+        None
+    }
+
+    fn backtrace(&self) -> Option<&Backtrace> {
+        None
+    }
+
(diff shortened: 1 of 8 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -295,12 +295,24 @@ impl BigTableClientImpl {
             .drop_row_range_async(&req)
             .map_err(|e| {
                 error!("{:?}", e);
-                error::BigTableError::Admin("Could not delete data from table (0)".to_owned())
+                error::BigTableError::Admin(
+                    format!(
+                        "Could send delete command for {}",
```
pjenvey, line 300:
```suggestion
                        "Could not send delete command for {}",
```

pjenvey (changes requested):

autopush-common/src/db/bigtable/bigtable_client/error.rs
```diff
@@ -1,26 +1,76 @@
...
     Read(String),
 
-    #[error("BigTable write error {0}")]
+    #[error("BigTable write error")]
     Write(String),
 
-    #[error("BigTable Admin Error {0}")]
-    Admin(String),
+    #[error("BigTable Admin Error")]
+    Admin(String, Option<String>),
```
pjenvey, line 21: Does it make sense for this to be an `Option<grpcio::Error>` instead of String (I guess this applies to the other variants as well)?

jrconlin: I was debating that. 

Using the grpcio::Error might give a bit more info, but most of the errors aren't really useful beyond the string. 

Meh, I'll convert them.

jrconlin: I'll also note that we basically convert these to strings in the only place that they're consumed, when we add them to `extras`, but it's good hygiene anyway.

jrconlin:
@pjenvey No idea why I can't re-request or dismiss your review for this PR, but it's ready anyway.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -295,12 +295,24 @@ impl BigTableClientImpl {
             .drop_row_range_async(&req)
             .map_err(|e| {
                 error!("{:?}", e);
-                error::BigTableError::Admin("Could not delete data from table (0)".to_owned())
+                error::BigTableError::Admin(
+                    format!(
+                        "Could not send delete command for {}",
+                        self.settings.table_name.clone()
```
pjenvey, line 301:
```suggestion
                        self.settings.table_name
```

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -295,12 +295,24 @@ impl BigTableClientImpl {
...
+                )
             })?
             .await
             .map_err(|e| {
                 error!("post await: {:?}", e);
-                error::BigTableError::Admin("Could not delete data from table (1)".to_owned())
+                error::BigTableError::Admin(
+                    format!(
+                        "Could not delete data from table {}",
+                        self.settings.table_name.clone()
```
pjenvey, line 312:
```suggestion
                        self.settings.table_name
```

pjenvey (changes requested):

pjenvey (approved):

PR #532, opened 2023-12-01
feat: Add integration test for mobile HTTP calls

jrconlin:
This adds a "stub" router, specifically for integration testing. (May
put this behind a feature flag later.) Depending on the "client" type
the stub router will either succeed or fail.

This also adds some tests to ensure that mobile devices get a proper
endpoint depending on presence of a `key` value.

Closes SYNC-4041


autoendpoint/src/routers/stub/router.rs
```diff
@@ -0,0 +1,88 @@
...
+            .router_data
+            .as_ref()
+            .ok_or(StubError::General("No Registration".to_owned()))?;
+
+        let default = Value::String("success".to_owned());
+        let client_type = router_data.get("type").unwrap_or(&default);
+        let client = self
+            .clients
+            .get(&client_type.to_string().to_lowercase())
+            .unwrap_or_else(|| self.clients.get("success").unwrap());
```
pjenvey, line 81: Maybe default to erroring instead?
```suggestion
            .unwrap_or_else(|| self.clients.get("error").unwrap());
```

autoendpoint/src/routers/stub/settings.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[serde(rename = "credentials")]
+    pub server_credentials: String,
+}
+
+/// Credential information for each application
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+pub struct StubServerSettings {
+    pub project_id: String,
+    pub error: String,
+}
```
pjenvey, line 16: The creds aren't used at all, kill?
```suggestion
}

```

pjenvey (changes requested):
I think this needs the option of disabling it completely on production, otherwise we could be allowing "stub" `router_type`s to exist, particularly in the db, which isn't ideal.

This is a little annoying to accomplish due to routers pretty much being "static" along with their `RouterType` enum but I think I accomplished it on [this branch](https://github.com/mozilla-services/autopush-rs/compare/feat/sync-4041-http...feat/sync-4041-http-disable), as there only seems to be the one extractor that actually parses/validates them.

autoendpoint/src/routers/stub/settings.rs
jrconlin: Ah! Thanks, forgot to make the change to router.

jrconlin:
I agree that this should probably have some flag that prevents it from going to production. I've added a `stub` feature flag.

autoendpoint/src/routers/stub/router.rs
```diff
@@ -0,0 +1,96 @@
...
+                StubError::General(format!("Unknown App ID: {}", app_id.to_owned())).into(),
+            );
+        }
+
+        let mut router_data = HashMap::new();
+        router_data.insert(
+            "token".to_string(),
+            serde_json::to_value(&router_data_input.token).unwrap(),
+        );
+        router_data.insert("app_id".to_string(), serde_json::to_value(app_id).unwrap());
```
pjenvey, line 65: A minor nit, but there's a mix of `to_string` and `to_owned` in this PR, let's just stick with one
```suggestion
            "token".to_owned(),
            serde_json::to_value(&router_data_input.token).unwrap(),
        );
        router_data.insert("app_id".to_owned(), serde_json::to_value(app_id).unwrap());
```

autoendpoint/src/routers/stub/settings.rs
```diff
@@ -0,0 +1,33 @@
...
+            error: "General Error".to_owned(),
+        }
+    }
+}
+
+impl Default for StubSettings {
+    fn default() -> Self {
+        Self {
+            url: "http://localhost:8080".to_owned(),
+            server_credentials: "{\"error\":\"General Error\"}".to_string(),
```
pjenvey, line 30:
```suggestion
            url: "http://localhost:8080".to_owned(),
            server_credentials: "{\"error\":\"General Error\"}".to_owned(),
```

pjenvey (changes requested):

pjenvey (dismissed):
Just needs the conflicts resolved

pjenvey (approved):

PR #533, opened 2023-12-04
docs: Document VAPID key should be base64 encoded

tarikeshaq:
Realized there was no documentation on the format of the VAPID key, read the source code and realized we call https://github.com/mozilla-services/autopush-rs/blob/6624a19ad97af90174cc7623b9b8683392c3aec9/autopush-common/src/util/mod.rs#L30-L32

(for reference, here's where we initiate the decoding while [making the endpoint](https://github.com/mozilla-services/autopush-rs/blob/6624a19ad97af90174cc7623b9b8683392c3aec9/autopush-common/src/endpoint.rs#L26)

cc @jrconlin

### DIFF
--- a/docs/src/http.md
+++ b/docs/src/http.md
@@ -250,7 +250,7 @@ An example of the Authorization header would be:
     Authorization: Bearer 00secret00
 ```
 **{vapidKey}**  
-_The [VAPID Key](https://datatracker.ietf.org/doc/html/draft-ietf-webpush-vapid-01) provided by the subscribing third party_
+_The [VAPID Key](https://datatracker.ietf.org/doc/html/rfc8292#section-3.2) provided by the subscribing third party_
 
 The VAPID key is optional and provides a way for an application server to voluntarily identify itself.
 When the VAPID key is provided, autopush will return an endpoint that can only be used by the application server that provided the key.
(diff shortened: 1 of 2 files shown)


jonalmeida (approved):
Nice!

As mentioned offline, if we want to add some simple checks to avoid common mistakes, we could do something similar to [Gecko's WebPushSubscription construction](https://searchfox.org/mozilla-central/rev/78a2c17cc80680a5a82446e4ce7c45a73b935383/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/WebPushSubscription.java#96-114).

docs/src/http.md
```diff
@@ -249,6 +249,14 @@ An example of the Authorization header would be:
 ```html
     Authorization: Bearer 00secret00
 ```
+**{vapidKey}**  
+_The [VAPID Key](https://datatracker.ietf.org/doc/html/draft-ietf-webpush-vapid-01) provided by the subscribing third party_
```
jrconlin, line 253: It's a real RFC, no need to refer to the draft (particularly a very old draft).

```suggestion
_The [VAPID Key](https://datatracker.ietf.org/doc/html/rfc8292#section-3.2) provided by the subscribing third party_
```

jrconlin (changes requested):
Really, just the one change to use the proper RFC.

jrconlin (approved):

tarikeshaq:
Hmm I've seen this sometimes in the past, lemme try closing and re-opening the PR

jrconlin:
Rather than fight CI, https://github.com/mozilla-services/autopush-rs/pull/535

PR #535, opened 2023-12-05
docs: Document VAPID key should be base64 encoded

jrconlin:
Repost of https://github.com/mozilla-services/autopush-rs/pull/533


tarikeshaq (approved):
Thank you for taking care of it 🙏

PR #539, opened 2023-12-15
bug: copy over the channels when migrating the user

jrconlin:
This will attempt to copy the existing channels for a user. This is important because desktop does not check and rebuild channels if the server doesn't have them. (Might want to make that a standardized feature in the future.)

Closes: SYNC-4050

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -700,11 +700,45 @@ impl DbClient for BigTableClientImpl {
             .conn
             .mutate_rows(&req)
             .map_err(error::BigTableError::Write)?;
-        // TODO: is there a better way to do this? I don't think we care about the return stream.
-        // We just want to know if it worked or not.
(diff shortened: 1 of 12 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -650,6 +651,64 @@ impl DbClient for BigTableClientImpl {
...
+        req.set_entries(entries);
+
+        let bigtable = self.pool.get().await?;
+
+        // ClientSStreamReceiver will cancel an operation if it's dropped before it's done.
+        let resp = bigtable
+            .conn
+            .mutate_rows(&req)
+            .map_err(error::BigTableError::Write)?;
+        // TODO: is there a better way to do this? I don't think we care about the return stream.
```
pjenvey, line 703: A problem is that it's a stream of results of each individual mutation. It appears some in the batch can succeed while others fail. I think at the very least we need to loop through the stream to check every result: this currently only checks the first result.

pjenvey (approved):

PR #542, opened 2023-12-21
fix update w/ a conditional filter

pjenvey:
This doesn't need an actual PR I just wanted to make sure you didn't miss it


jrconlin:
Sigh, glad to see that this was something stupid.

jrconlin:
Heh, I was trying to see if we could combine multiple chains, but yeah, that didn't work.

Thanks! I'll incorporate all this into the branch I was working on.

pjenvey:
The chain reduces the number of rows/cells for each filter entry in the chain, so once it got to that second set, all it has left to filter on is a 'router_type:webpush' cell.

I didn't actually do this this but I imagine it can be helpful to use the filter on a read to see what it produces, potentially at each step, before putting it on a write

Also they discourage conditionals because they can increase latency but hopefully this one is simple enough?
https://cloud.google.com/bigtable/docs/filters#performance
_The [conditions](https://cloud.google.com/bigtable/docs/using-filters#condition) filter, in particular, can increase latency, because conditions are much slower than other filters. If your read request is extremely performance-sensitive, do not use conditions in the request._

https://cloud.google.com/bigtable/docs/using-filters#condition
_Caution: The condition filter is much slower than other filters, especially when the condition filter specifies a false filter. We recommend that you not use condition filters._

PR #543, opened 2024-01-05
feat: Add `must_use` to `update_user` functions

jrconlin:
Closes SYNC-3983


pjenvey (commented):
The plain #[must_use] unfortunately doesn't do it for wrapper types: https://stackoverflow.com/questions/71873348/how-to-make-the-inner-value-of-a-result-must-use

I mentioned in the bug that this would require the use of something like a `MustUse<T>` wrapper type to enforce the usage of the actual bool. Admittedly this is a bit annoying but this issue isn't high priority either.

jrconlin:
I'll convert this to a draft so that it's no longer a priority for review. 
(Sorry, thought it was a quick win)

gruberb:
I think we can close this @jrconlin ?

PR #547, opened 2024-01-10
feat: run unit tests against the bigtable emulator

pjenvey:
also make BigTableError more verbose and use the try! operator in the run_gauntlet test

Closes: SYNC-4059

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/error.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/error.rs
@@ -1,3 +1,5 @@
+use std::fmt::{self, Display};
+
 use backtrace::Backtrace;
 use thiserror::Error;
 
@@ -55,9 +57,9 @@ impl From<i32> for MutateRowStatus {
     }
 }
 
-impl ToString for MutateRowStatus {
-    fn to_string(&self) -> String {
-        match self {
+impl Display for MutateRowStatus {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
             MutateRowStatus::OK => "Ok",
             MutateRowStatus::Cancelled => "Cancelled",
             MutateRowStatus::Unknown => "Unknown",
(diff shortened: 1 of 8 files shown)


jrconlin (approved):

tests/integration/test_integration_all_rust.py
```diff
@@ -120,7 +121,7 @@ def get_db_settings() -> str | dict[str, str | int | float] | None:
             router_table=ROUTER_TABLE,
             message_table=MESSAGE_TABLE,
             current_message_month=MESSAGE_TABLE,
-            table_name="projects/test/instances/test/tables/autopush",
+            table_name="projects/test/instances/localhost:8086/tables/autopush",
```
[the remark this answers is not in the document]
pjenvey: Switched everything back to project "test" which apparently now works for me, oops.

scripts/setup_bt.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -eu
```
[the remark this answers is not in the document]
pjenvey: 👍 I had already made similar defaults for the other fields for a separate PR, brought that all here

autopush-common/src/db/bigtable/bigtable_client/error.rs
```diff
@@ -75,41 +77,40 @@ impl ToString for MutateRowStatus {
...
-        .to_owned()
+        })
     }
 }
 
 #[derive(Debug, Error)]
 pub enum BigTableError {
-    #[error("Invalid Row Response")]
-    InvalidRowResponse(grpcio::Error),
+    #[error("Invalid Row Response: {0}")]
```
jrconlin, line 86: Won't these produce errors that are harder to collect in Sentry? 
The argument is being displayed in the format and as part of the event, so it's not getting lost.

pjenvey: I don't think this really complicates Sentry's collection -- we do [the same in syncstorage](https://github.com/mozilla-services/syncstorage-rs/blob/cc8d5847dc37de321a484321b222e0757bc322ab/syncstorage-spanner/src/error.rs#L66)

Outside of Sentry this is helpful, e.g. when diagnosing a db error locally which you typically aren't using Sentry with.

jrconlin: Fair.

jrconlin (approved):

PR #548, opened 2024-01-11
bug: filter by timestamp

jrconlin:
This includes a filter by cell timestamp. This also includes a lot of inline comments pointing out potential hazards and pitfalls that I experienced working on this.

NOTE: There is a (hopefully remote) possiblity that a record could be returned with a specific cell missing if that cell's timestamp has expired. Timestamps are per cell, not per row.

Closes: SYNC-4060

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -291,7 +291,6 @@ impl BigTableClientImpl {
         let mut req = DropRowRangeRequest::new();
         req.set_name(self.settings.table_name.clone());
         req.set_row_key_prefix(row_key.as_bytes().to_vec());
-        req.set_delete_all_data_from_table(true);
         admin
             .drop_row_range_async(&req)
             .map_err(|e| {
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -70,6 +70,47 @@ pub struct BigTableClientImpl {
     pool: BigTablePool,
 }
 
+fn timestamp_filter() -> Result<data::RowFilter, error::BigTableError> {
+    let mut timestamp_filter = data::RowFilter::default();
+    let bt_now: i64 = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .map_err(error::BigTableError::WriteTime)?
+        .as_millis() as i64;
+    let mut range_filter = data::TimestampRange::default();
+    range_filter.set_start_timestamp_micros(bt_now * 1000);
+    timestamp_filter.set_timestamp_range_filter(range_filter);
+
+    Ok(timestamp_filter)
+}
+
+/*
+/// Filter based on the calculated expiry of the data row.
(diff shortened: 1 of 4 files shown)


jrconlin:
I was concerned that there might be a (however slight) chance that during a read either `data` or `headers` could be lost due to per cell TTL, but I'm going to presume that the read operation is atomic at that level and all data associated with the message would be returned. This is mostly because there are two types of messages, `aesgcm` which requires headers and `aes128gcm` which does not, and adding the code to validate the pulled data would add significant complexity.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1167,15 +1186,28 @@ impl DbClient for BigTableClientImpl {
...
                 format!("^{}#[^#]+#01:.+", uaid.simple())
                     .as_bytes()
                     .to_vec(),
             );
-            regex_filter
+
+            let time_filter = timestamp_filter()?;
+            filter_chain.push(time_filter);
+
+            filter_chain.push(row_filter);
```
pjenvey, line 1206: We definitely want the timestamp filter to be the last in the chain (like you have in the other method): https://cloud.google.com/bigtable/docs/using-filters#limiting-filters

```suggestion
            filter_chain.push(row_filter);

            filter_chain.push(time_filter);
```

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1219,15 +1257,34 @@ impl DbClient for BigTableClientImpl {
...
                 uaid.simple(),
                 STANDARD_NOTIFICATION_PREFIX,
             );
             trace!("🉑 regex filter {:?}", pattern);
-            filter.set_row_key_regex_filter(pattern.as_bytes().to_vec());
+            row_filter.set_row_key_regex_filter(pattern.as_bytes().to_vec());
+            filter_chain.push(row_filter);
+
+            // Filter on our TTL.
+            // let time_filter = expiry_filter()?;
```
pjenvey, line 1272:
```suggestion
```

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -73,6 +73,47 @@ pub struct BigTableClientImpl {
...
+    timestamp_filter.set_timestamp_range_filter(range_filter);
+
+    Ok(timestamp_filter)
+}
+
+/*
+/// Filter based on the calculated expiry of the data row.
+/// NOTE: Filtering on a specific cell value will return only that
+/// cell value in the row.
+fn expiry_filter() -> Result<data::RowFilter, error::BigTableError> {
```
pjenvey, line 93: Do we want to keep this?

pjenvey (changes requested):

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: I was going to make an argument that we should keep the comment in lieu of a document that includes the various "Bigtable Learnings" we've had to make (because Bigtable is weird), but then I realized I should just create that document.

pjenvey (approved):

PR #557, opened 2024-01-18
feat: Be more clear and consistent about features around `dual`

jrconlin:
This patch makes the feature sets consistent between the major crates.

Closes #SYNC-4075


pjenvey:
But AFAICT dual isn't currently able to be configured with autoconnect:

https://github.com/mozilla-services/autopush-rs/blob/8f6e03e/autoconnect/autoconnect-settings/src/app_state.rs#L80

vs autoendpoint

https://github.com/mozilla-services/autopush-rs/blob/6624a19/autoendpoint/src/server.rs#L82

pjenvey (commented):
Want to take a crack at enabling integration tests against dual in CI?

pjenvey (approved):

Issue #366, opened 2023-04-19
Improve Uaid usage (consider a different type)


data-sync-user:
UAIDs are represented by `Uuid` as are channel_ids. This makes a lot sense except UAID’s representation in the db has historically been of the Uuid “simple” format of all lowercase no dashes (the Python `Uuid.hex` property returns this format).

Whereas channel_id is stored in the default “hyphenated” style, returned by `Uuid::to_string()` (and Python’s `__str__`).

This can become a source of subtle bugs: it’s easy to forget which type of Uuid renders which way.

I think we can avoid this by creating a separate Uaid type wrapping Uuid and rendering the “simple” format via its `to_string`. Despite using “simple” in the db, we have historically rendered the "hyphenated" style of the Uaid to the client. So this new type's ser/deser would still render hyphenated as the `Uuid` type does.

Alongside this channel_id could have its own type but it would be a type alias of Uuid.

Then we can call `to_string` on both types in the db code without worrying about which is formatted which way. Serializing them to the client would continue working as is (rendering “hyphenated”).

A prototype: [https://github.com/mozilla-services/autopush-rs/compare/b4298eab64a354...feat/uaid-simple](https://github.com/mozilla-services/autopush-rs/compare/b4298eab64a354...feat/uaid-simple|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3690)

jrconlin:
I believe that this is a dup of a later issue. 

What we should do is consider the UAID and CHID as discrete types in Autopush and treat them separately (e.g. UAID can be collapsed down, but the CHID is an opaque string provided by the UA.)

PR #561, opened 2024-01-19
fix: channel_ids should continue being represented in hyphenated format

pjenvey:
as they already are in the channel_id column and in chidmessageid

Closes: SYNC-4076


jrconlin (approved):
those damn channelIDs are going to be the death of me.

jrconlin:
see also https://github.com/mozilla-services/autopush-rs/issues/562

pjenvey:
https://github.com/mozilla-services/autopush-rs/issues/366 also related

PR #569, opened 2024-01-24
fix: errors when deleting and provisioning load test clusters

Trinaa:
Replaced the load test configuration to use c3d-standard-4 instead of n1-standard-2 machine types increasing the user capacity per node 3-fold

Issue: [SYNC-4082](https://mozilla-hub.atlassian.net/browse/SYNC-4082)

[SYNC-4082]: https://mozilla-hub.atlassian.net/browse/SYNC-4082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/load/README.md
+++ b/tests/load/README.md
@@ -234,7 +234,7 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
 [13]: https://docs.locust.io/en/stable/writing-a-locustfile.html#wait-time
 [14]: https://docs.locust.io/en/stable/writing-a-locustfile.html#task-decorator
 [15]: https://console.cloud.google.com/home/dashboard?q=search&referrer=search&project=spheric-keel-331521&cloudshell=false
-[16]: https://console.cloud.google.com/compute/instances?project=spheric-keel-331521
+[16]: https://console.cloud.google.com/gcr/images/spheric-keel-331521/global/locust-autopush?project=spheric-keel-331521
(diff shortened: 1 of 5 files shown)


Trinaa:
@dlactin

With the C3 machine we are able to host 3x users per node, reducing the cluster size significantly. I no longer have errors when spinning the cluster up and down. There are a lot of resources to spare on these nodes, we can probably reduce the cluster by another 2x or 3x if we can reserve a CPU per Locust process.

See [Test Results](https://docs.google.com/spreadsheets/d/1-nF8zR98IiLZlwP0ISwOB8cWRs74tJGM1PpgDSWdayE/edit#gid=0)

Note: There is an issue with total requests being sent atm, which may or may not be related to the use of the C3 machines. @jrconlin is helping me investigate.

Trinaa:
> Note: There is an issue with total requests being sent at the moment, which may or may not be related to the use of the C3 machines.

On 2024/01/25 we confirmed that the issue with total request limits is isolated to test runs with the _c3d-standard-4_ nodes only. Furthermore, we noticed that the total request count is caped at ~ 1250000, from running the load tests with drastically different users/worker counts.

On 2024/01/29 we ran an experiment, isolating an increase in node count, which seems to confirm we are experiencing IP rate limiting.

PR #573, opened 2024-01-24
bug: Clarify how the index for Row is used between reading and writing to Bigtable

jrconlin:
There is some strong confusion about the row collection in Bigtable due
to a mis-match of the types being used. This PR corrects that so the
`bigtable_client::row::Row.cells` are defined as
`HashMap<Qualifier,Vec<Cell>>`. This means that when adding cells,
you specify it as
```rust
row.cells.insert(
        "updated".to_owned(),
        vec![cell::Cell{
                family: ROUTER_FAMILY.to_owned(),
                value: now.to_be_bytes().to_vec(),
                ..Default::default()
        }],
);
```

`get_mutations(..)` will use the specified qualifier to populate the
cells when generating the list of changes.

Closes: [SYNC-4093](https://mozilla-hub.atlassian.net/browse/SYNC-4093)

[SYNC-4093]: https://mozilla-hub.atlassian.net/browse/SYNC-4093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autopush-common/Cargo.toml
+++ b/autopush-common/Cargo.toml
@@ -80,6 +80,7 @@ actix-rt = "2.8"
 default = ["dynamodb"]
 bigtable = ["dep:google-cloud-rust-raw", "dep:grpcio", "dep:protobuf"]
 dynamodb = ["dep:rusoto_core", "dep:rusoto_credential", "dep:rusoto_dynamodb"]
+dual = ["dynamodb", "bigtable"]
 emulator = [
     "bigtable",
 ] # used for testing big table, requires an external bigtable emulator running.
--- a/autopush-common/Cargo.toml
+++ b/autopush-common/Cargo.toml
@@ -76,8 +76,8 @@ actix-rt = "2.8"
 
 [features]
 # for testing:
-# default = ["emulator", "bigtable", "dynamodb"]
-default = ["dynamodb"]
+default = ["emulator", "bigtable", "dynamodb"]
+# default = ["dynamodb"]
 bigtable = ["dep:google-cloud-rust-raw", "dep:grpcio", "dep:protobuf"]
 dynamodb = ["dep:rusoto_core", "dep:rusoto_credential", "dep:rusoto_dynamodb"]
 dual = ["dynamodb", "bigtable"]
--- a/autopush-common/Cargo.toml
+++ b/autopush-common/Cargo.toml
@@ -75,8 +75,8 @@ tokio = { version = "0.2", features = ["macros"] }
 actix-rt = "2.8"
 
 [features]
-# default = ["dual"]
-default = ["dual", "emulator"]
+# NOTE: Do not set a `default` here, rather specify them in the calling library.
+# This is to reduce complexity around feature specification.
 bigtable = ["dep:google-cloud-rust-raw", "dep:grpcio", "dep:protobuf"]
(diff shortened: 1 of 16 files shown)


pjenvey (commented):
IIRC the `last_family_cells` and probably `cells` and in merge `PartialRow` also need the same change?

pjenvey:
This breaks `user_to_row` (failing test) and looks like `save_message` too

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1033,9 +1022,8 @@ impl DbClient for BigTableClientImpl {
         };
 
         row.cells.insert(
-            MESSAGE_FAMILY.to_owned(),
+            ROUTER_FAMILY.to_owned(),
```
jrconlin, line 1025: When we are [generating this value](https://github.com/mozilla-services/autopush-rs/blob/763e2afb3c87b06429868b09c0a365021e4cf2f5/autopush-common/src/db/bigtable/bigtable_client/mod.rs#L461-L462) in `user_to_row`, it's in the `router` family.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1337,6 +1325,17 @@ mod tests {
             client.get_user(&uaid).await?.unwrap().connected_at
         );
 
+        // can we increment the storage for the user?
+        client
```
jrconlin, line 1329: Added test to verify we can increment storage (generally checks to see that we're not leaking a qualifier as the family id)

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -434,13 +438,11 @@ impl BigTableClientImpl {
 
         let mut cells: Vec<cell::Cell> = vec![
             cell::Cell {
-                family: ROUTER_FAMILY.to_owned(),
```
jrconlin, line 437: These are removed because we inherit the family from the index identifier.

autopush-common/Cargo.toml
```diff
@@ -76,8 +76,8 @@ actix-rt = "2.8"
 
 [features]
 # for testing:
-# default = ["emulator", "bigtable", "dynamodb"]
-default = ["dynamodb"]
+default = ["emulator", "bigtable", "dynamodb"]
```
pjenvey, line 79: Do we want this change included? It's probably time to switch these on by default but this looks like a local edit (I would imagine switching it on by default would touch more `Cargo.toml`s?)

jrconlin: well, we should probably remove `emulator` from the list.

pjenvey (dismissed):

pjenvey:
Oddly enough I guess the default feature changes now causes clippy to warn about a couple small things in the bigtable_client module, even though we always run clippy with `--all-features` 🤨

pjenvey (approved):
I didn't notice before but the clippy failures were only in the tests -- but we also run with`--all-targets` which should have caught it alongside `--all-features`? 🤨 🤨

jrconlin:
Yeah, apparently when I ran clippy, i missed the "--all-features".

PR #575, opened 2024-01-25
chore: stop building legacy autopush

pjenvey:
as its pending removal

Issue: SYNC-4036

I don't want to finish the rest of SYNC-4036 at this moment but this first step will save build time (mostly for local builds)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -326,14 +326,14 @@ workflows:
           filters:
             tags:
               only: /.*/
-      - build:
-          name: build-autopush
-          image: autopush:build
-          crate: autopush
-          binary: autopush_rs
-          filters:
-            tags:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -326,14 +326,14 @@ workflows:
           filters:
             tags:
               only: /.*/
-      - build:
```
jrconlin, line 329: I believe you'd yell at me that these should be deleted, not commented.

jrconlin (approved):

.circleci/config.yml
pjenvey: Didn't bother since we're probably killing the rest of this shortly but fair enough

jrconlin (approved):

PR #582, opened 2024-01-29
feat: Add integration test for `dual` mode

jrconlin:
This adds an integration test for "Dual" mode. Note that this uses a database settings file `dual_test.json` to contain the doubly escaped values. The values in this file should match those used in `test_integration_all_rust.py` for data store configuration and set-up.

This also adds a bit more `debug` messages to indicate successful operations. This was useful while determining a few problems with the configuration, but are not strongly required. I can remove them if requested.

Closes: https://mozilla-hub.atlassian.net/browse/SYNC-4111

### DIFF
--- a/tests/integration/test_integration_all_rust.py
+++ b/tests/integration/test_integration_all_rust.py
@@ -69,8 +69,8 @@ ROUTER_PORT = 9170
 MP_CONNECTION_PORT = 9052
 MP_ROUTER_PORT = 9072
 
-CONNECTION_BINARY = os.environ.get("CONNECTION_BINARY", "autopush_rs")
-CONNECTION_SETTINGS_PREFIX = os.environ.get("CONNECTION_SETTINGS_PREFIX", "autopush__")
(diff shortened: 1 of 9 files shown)


tests/integration/test_integration_all_rust.py
```diff
@@ -69,8 +69,8 @@
 MP_CONNECTION_PORT = 9052
 MP_ROUTER_PORT = 9072
 
-CONNECTION_BINARY = os.environ.get("CONNECTION_BINARY", "autopush_rs")
-CONNECTION_SETTINGS_PREFIX = os.environ.get("CONNECTION_SETTINGS_PREFIX", "autopush__")
+CONNECTION_BINARY = os.environ.get("CONNECTION_BINARY", "autoconnect")
```
jrconlin, line 72: since we no longer use `autopush_rs`, switching to `autoconnect` for this.

pjenvey: Can you kill the `CONNECTION_BINARY/SETTINGS` override in the Makefile's integration-test target along with this?

pjenvey (approved):

PR #583, opened 2024-01-29
refactor: cleanup add_channels to use mutate_row

pjenvey:
now that it only writes to a single row it doesn't need mutate_rows

begin writing expiry to router timestamps (though not currently used by GC)

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -24,7 +24,7 @@ use uuid::Uuid;
 use crate::db::{
     client::{DbClient, FetchMessageResponse},
     error::{DbError, DbResult},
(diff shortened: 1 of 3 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1366,17 +1333,21 @@ mod tests {
...
         assert!(msgs.is_empty());
 
+        assert!(client
+            .remove_node_id(&uaid, &node_id, connected_at)
+            .await
+            .is_ok());
+        // TODO:
+        // did we remove it?
+        //let fetched = client.get_user(&uaid).await?.unwrap();
+        //assert_eq!(fetched.node_id, None);
```
pjenvey, line 1349: Note this test is also currently failing on master (which doesn't set MAX_ROUTER_TTL timestamps) -- I'm going to fix this in a subsequent PR.

jrconlin (approved):

PR #584, opened 2024-01-29
feat: Load test fixups

jrconlin:
Some minor tweaks to the load tests to make them easier to run locally.

These include:
* normalizing the `--config` option across binaries (this is not used in production)
* Change the load test to use a hash of the payload instead of the payload itself (freeing memory and making checks faster)
* Make the `WORKER_COUNT` an environment parameter


tests/load/setup_k8s.sh
```diff
@@ -10,7 +10,7 @@ CLUSTER='autopush-locust-load-test'
 TARGET='https://updates-autopush.stage.mozaws.net'
 SCOPE='https://www.googleapis.com/auth/cloud-platform'
 REGION='us-central1'
-WORKER_COUNT=300
```
Trinaa, line 13: The WORKER_COUNT found here has to correspond with the constant of the same name in [load.py](https://github.com/mozilla-services/autopush-rs/blob/master/tests/load/locustfiles/load.py#L47).

Trinaa (changes requested):

tests/load/setup_k8s.sh
```diff
@@ -10,7 +10,7 @@ CLUSTER='autopush-locust-load-test'
...
+WORKER_COUNT=${WORKER_COUNT:-300}
```
Trinaa, line 13: The WORKER_COUNT found here has to correspond with the constant of the same name in [load.py](https://github.com/mozilla-services/autopush-rs/blob/master/tests/load/locustfiles/load.py#L47).

I also just noticed the description on that class has an [erroneous total user count in the description](https://github.com/mozilla-services/autopush-rs/blob/master/tests/load/locustfiles/load.py#L39C8-L39C82)

Trinaa (changes requested):

jrbenny35:
I ran this locally and was getting some errors.
image

@Trinaa says that this is from a race condition that comes from using dict's as the data type for storing the records. I want to look into this a bit so I can understand what is happening.

jrconlin:
I guess I'm not running the load tests correctly, because I'm not seeing those failures:
image

Trinaa:
> I guess I'm not running the load tests correctly, because I'm not seeing those failures...

@jrconlin the list of request names looks a bit short in your screenshot. There should be some `POST notification - send` and `WSS notification - recv` in there...

jrbenny35:
> Yeah, I realized that I was running low on node handles. I ran another test now that I've kicked the count up to stupid levels. Screenshot 2024-02-01 at 16-35-57 Locust for None
> 
> The 1 failure was a database save error resulting in a 503

Can you try running against the stage server? The URL should be set by default.

tests/load/locustfiles/locustfile.py
```diff
@@ -217,7 +217,7 @@ def post_notification(self, endpoint_url: str) -> None:
         )
 
         record = NotificationRecord(send_time=time.perf_counter(), data=data)
-        self.notification_records.append(record)
+        self.notification_records[sha1(data.encode()).digest()] = record
```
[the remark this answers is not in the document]
jrconlin: `self.notification_records` was converted to a `dict` type. What I'm doing is indexing that `dict` by creating a `sha1` of the data. This should produce a unique hash which will be much faster to look up than scanning the list of records, particularly for very large sets of data. 

(It's technically possible to have a hash conflict with `sha1`, but the probability is very low in our case.)

jrconlin:
I can, the reason I was running locally was to load test `dual` mode. (I had configured autopush to run as that locally as well) I'm presuming that stage is still configured to run against the default dynamodb, which is fine, since we can get a good baseline that way. I'll work with @emaydeck-mozilla about what the dual configuration should be.

tests/load/setup_k8s.sh
jrconlin: I updated [load.py](https://github.com/mozilla-services/autopush-rs/pull/584/files#diff-7d05ac34f9a820145966c4583f25e689fbcb439fe6ec064f7cfa19c7cb9e8950R50) to also use the `WORKER_COUNT` environment variable, if it's specified.

.circleci/config.yml
```diff
@@ -164,6 +164,7 @@ jobs:
           name: Integration tests (DynamoDB)
           command: make integration-test
           environment:
+            DB_DSN: grpc://localhost:8086
```
pjenvey, line 167: As this is the DynamoDB run
```suggestion
```

pjenvey (changes requested):

Trinaa:
> The 1 failure was a database save error resulting in a 503

Interesting okay! As long as you aren't getting those "There is no record of the 'notification' message" then my race condition fears are allayed.

I'm still worried about the WORKER_COUNT environment variable a little. In the setup_k8s.sh, we set the `total-max-nodes=75` that means that we can handle a maximum worker count of 225 (75x3 locust worker pods) on the C3s before we start getting cryptic errors.

Trinaa (commented):
As long as you aren't getting those "There is no record of the 'notification' message" then my race condition fears are allayed.

I'm still worried about the WORKER_COUNT environment variable a little. In the setup_k8s.sh, we set the total-max-nodes=75 that means that we can handle a maximum worker count of 225 (75x3 locust worker pods) on the C3s before we start getting cryptic errors. I won't block on this though cause I don't have a better solution... unless you want to do some math like total-max-nodes=roundup(WORKER_COUNT/3)

Trinaa (approved):

jrbenny35:
@jrconlin Do we want to merge this still?

jrconlin:
If it's still relevant and useful. yes.

pjenvey (approved):

PR #586, opened 2024-01-30
feat: add metadata headers for bigtable

jrconlin:
Closes: [SYNC-3967](https://mozilla-hub.atlassian.net/browse/SYNC-3967)

[SYNC-3967]: https://mozilla-hub.atlassian.net/browse/SYNC-3967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pjenvey (changes requested):
This only covers `read_rows`' prefix but all the ops we use need metadata headers, with both prefixes and routing_params.

Prefix may not be required with routing params according to https://github.com/googleapis/google-cloud-cpp/issues/190 but we might as well include it as well, the values needed for both are shown here: https://github.com/googleapis/google-cloud-cpp/issues/190#issuecomment-370520185 

The routing params (`x-goog-request-params`) in that PR should correlate to those shown in https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable.proto and https://github.com/googleapis/googleapis/blob/master/google/bigtable/admin/v2/bigtable_table_admin.proto (which appears to be the case when I spot checked it):

Basically every op except the admin drop row range needs a `prefix` of the table name and also routing param named `table_name` with the same value. 

Whereas the admin droprowrange which wants a `prefix` of the instance (`/projects/../instance/..`) and a routing param of `name` with a value of the table name.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -472,8 +486,29 @@ impl BigTableClientImpl {
...
+            .collect::<Vec<&str>>()
+            .first()
+            .copied();
+        if prefix.is_none() {
+            return Err(error::BigTableError::Admin(
+                "Invalid table name specified".to_string(),
+                None,
+            ));
+        }
+        let admin_meta = MetadataBuilder::with_prefix(prefix.unwrap())
```
pjenvey, line 504: Let's move all this to the constructor

Also you can just `next()` on the iterator here:
```suggestion
        let prefix = self
            .settings
            .table_name
            .split("/tables/")
            .next();
        let Some(prefix) = prefix else {
            return Err(error::BigTableError::Admin(
                "Invalid table name specified".to_string(),
                None,
            ));
        };
        let admin_meta = MetadataBuilder::with_prefix(prefix)
```

But I think `split_once` which was added a few years ago is even better (I like that it's a bit more strict, returning `None` when there's no match instead of producing empty strings):

```suggestion
        let Some((prefix, _)) = self.settings.table_name.split_once("/tables/") else {
            return Err(error::BigTableError::Admin(
                "Invalid table name specified".to_string(),
                None,
            ));
        };
        let admin_meta = MetadataBuilder::with_prefix(prefix)
```

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -154,7 +155,11 @@ impl Manager for BigtableClientManager {
     /// `BigtableClient` is the most atomic we can go.
     async fn create(&self) -> Result<BigtableDb, DbError> {
         debug!("🏊 Create a new pool entry.");
-        let entry = BigtableDb::new(self.get_channel()?);
+        let metadata = MetadataBuilder::with_prefix(&self.settings.table_name)
+            .route_to_leader(self.settings.route_to_leader)
```
pjenvey, line 159: This is currently only used for the health check's `read_rows`, so it also needs a `table_name` routing param

pjenvey (changes requested):

pjenvey (dismissed):

pjenvey (approved):

PR #588, opened 2024-01-30
feat: Specify feature chains for sub crates to allow faster incremental builds

jrconlin:
This also introduces a `gen_test_uaid` which generates a UAID that starts with a `DEADBEEF` prefix to allow for some test value isolation and removal. This can prevent some issues with multi-threaded tests causing conflicting operations on a given UAID record. 

Closes: SYNC-4112


jrconlin:
Closed, OBE.

PR #590, opened 2024-01-31
fix: fix remove_node_id's condition check to be more reliable

pjenvey:
this adds a "version" column to the router CF to act as an optimistic locking version number which the majority of writes reset the value of.

whether version has changed since the router record was read becomes the condition for update_user/remove_node_id

Closes: SYNC-4115

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -3,7 +3,7 @@ use std::fmt;
 use std::fmt::Display;
 use std::str::FromStr;
 use std::sync::Arc;
-use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime};
 
 use async_trait::async_trait;
 use cadence::StatsdClient;
@@ -12,9 +12,7 @@ use google_cloud_rust_raw::bigtable::admin::v2::bigtable_table_admin::DropRowRan
 use google_cloud_rust_raw::bigtable::admin::v2::bigtable_table_admin_grpc::BigtableTableAdminClient;
 use google_cloud_rust_raw::bigtable::v2::bigtable::ReadRowsRequest;
 use google_cloud_rust_raw::bigtable::v2::bigtable_grpc::BigtableClient;
-use google_cloud_rust_raw::bigtable::v2::data::{
-    RowFilter, RowFilter_Chain, RowFilter_Condition, ValueRange,
-};
+use google_cloud_rust_raw::bigtable::v2::data::{RowFilter, RowFilter_Chain};
 use google_cloud_rust_raw::bigtable::v2::{bigtable, data};
 use grpcio::Channel;
 use protobuf::RepeatedField;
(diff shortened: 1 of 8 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -89,6 +87,47 @@ fn timestamp_filter() -> Result<data::RowFilter, error::BigTableError> {
...
+    router_filter_chain.set_filters(filter_set);
+    let mut router_filter = RowFilter::default();
+    router_filter.set_chain(router_filter_chain);
+    router_filter
+}
+
+/// Return a newly generated `version` column `Cell`
+fn new_version_cell(timestamp: SystemTime) -> cell::Cell {
+    let value = Uuid::new_v4()
+        .as_hyphenated()
```
jrconlin, line 119: I'm curious why you want this value to be hyphenated and not simple?

jrconlin (dismissed):

jrconlin:
I like the code restructuring to use those functions. It was something I was planning on getting around to once things stablized a bit.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
pjenvey: Happy to switch to simple, I don't really care (I even attempted to store it as a more compact type but I ran into problems with the regex_filter'ing, maybe I wasn't using it right).

pjenvey: switched it to binary

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -89,6 +87,63 @@ fn timestamp_filter() -> Result<data::RowFilter, error::BigTableError> {
...
+    filter_set.push(value_filter);
+
+    router_filter_chain.set_filters(filter_set);
+    let mut router_filter = RowFilter::default();
+    router_filter.set_chain(router_filter_chain);
+    router_filter
+}
+
+/// Return a newly generated `version` column `Cell`
+fn new_version_cell(timestamp: SystemTime) -> cell::Cell {
```
jrconlin, line 138: will the timestamp ever be anything other than `std::time::SystemTime::now() + Duration::from_secs(MAX_ROUTER_TTL)`?

jrconlin (approved):

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -89,6 +87,63 @@ fn timestamp_filter() -> Result<data::RowFilter, error::BigTableError> {
     Ok(timestamp_filter)
 }
 
+/// Escape bytes for RE values
+///
+/// Based off google-re2/perl's quotemeta function
+fn escape_bytes(bytes: &[u8]) -> Vec<u8> {
```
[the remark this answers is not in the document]
pjenvey: Same -- probably the slight annoyance of deciding between serialization formats made me look around more, so I noticed the Java client's [exactMatch(bytes)](https://javadoc.io/static/com.google.cloud/google-cloud-bigtable/0.39.0-alpha/com/google/cloud/bigtable/data/v2/models/Filters.KeyFilter.html#exactMatch-com.google.protobuf.ByteString-) and its escaping code based off [re2's](https://github.com/google/re2/blob/70f66454c255080a54a8da806c52d1f618707f8a/re2/re2.cc#L467) (and least one library writing uuids as bytes to Bigtable https://github.com/spotify/magnolify/blob/4e87c93ab6aad3423179a7f0e6e7794057b0ac57/docs/mapping.md?plain=1#L35)

autopush-common/src/db/bigtable/bigtable_client/mod.rs
pjenvey: Nope but I'd rather this just 100% match with other cells in the same operation, whereas having this function make its own `SystemTime` could potentially make them differ by a ms or two.

PR #591, opened 2024-01-31
feat: Make bigtable calls retryable

jrconlin:
Closes [SYNC-4085](https://mozilla-hub.atlassian.net/browse/SYNC-4085)

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -5,8 +5,9 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
+use again::RetryPolicy;
 use async_trait::async_trait;
-use cadence::StatsdClient;
+use cadence::{CountedExt, StatsdClient};
 use futures_util::StreamExt;
 use google_cloud_rust_raw::bigtable::admin::v2::bigtable_table_admin::DropRowRangeRequest;
 use google_cloud_rust_raw::bigtable::admin::v2::bigtable_table_admin_grpc::BigtableTableAdminClient;
@@ -71,7 +72,7 @@ impl From<Uaid> for String {
 pub struct BigTableClientImpl {
     pub(crate) settings: BigTableDbSettings,
     /// Metrics client
-    _metrics: Arc<StatsdClient>,
+    metrics: Arc<StatsdClient>,
     /// Connection Channel (used for alternate calls)
     pool: BigTablePool,
 }
@@ -117,6 +118,55 @@ fn to_string(value: Vec<u8>, name: &str) -> Result<String, DbError> {
     })
 }
 
+pub fn retry_policy(max: usize) -> RetryPolicy {
+    RetryPolicy::default()
+        .with_max_retries(max)
+        .with_jitter(true)
+}
+
+pub fn retryable_describe_table_error(
+    metrics: Arc<StatsdClient>,
+) -> impl Fn(&grpcio::Error) -> bool {
+    move |err| {
+        debug!("🉑 Checking error...{err}");
+        match err {
+            grpcio::Error::RpcFailure(_) => {
+                metrics
+                    .incr_with_tags("database.retry")
+                    .with_tag("error", "RpcFailure")
+                    .with_tag("type", "bigtable")
+                    .send();
+                true
+            }
+            grpcio::Error::QueueShutdown => {
+                metrics
+                    .incr_with_tags("database.retry")
+                    .with_tag("error", "QueueShutdown")
+                    .with_tag("type", "bigtable")
+                    .send();
+                true
+            }
+            grpcio::Error::BindFail(_) => {
+                metrics
+                    .incr_with_tags("database.retry")
+                    .with_tag("error", "BindFail")
+                    .with_tag("type", "bigtable")
+                    .send();
+                true
+            }
+            grpcio::Error::CallFailure(_) => {
+                metrics
+                    .incr_with_tags("database.retry")
+                    .with_tag("error", "CallFailure")
+                    .with_tag("type", "bigtable")
+                    .send();
+                true
+            }
+            _ => false,
+        }
+    }
+}
+
 /// Connect to a BigTable storage model.
 ///
 /// BigTable is available via the Google Console, and is a schema less storage system.
@@ -150,7 +200,7 @@ impl BigTableClientImpl {
         let pool = BigTablePool::new(settings, &metrics)?;
         Ok(Self {
             settings: db_settings,
-            _metrics: metrics,
+            metrics,
             pool,
         })
     }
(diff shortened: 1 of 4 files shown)


autopush-common/src/db/bigtable/mod.rs
```diff
@@ -69,12 +73,14 @@ pub struct BigTableDbSettings {
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
     pub database_pool_connection_ttl: Option<Duration>,
     /// Max idle time(in seconds) for a connection
-    #[serde(default)]
```
pjenvey, line 72: Probably removed mistakenly? (it's causing the test failures)

jrconlin: yeah, possibly due to the merge. (Sigh)

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -193,6 +194,85 @@ fn to_string(value: Vec<u8>, name: &str) -> Result<String, DbError> {
...
+            "error occurred when fetching oauth2 token" == status.message().to_ascii_lowercase()
+        }
+        RpcStatusCode::INTERNAL => [
+            "rst_stream",
+            "rst stream",
+            "received unexpected eos on data from from server",
+        ]
+        .contains(&status.message().to_lowercase().as_str()),
+        RpcStatusCode::UNAVAILABLE => true,
+        RpcStatusCode::DEADLINE_EXCEEDED => true,
```
pjenvey, line 215:
```suggestion
        RpcStatusCode::UNAVAILABLE | RpcStatusCode::DEADLINE_EXCEEDED => true,
```

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -193,6 +194,85 @@ fn to_string(value: Vec<u8>, name: &str) -> Result<String, DbError> {
...
+        _ => false,
+    }
+}
+
+pub fn retryable_describe_table_error(
```
pjenvey, line 220:
```suggestion
pub fn retryable_error(
```

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -193,6 +194,85 @@ fn to_string(value: Vec<u8>, name: &str) -> Result<String, DbError> {
...
+            // The parameter here is a [grpcio_sys::grpc_call_error] enum
+            // Not all of these are retriable.
+            grpcio::Error::CallFailure(grpc_call_status) => {
+                metrics
+                    .incr_with_tags("database.retry")
+                    .with_tag("error", "CallFailure")
+                    .with_tag("type", "bigtable")
+                    .with_tag("code", &format!("{:?}", grpc_call_status))
+                    .send();
+                grpc_call_status == &grpcio_sys::grpc_call_error::GRPC_CALL_ERROR
```
pjenvey, line 261: Where did you copy these retry checks from, I thought all I saw in the Python client was basically the `RpcFailure` block above (I'm not sure what `QueueShutdown` or `ShutdownFailed` are but their description makes them sound like they shouldn't be retried).

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -268,10 +348,15 @@ impl BigTableClientImpl {
         req: bigtable::MutateRowRequest,
     ) -> Result<(), error::BigTableError> {
         let bigtable = self.pool.get().await?;
-        bigtable
-            .conn
-            .mutate_row_async_opt(&req, call_opts(self.metadata.clone()))
-            .map_err(error::BigTableError::Write)?
+        retry_policy(self.settings.retry_count)
```
pjenvey, line 351: Should make this a method

```suggestion
        self.retry_policy()
```

autopush-common/src/db/bigtable/mod.rs
```diff
@@ -33,6 +33,10 @@ use crate::db::bigtable::bigtable_client::MetadataBuilder;
 use crate::db::error::DbError;
 use crate::util::deserialize_opt_u32_to_duration;
 
+fn retry_default() -> usize {
+    10
```
pjenvey, line 37: I'm inclined to stick with the default of 5 for now to match the dynamo impl

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -193,6 +194,85 @@ fn to_string(value: Vec<u8>, name: &str) -> Result<String, DbError> {
...
+
+pub fn retryable_describe_table_error(
+    metrics: Arc<StatsdClient>,
+) -> impl Fn(&grpcio::Error) -> bool {
+    move |err| {
+        debug!("🉑 Checking error...{err}");
+        match err {
+            grpcio::Error::RpcFailure(status) => {
+                info!("GRPC Failure :{:?}", status);
+                metrics
```
pjenvey, line 228: Let's make the metric calls a method (or something along those lines) to avoid all the duplication in this fn.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: A combination of the python and java code. I'm fine not calling those, so I'll remove them. 

Both were a bit vague, partly because if the shutdown succeeds, then the next request goes to a different source or fails for a different reason. These appeared to be more warning messages.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: (Remembers why it's not): 
I use the same `retry_policy` for the `heath_check` which doesn't take a `BigClientTableImpl`. 
I may need to restructure things a bit to do that.

pjenvey: I noticed that, I'm fine with a one off for it. We have an open issue to rework the health check so that whole method will probably change quite a bit sooner rather than later anyway.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: Not really sure it makes that much of a difference, but sure.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: Nope, gonna push back on this one. 
The function takes one parameter, and we're using it in different paths, so making this dependent on `BigTableClientImpl` would over complicate things right now. 

We might want to revisit the structure of all of this later, but that should be a different ticket.

jrconlin: Yeah, there's a lot of "Get It Working" in this. There's lots of room for fit and finish later.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -659,16 +747,26 @@ impl BigtableDb {
...
         // Create a request that is GRPC valid, but does not point to a valid row.
         let mut req = read_row_request(table_name, "NOT FOUND");
         let mut filter = data::RowFilter::default();
         filter.set_block_all_filter(true);
         req.set_filter(filter);
 
-        let r = self
-            .conn
-            .read_rows_opt(&req, call_opts(self.metadata.clone()))
+        let r = retry_policy(10)
```
pjenvey, line 761:
```suggestion
        let r = retry_policy(5)
```

pjenvey (changes requested):

pjenvey (dismissed):

pjenvey (approved):

PR #593, opened 2024-02-01
fix: errors when deleting and provisioning load test clusters

Trinaa:
Note this is a continuation from PR: https://github.com/mozilla-services/autopush-rs/pull/569

Updated the load test configuration to:
- Use c3d-standard-4 instead of n1-standard-2 node types
- To enable the dataplane-v2 for better network performance
- To leveraging a static policy when scheduling pods on nodes, allowing us to host multiple locust processes on a single node
- To increase the node connection table capacity to prevent connection throttling

Issue: [SYNC-4082](https://mozilla-hub.atlassian.net/browse/SYNC-4082)

[SYNC-4082]: https://mozilla-hub.atlassian.net/browse/SYNC-4082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/load/kubernetes-config/locust-worker-daemonset.yml
+++ b/tests/load/kubernetes-config/locust-worker-daemonset.yml
@@ -1,0 +1,49 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: "locust-worker-sysctl"
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: "locust-worker"
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: "locust-worker"
+    spec:
+      nodeSelector:
+        cloud.google.com/gke-nodepool: "locust-workers"
+      hostNetwork: true
+      initContainers:
+        - name: sysctl
+          image: alpine:3
+          command:
+            - /bin/sh
+            - -c
+            - |
+              sysctl net.netfilter.nf_conntrack_max=1048576
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              cpu: 10m
+              memory: 10Mi
+            limits:
+              cpu: 10m
+              memory: 10Mi
+      containers:
+        - name: sleep
+          image: alpine:3
+          command:
+            - /bin/sh
+            - -c
+            - |
+              while true; do sleep 60s; done
+          resources:
+            requests:
+              cpu: 10m
+              memory: 10Mi
+            limits:
+              cpu: 10m
+              memory: 10Mi
--- a/tests/load/setup_k8s.sh
+++ b/tests/load/setup_k8s.sh
@@ -10,7 +10,7 @@ CLUSTER='autopush-locust-load-test'
 TARGET='https://updates-autopush.stage.mozaws.net'
 SCOPE='https://www.googleapis.com/auth/cloud-platform'
 REGION='us-central1'
-WORKER_COUNT=100
+WORKER_COUNT=150
 MACHINE_TYPE='c3d-standard-4' # 4 CPUs + 16GB Memory
 BOLD=$(tput bold)
 NORM=$(tput sgr0)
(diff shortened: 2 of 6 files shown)


Trinaa:
Notes from @emaydeck-mozilla:

1. enabling dataplane-v2

  a. this was a suggestion made previously by GCP support
  b. this uses a GKE dataplane that's optimized for kubernetes networking. more info: https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2
  c. it's implemented pretty easily by just adding a flag for it in the gcloud containers create command
  d. here's a commit showing that change: https://github.com/mozilla-services/autopush-rs/pull/593/commits/45bce63c75acd2ba3e4fbb75359d7a04c2c97361
 
2. leveraging a static policy when scheduling pods on nodes

a. this allows containers in pods running in the Guaranteed QoS class access to exclusive CPUs on the worker nodes
b. [the distributed load generation locust doc](https://docs.locust.io/en/stable/running-distributed.html) recommends 1 core per worker, so i figured scheduling 4 pods @ 1 CPU each would achieve our goal
c. in order to make this change (and to make it easier for us to tweak kubernetes settings for worker nodes), i created a new node pool using the static policy configuration that's dedicated to just the locust workers
    - the worker pods were configured to specifically only schedule on nodes within this dedicated node pool
    - the master controller will still run on the default node pool
d. some other changes to the worker yaml file:
    - in order to run using the Guaranteed QoS class, the CPU requests and limits must be matching integers... in this case, 1 for both
    - i bumped memory to 3 GB for both requests and limits since the machine types have 16 GB of memory in total
e. here's a commit showing that change: https://github.com/mozilla-services/autopush-rs/pull/593/commits/1eda9cdf3dbdeff40c7e92df61d705efd03f3a08

tests/load/setup_k8s.sh
```diff
@@ -98,13 +99,16 @@ do
     case $response in
         create) #Setup Kubernetes Cluster
             echo -e "==================== Creating the GKE cluster "
-            # The total-max-nodes = WORKER_COUNT + 1 (MASTER)
-            $GCLOUD container clusters create $CLUSTER --region $REGION --scopes $SCOPE --enable-autoscaling --total-min-nodes "1" --total-max-nodes "301" --scopes=logging-write,storage-ro --addons HorizontalPodAutoscaling,HttpLoadBalancing  --machine-type $MACHINE_TYPE
+            $GCLOUD container clusters create $CLUSTER --region $REGION --scopes $SCOPE --enable-autoscaling --scopes=logging-write,storage-ro --addons HorizontalPodAutoscaling,HttpLoadBalancing --enable-dataplane-v2
```
Trinaa, line 102: Are we missing `--machine-type=$MACHINE_TYPE` here cause I noticed an e2-medium is being assigned:
image

Trinaa:
Working on the following error:

```
==================== Deploy the Locust master and worker nodes
deployment.apps/locust-master created
service/locust-master created
Error from server (BadRequest): error when creating "/home/kanderson/autopush-rs/tests/load/kubernetes-config/locust-worker-controller.yml": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.template.spec.containers[0].nodeSelector"
```

tests/load/setup_k8s.sh
emaydeck-mozilla: Ah yeah, we could add something in if we wanted a machine type other than an e2-medium there. This node pool should only be used for the Kubernetes management stuff and the Locust master service deployment.

Trinaa: I do not know the capacity of the e2-medium and if it can handle the locust-master. locust-master needs to handle a lot of networking with the workers, so tends to be hungry.

tests/load/kubernetes-config/locust-worker-daemonset.yml
```diff
@@ -0,0 +1,49 @@
...
+            privileged: true
+          resources:
+            requests:
+              cpu: 10m
+              memory: 10Mi
+            limits:
+              cpu: 10m
+              memory: 10Mi
+      containers:
+        - name: sleep
```
jrconlin, line 36: Curious what this container is for.

jrconlin (approved):

tests/load/kubernetes-config/locust-worker-daemonset.yml
Trinaa: Good question.... @dlactin or @emaydeck-mozilla?

jrconlin (approved):

pjenvey (approved):

PR #597, opened 2024-02-01
bug: add secondary_write for channels

jrconlin:
* also set the default `write_to_secondary` to true

Closes SYNC-4121

### DIFF
--- a/autopush-common/src/db/dual/mod.rs
+++ b/autopush-common/src/db/dual/mod.rs
@@ -27,9 +27,15 @@ use super::StorageType;
 
 #[derive(Clone)]
 pub struct DualClientImpl {
+    /// The primary data store, which will always be Bigtable.
     primary: BigTableClientImpl,
+    /// The secondary data store, which will always be DynamoDB.
     secondary: DdbClientImpl,
+    /// Write changes to the secondary, including messages and updates
+    /// as well as account and channel additions/deletions.
     write_to_secondary: bool,
+    /// Hex value to use to specify the first byte of the median offset.
+    /// e.g. "0a" will start from include all UUIDs upto and including "0a"
     median: Option<u8>,
 }
 
@@ -39,10 +45,16 @@ fn default_true() -> bool {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct DualDbSettings {
+    /// The primary data store, which will always be Bigtable.
     primary: DbSettings,
+    /// The secondary data store, which will always be DynamoDB.
     secondary: DbSettings,
+    /// Write changes to the secondary, including messages and updates
+    /// as well as account and channel additions/deletions.
     #[serde(default = "default_true")]
     write_to_secondary: bool,
+    /// Hex value to use to specify the first byte of the median offset.
+    /// e.g. "0a" will start from include all UUIDs upto and including "0a"
     #[serde(default)]
     median: Option<String>,
 }
@@ -50,11 +62,11 @@ pub struct DualDbSettings {
 impl DualClientImpl {
     pub fn new(metrics: Arc<StatsdClient>, settings: &DbSettings) -> DbResult<Self> {
         // Not really sure we need the dsn here.
-        info!("Trying: {:?}", settings.db_settings);
+        info!("⚖ Trying: {:?}", settings.db_settings);
         let db_settings: DualDbSettings = from_str(&settings.db_settings).map_err(|e| {
             DbError::General(format!("Could not parse DualDBSettings string {:?}", e))
         })?;
-        debug!("settings: {:?}", &db_settings.median);
+        debug!("⚖ settings: {:?}", &db_settings);
         if StorageType::from_dsn(&db_settings.primary.dsn) != StorageType::BigTable {
             return Err(DbError::General(
                 "Invalid primary DSN specified (must be BigTable type)".to_owned(),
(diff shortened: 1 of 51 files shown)


jrconlin:
There are a few extra debug and trace bits in there to help me identify a problem with my local testing.

autopush-common/src/db/dual/mod.rs
```diff
@@ -172,14 +183,22 @@ impl DbClient for DualClientImpl {
 
     async fn add_channel(&self, uaid: &Uuid, channel_id: &Uuid) -> DbResult<()> {
         debug!("⚖ getting target");
-        let (target, _) = self.allot(uaid).await?;
+        let (target, is_primary) = self.allot(uaid).await?;
         debug!("⚖ Adding channel to {}", target.name());
-        target.add_channel(uaid, channel_id).await
+        let result = target.add_channel(uaid, channel_id).await;
+        if is_primary {
```
pjenvey, line 189: These `if is_primary { <write to secondary>` additions as well as some existing others don't include `&& self.write_to_secondary`. Is that intentional?

jrconlin: A bit, yes. I figured that we always want to perform those operations. The other ones optionally alter data, but otherwise leave records in tact. Since these are changes to the user record or subscription, I figured we should reflect them.

autopush-common/src/db/dual/mod.rs
```diff
@@ -119,22 +141,37 @@ impl DualClientImpl {
...
         //  If the UAID is in the allowance, move them to the new data store
         let (target, is_primary) = self.allot(&user.uaid).await?;
+        let result = target.update_user(user).await?;
         if is_primary && self.write_to_secondary {
-            let _ = self.secondary.add_user(user).await?;
+            let _ = self.secondary.add_user(user).await.map_err(|e| {
+                error!("⚡ Error: {:?}", e);
+                self.metrics
+                    .incr_with_tags("database.dual.error")
+                    .with_tag("func", "remove_user")
```
pjenvey, line 169:
```suggestion
                    .with_tag("func", "update_user")
```

autopush-common/src/db/dual/mod.rs
```diff
@@ -165,21 +202,56 @@ impl DbClient for DualClientImpl {
...
-        target.add_channel(uaid, channel_id).await
+        let result = target.add_channel(uaid, channel_id).await;
+        if is_primary && self.write_to_secondary {
+            let _ = self
+                .secondary
+                .add_channel(uaid, channel_id)
+                .await
+                .map_err(|e| {
+                    self.metrics
+                        .incr_with_tags("database.dual.allot")
```
pjenvey, line 229:
```suggestion
                        .incr_with_tags("database.dual.error")
```

autopush-common/src/db/dual/mod.rs
```diff
@@ -165,21 +202,56 @@ impl DbClient for DualClientImpl {
...
+        let (target, is_primary) = self.allot(uaid).await?;
+        let result = target.add_channels(uaid, channels.clone()).await;
+        if is_primary && self.write_to_secondary {
+            let _ = self
+                .secondary
+                .add_channels(uaid, channels)
+                .await
+                .map_err(|e| {
+                    self.metrics
+                        .incr_with_tags("database.dual.allot")
```
pjenvey, line 248:
```suggestion
                        .incr_with_tags("database.dual.error")
```

pjenvey (changes requested):
couple things:

- save_message and save_messages are missing the metrics on ignoring the secondary error
- another thing we should fix while we're in here: `get_channels` has an inaccurate comment about having to copy over channels, this is probably from before we added the `add_channels` copying in `get_user`. Also I would argue that it's unique in that it doesn't need to check secondary at all, ever? So it would probably make sense to remove that block entirely? (It certainly shouldn't need to check secondary when `channels.is_empty()` I would more strongly argue)

pjenvey:
> * another thing we should fix while we're in here: `get_channels` has an inaccurate comment about having to copy over channels, this is probably from before we added the `add_channels` copying in `get_user`. Also I would argue that it's unique in that it doesn't need to check secondary at all, ever? So it would probably make sense to remove that block entirely? (It certainly shouldn't need to check secondary when `channels.is_empty()` I would more strongly argue)

though I suppose we could always have an edge case where node A migrates the user record, then node B reads it and the channels from bigtable before node A migrates the channels..

jrconlin:
>     * save_message and save_messages are missing the metrics on ignoring the secondary error
Both of those return `DbResult<()>` so we're passing the errors up. I figured we don't need to capture metrics for those because  we're not eating the errors. 
We are eating the `Ok`, but then it's just returning `()` anyway.

jrconlin:
> though I suppose we could always have an edge case where node A migrates the user record, then node B reads it and the channels from bigtable before node A migrates the channels.. 

That comment and the code pre-date your change to the channels, so it's fair to just drop it. I think the edge case is fairly minor and at worst would result in the client rebuilding it's list (if it's mobile, at least). We *might* also hit an equally edge case of an endpoint fetching the User Record and not finding a channel, then returning a 404 response, but that may also be temporary since we know providers don't generally pay attention to return codes.

pjenvey:
> Both of those return `DbResult<()>` so we're passing the errors up. I figured we don't need to capture metrics for those because we're not eating the errors.
> We are eating the `Ok`, but then it's just returning `()` anyway.

ah, my mistake 👍

PR #599, opened 2024-02-05
feat: filter reads by latest cell

pjenvey:
and generally add the exact GC policy of the column family to all reads

also fix User::version from writing to DynamoDB (it's Bigtable only)

Issue: SYNC-4094

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -74,7 +74,16 @@ pub struct BigTableClientImpl {
     pool: BigTablePool,
 }
 
-fn timestamp_filter() -> Result<data::RowFilter, error::BigTableError> {
+/// Return a a RowFilter matching the GC policy of the router Column Family
+fn router_gc_policy_filter() -> data::RowFilter {
+    let mut latest_cell_filter = data::RowFilter::default();
+    latest_cell_filter.set_cells_per_column_limit_filter(1);
+    latest_cell_filter
+}
+
+/// Return a chain of RowFilters matching the GC policy of the message Column
+/// Families
+fn message_gc_policy_filter() -> Result<Vec<data::RowFilter>, error::BigTableError> {
     let mut timestamp_filter = data::RowFilter::default();
     let bt_now: i64 = SystemTime::now()
         .duration_since(SystemTime::UNIX_EPOCH)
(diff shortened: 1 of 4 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1426,12 +1440,10 @@ mod tests {
         client.remove_user(&uaid).await
     }
 
-    /*
-    // XXX: uncomment after the uaid clashing fix
     #[actix_rt::test]
     async fn add_user_existing() {
         let client = new_client().unwrap();
-        let uaid = Uuid::parse_str(TEST_USER).unwrap();
+        let uaid = Uuid::new_v4();
```
jrconlin, line 1446: Not critical, but I still would like to tag these just in case someone does something bad like run unit tests against a production database.

jrconlin (approved):

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1442,5 +1454,24 @@ mod tests {
...
+
+    #[actix_rt::test]
+    async fn version_check() {
+        let client = new_client().unwrap();
+        let uaid = Uuid::new_v4();
+        let user = User {
+            uaid,
+            ..Default::default()
+        };
+        client.remove_user(&uaid).await.unwrap();
```
jrconlin, line 1466: We definitely want to use a "safer" value here. 

probably also should clean it up at the end of the test.

jrconlin (changes requested):

autopush-common/src/db/bigtable/bigtable_client/mod.rs
pjenvey: I believe you had some code to randomize these uuids with a prefix somewhere, if you point me to it I'll add it to this branch to resolve both comments

jrconlin: https://github.com/mozilla-services/autopush-rs/pull/588/files#diff-773a699c0731038f9168c5bb9ffe0ce0aad1f5d05c835bdd6edfe20f2dfa050fR1

pjenvey: I thought it was on some WIP/draft branch, no of course it's in a PR I've been ignoring 🤦‍♂️

jrconlin (approved):

jrconlin (approved):

PR #608, opened 2024-02-08
Feat/[SYNC-4131] add lint and bandit

taddes:
To increase developer experience for the python portion of our service - integration and load testing - this patch adds the following:

- an auto format command
- scoped lint commands for mypy, pydocstyle, black, flake8, etc with more rich output. This is useful in larger features (most especially mypy) where mass output at the lint stage can be somewhat unhelpful.
- Addition of bandit as part of the lint flow. Bandit is used for static analysis of the code for potential security vulnerabilities. 

Note: added `#  nosec` annotations for several unworrisome issues. Do want some opinion on `requests.get` not having timeouts and a possible replacement (may be docker/testcontainers) for the `subprocess.call` invocations, however since these are in the tests they're not a huge concern.

### DIFF
--- a/tests/integration/push_test_client.py
+++ b/tests/integration/push_test_client.py
@@ -149,7 +149,7 @@ keyid="http://example.org/bob/keys/123";salt="XZwpw6o37R-6qoZjw6KwAw=="\
             self.channels[chid] = result["pushEndpoint"]
         return result
 
-    def unregister(self, chid):
+    def unregister(self, chid) -> Any:
         """Unregister the ChannelID, which should invalidate the associated Endpoint."""
         if not self.ws:
             raise Exception("WebSocket client not available as expected")


Trinaa (dismissed):
This LGTM! I was able to execute the make command locally without error.

imo the requests.get could have timeouts, it could be a helpful signal of performance degradation should they trigger. I'm not familiar with subprocess.call warnings, but the solution sounds like it would be outside of scope for this task if we address it.

jrconlin (dismissed):
Heh, we include bandit, and then disable it for a lot of things. 

We might want to indicate why we're calling `#nosec`.

taddes:
Hey @Trinaa & @jrconlin  , thanks! In terms of timeouts, I know that push is sensitive when it comes to latency, so in being cautious, would something like a 5 s timeout be sufficient, in your opinions? For `subprocess.call` I will note that going forward, may be a different way to do it once we get to Dockerized versions of the tests.

And yes @jrconlin , haha, we bring Bandit it only to ignore it. Sigh. We have the same in Merino test repos, since it's not the production code that's at risk, so little `# nosec` annotations appear throughout for things that end up actually not being risky. Especially with the `random.choice` calls or `sha1` in load tests, it really doesn't like that (though for good reason were it a part of the running code!). I'll add the refs as comments there, that is very wise.

jrconlin:
5s might be a bit too aggressive. 30s might be better? Depends on how bogged down the CI servers might be, I presume.

taddes:
@jrconlin added those timeouts and made some comments on nosec. Had to rebase onto master again, hence a re-approval needed.

jrconlin (approved):

PR #610, opened 2024-02-08
bug: Add errors for health check reporting

jrconlin:
(NOTE: This uses the `error!()` macro.)

### DIFF
--- a/autoconnect/autoconnect-web/src/dockerflow.rs
+++ b/autoconnect/autoconnect-web/src/dockerflow.rs
@@ -26,6 +26,7 @@ pub fn config(config: &mut web::ServiceConfig) {
 }
 
 /// Handle the `/health` and `/__heartbeat__` routes
+#[allow(clippy::blocks_in_conditions)]
 pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
     let status = if state
         .db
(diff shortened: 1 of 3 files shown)


autoconnect/autoconnect-web/src/dockerflow.rs
```diff
@@ -26,8 +26,19 @@ pub fn config(config: &mut web::ServiceConfig) {
 }
 
 /// Handle the `/health` and `/__heartbeat__` routes
+// note, this changes to `blocks_in_conditions` for 1.76+
+#[allow(clippy::blocks_in_if_conditions)]
```
pjenvey, line 30: Half of me wonders if this case is even supposed to trigger `blocks_in_if_conditions` whereas half of me also agrees we should just put this entire, longer expression on its own let line instead of quieting clippy (but not going to block on it)

pjenvey (approved):

autoconnect/autoconnect-web/src/dockerflow.rs
jrconlin: I was honestly wondering the same thing, but decided to do this mostly because I found the long form really hard to read in comparison. 

Here, at the point we've gotten the error, we're firing an `error!()` macro. We *could* do something with `.else()` but then getting the error back would be a pain, so you have to a `match` and blargh.

PR #611, opened 2024-02-08
feat: Add timeouts for bigtable operations

jrconlin:


pjenvey (changes requested):
Few things:

- Missed the `drop_row_range_async` call
- Thinking about this more I realize though these are nice to have, they don't really solve the timeout problem we just had, which is actually in the `Pool` when establishing a connection. We already have a timeout solution for that: `BigTableSettings::database_pool_connection_timeout` -- we should just set one in our deployment config
- 5 seems maybe a bit too strict (might result in some more sentry chatter)? Looking at the [backend_latencies](https://earthangel-b40313e5.influxcloud.net/d/do4mmwcVz/autopush-gcp?orgId=1&editPanel=31&from=now-30d&to=now&refresh=1m) if I'm reading it correctly (switching it to max, not sure it's accurate because it claims to be a mean of [backend_latencies](https://cloud.google.com/load-balancing/docs/metrics)) we do get some longer spikes that tend to max out at 30s on DynamoDB which I bet has a server side timeout, I won't be surprised if Bigtable does too
- I would also prefer `Option`s for this and the related pool settings so they can be totally disabled

jrconlin:
Switching things over to use Options (which I agree, is a better idea). Also removed the default timeout since we made the request timeout optional as well. We can specify our own value that way, which is probably less of a foot-gun.

I thought that the pool has [options]( for how long to wait for a slot, how long to wait to create a slot, and how long before recycling. Nothing for how long a given request should take. That seemed to be controlled by grpc, and I think the only way to do that was with the `CallOptions`.

pjenvey:
> I thought that the pool has [options](https://docs.rs/deadpool/latest/deadpool/managed/struct.Timeouts.html) for how long to wait for a slot, how long to wait to create a slot, and how long before recycling. Nothing for how long a given request should take. That seemed to be controlled by grpc, and I think the only way to do that was with the `CallOptions`.

Indeed all those deadpool timeouts don't apply to operations this PR modifies (so like I said these additions are nice to have also) but we apply our `database_pool_connection_timeout` setting to `Timeouts::create` which is responsible for [creating the actual Bigtable connection](https://github.com/mozilla-services/autopush-rs/blob/cb1482de2f5cedef0047ca9a7b01db0d7103e5ab/autopush-common/src/db/bigtable/pool.rs#L155).

My point was the problem we recently encountered was rooted there: we failed to ever [establish the initial connection](https://github.com/mozilla-services/autopush-rs/blob/cb1482de2f5cedef0047ca9a7b01db0d7103e5ab/autopush-common/src/db/bigtable/pool.rs#L208) due to the bad port number. If we had a reasonable `deadpool::Timeouts::create` it would have cancelled the task (via [tokio::timeout](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html)) a lot quicker than apparently the 5+ minutes `connect` was taking.

I'll also note that syncstorage-rs applies the same `database_pool_connection_timeout` setting to `Timeouts::wait` instead of `create`, we should probably make these config knobs act the same between the two, and maybe we should support setting all 3 of the `Timeouts` variants, or apply one config knob to all of them (but maybe not that in the long term), we can clean that up later -- but we might need to support a `wait` setting in Bigtable sooner rather than later.

pjenvey (changes requested):

pjenvey (approved):

PR #619, opened 2024-02-16
fix: fix add/update_user version field handling on migrations

pjenvey:
- explicitly add a version (which there is None from DynamoDB) in dual mode migration
- Bigtable::add_user now writes the initial version passed in
- Bigtable::update_user continues always writing a new version but now modifies the &mut User upon success

Closes: SYNC-4147


jrconlin (approved):
I'm a bit concerned about the `debug_assert!` winding up in production, but not enough to block this.

jrconlin (approved):
I'm a bit concerned about the `debug_assert!` winding up in production, but not enough to block this.

pjenvey:
@jrconlin release builds compile it out, so we have bigger problems if that ever happens

PR #622, opened 2024-02-19
docs: update path to push architecture schema

DanTrofimov:
Fixes 404 for push architecture schema at docs 

image

PR #623, opened 2024-02-20
bug: Add missing doc diagram.

jrconlin:
This fixes an error introduced by #613 and resolves PR #622


pjenvey (changes requested):
#613 replaced the old `docs/src/assets/push_architecture.svg` with a new one so we don't need to upload a [new one](https://github.com/mozilla-services/autopush-rs/tree/9f361c9cef9c5113fd70907bfd07eaec286ece01/docs/src/assets), just fix the link.

jrconlin:
I would have liked to keep the older image around, partly for history and partly because of back link reasons, but whatever.

pjenvey (approved):

PR #627, opened 2024-02-20
feat: have Bigtable match DynamoDB's metrics

pjenvey:
(not including conversion errors which never occur)

Closes: SYNC-4150


autopush-common/src/db/dual/mod.rs
```diff
@@ -128,10 +134,6 @@ impl DualClientImpl {
         } else {
             (Box::new(&self.primary), true)
         };
-        self.metrics
```
pjenvey, line 131: This was using the wrong metric label but it seemed too chatty to emit it for every single `allot` call, so I've moved this to emit its configured median only once at startup.

I think the combination of the added metrics/tags in this PR and the db's own metrics should give us a sense of what `allot` is doing anyway.

jrconlin (approved):

jrconlin (approved):

PR #634, opened 2024-02-22
fix: don't add_channels when there's none to add

pjenvey:
Closes: SYNC-4154


jrconlin (approved):

jrconlin:
We should probably also have `BigTableClientImpl::add_channels()` return `Ok(())` if `channels.is_empty()`

PR #636, opened 2024-02-22
feat: prefer the simpler delete to avoid the admin API

pjenvey:
Closes: SYNC-4155


jrconlin (approved):

jrconlin:
We may be eating the `raw` error here since it's being included as `extra` and that's not propagating correctly until https://github.com/mozilla-services/autopush-rs/pull/632 lands. If this error continues after this fix, we'll at least have better info about why.

PR #639, opened 2024-02-24
feat: Add `profile_app_id` routing hint to bigtable requests

jrconlin:
Based on [what I found](https://github.com/googleapis/google-cloud-python/blob/6838a4f11162544c0654f3ce99fac5a7c510237a/bigtable/google/cloud/bigtable_v2/proto/bigtable_pb2.py#L1245) in th e python lib.

Closes: SYNC-4153

### DIFF
--- a/autopush-common/src/db/bigtable/mod.rs
+++ b/autopush-common/src/db/bigtable/mod.rs
@@ -44,6 +44,10 @@ pub struct BigTableDbSettings {
     /// bigtable.
     #[serde(default)]
     pub table_name: String,
+    /// Routing replication profile id.
+    /// Should be used everywhere we set `table_name` when creating requests
+    #[serde(default)]
+    pub profile_id: String,
     #[serde(default)]
     pub router_family: String,
     #[serde(default)]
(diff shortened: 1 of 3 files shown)


autopush-common/src/db/bigtable/mod.rs
```diff
@@ -48,6 +48,10 @@ pub struct BigTableDbSettings {
     /// bigtable.
     #[serde(default)]
     pub table_name: String,
+    /// Routing replication profile id.
+    /// Should be used everywhere we set `table_name` when creating requests
+    #[serde(default)]
+    pub profile_id: String,
```
pjenvey, line 54: Could we rename this to `app_profile_id` (Bigtable docs seem to always say  "app" along with profile). I think we can just do this with a `serde(rename = "app_profile_id")` if you want to save the shorter name for the internal code.

pjenvey (changes requested):

autopush-common/src/db/bigtable/mod.rs
jrconlin: I'm fine just changing the symbol name to `app_profile_id`. Might as well not make things confusing later.

pjenvey (approved):

PR #640, opened 2024-02-26
fix: don't ignore add_user errors

pjenvey:
follow up to 87ed38c3

(where we lost a trailing `return Err(e);` line: so refactored to make that clearer)


jrconlin (approved):
still really wish that there was a different negation symbol for macros other than `!macro!(`

PR #643, opened 2024-02-26
feat: Attempt retry for Bigtable read row.

jrconlin:
This is an exploritory patch to see if we should retry read operations if we get a Bigtable oauth2 token request error.

Issue: SYNC-4157


jrconlin:
This would be aided by https://github.com/mozilla-services/autopush-rs/pull/591

jrconlin:
Closing in favor of #591

PR #644, opened 2024-02-26
feat: report db pool metrics

pjenvey:
and support configuring all of deadpool's Timeouts

Closes: SYNC-4158


jrconlin (commented):
I take it we can drop https://github.com/mozilla-services/autopush-rs/pull/611?

(If so, we should probably change the default GRPC port as well.

jrconlin (approved):

PR #648, opened 2024-03-06
feat: include tags/extras across the exception chain

pjenvey:
- add extras for diagnosing db errors in SYNC-4172
- update mio per RUSTSEC-2024-0019

Closes: SYNC-4145
Issue: SYNC-4172


jrconlin (approved):
Heh, at first I thought that this duplicated #632

PR #660, opened 2024-03-14
bug: Fix metric message text for source of message

jrconlin:
Closes SYNC-4193


pjenvey (changes requested):
The way I'm understanding this metric it seems to have the correct name: it's identical to the other `message_data` metric emitted by `incr_success_metrics`. They're both called when a message has been successfully delivered, either via webpush or an external bridge.

If anything maybe we should be calling `incr_success_metrics` here but we reserve its `notification.bridge.sent` metric solely for external bridges (not webpush). Its third metric `notification.total_request_time` we emit elsewhere in this module (for some reason, only after `trigger_notification_check`. I think we forgot to add the same metric for after a successful `send_notification` call).

I think this should probably be calling a new common function that solely emits the `message_data` metric that `incr_success_metrics` would also call? Then both versions would include the same tags (they currently don't).

jrconlin:
Huh. Good call on having a common function that emits the set of metrics. I'll dig into that. 

Mostly, this was just me trying to fix up some of the weird discrepancies around metric calls so centralizing them a bit will help.

PR #661, opened 2024-03-14
test(load): Document the process for calibrating users per worker 

Trinaa:
SYNC-4195

### DIFF
--- a/tests/load/README.md
+++ b/tests/load/README.md
@@ -100,7 +100,8 @@ Follow the steps bellow to execute the distributed load tests on GCP:
 
 The load tests can be executed from the [contextual-services-test-eng cloud shell][15].
 If executing a load test for the first time, the git autopush repository will need to
-be cloned locally into the root directory of the machine. Use the HTTPS git clone functionality as SSH is not supported in this environment.
+be cloned locally into the root directory of the machine. Use the HTTPS git clone
+functionality as SSH is not supported in this environment.
 
 #### 2. Configure the Bash Script
 
@@ -210,9 +211,10 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
 
 ## Calibration
 
-Following the addition of new features, such as a Locust Task or Locust User, or environmental 
-changes, such as node size or the upgrade of a major dependency like the python version image, it may be necessary to re-establish the recommended parameters of a 
-performance test.
+Following the addition of new features, such as a Locust Task or Locust User, or
+environmental changes, such as node size or the upgrade of a major dependency like the
+python version image, it may be necessary to re-establish the recommended parameters of
+the performance test.
 
 | Parameter         | Description                                                                                                                                                                                                      |
 |-------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
@@ -261,11 +263,13 @@ be cloned locally.
 
 ### Calibrate 
 
-Repeat steps 1 to 3, using a process of elimination to determine the maximum 
-`USERS_PER_WORKER`. The load tests are considered optimized when CPU and memory resources
-are maximally utilized. This step is meant to determine the maximum user count that a
-node can accommodate by observing CPU and memory usage while steadily increasing or
-decreasing the user count. You can monitor the CPU percentage in the Locust UI but also in the Kubernetes engine Workloads tab where both memory and CPU are visualized on charts. 
+Repeat steps 1 to 3, using a process of elimination, such as the bisection method, to
+determine the maximum `USERS_PER_WORKER`. The load tests are considered optimized when
+CPU and memory resources are maximally utilized. This step is meant to determine the
+maximum user count that a node can accommodate by observing CPU and memory usage while
+steadily increasing or decreasing the user count. You can monitor the CPU percentage in
+the Locust UI but also in the Kubernetes engine Workloads tab where both memory and CPU
+are visualized on charts. 
 
 #### 1. Start Load Test
 
--- a/tests/load/README.md
+++ b/tests/load/README.md
@@ -100,7 +100,7 @@ Follow the steps bellow to execute the distributed load tests on GCP:
 
 The load tests can be executed from the [contextual-services-test-eng cloud shell][15].
 If executing a load test for the first time, the git autopush repository will need to
-be cloned locally.
+be cloned locally into the root directory of the machine. Use the HTTPS git clone functionality as SSH is not supported in this environment.
 
 #### 2. Configure the Bash Script
 
@@ -211,15 +211,15 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
 ## Calibration
 
 Following the addition of new features, such as a Locust Task or Locust User, or environmental 
-changes, such as node size it may be necessary to re-establish the recommended parameters of a 
+changes, such as node size or the upgrade of a major dependency like the python version image, it may be necessary to re-establish the recommended parameters of a 
 performance test.
 
 | Parameter         | Description                                                                                                                                                                                                      |
 |-------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | [WAIT TIME][13]   | - Changing this cadence will increase or decrease the number of channel subscriptions and notifications sent by an AutopushUser. <br/>- This value can be set in the Locust UI.                                  |
 | [TASK WEIGHT][14] | - Changing this weight impacts the probability of a task being chosen for execution. <br/>- This value is hardcoded in the task decorators of the AutopushUser class.                                            |
-| USERS_PER_WORKER  | - This value should be set to the maximum number of users a Locust worker can support given CPU and memory constraints. <br/>- This value is hardcoded in the AutopushLoadTestShape class.                       |
-| WORKER_COUNT      | - This value is derived by dividing the total number of users needed for the performance test by the 'USERS_PER_WORKER'. <br>- This value is hardcoded in the AutopushLoadTestShape and the setup_k8s.sh script. |
+| `USERS_PER_WORKER`  | - This value should be set to the maximum number of users a Locust worker can support given CPU and memory constraints. <br/>- This value is hardcoded in the AutopushLoadTestShape class.                       |
+| `WORKER_COUNT`      | - This value is derived by dividing the total number of users needed for the performance test by the `USERS_PER_WORKER`. <br>- This value is hardcoded in the AutopushLoadTestShape and the setup_k8s.sh script. |
 
 ## Calibrating for USERS_PER_WORKER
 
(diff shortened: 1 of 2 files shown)


tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+
+#### 4. Report Results
+
+* See [Distributed GCP Execution - Report Results](#4-report-results-1)
+* Only client-side measures, provided by Locust, are available for local execution
+
+#### 5. Update Shape and Script Values
+
+* WORKER_COUNT = MAX_USERS/USERS_PER_WORKER
+    * If MAX_USERS is unknown, calibrate to determine WORKER_COUNT
```
Trinaa, line 318: Note: Instructions for this are coming [SYNC-4196](https://mozilla-hub.atlassian.net/browse/SYNC-4196)

jrbenny35 (dismissed):
Just some small nits, looks good though.

tests/load/README.md
taddes: Great! Thanks for that @Trinaa

tests/load/README.md
```diff
@@ -64,10 +64,10 @@ make load
 * Set up the load test parameters:
     * ShapeClass: Default
     * Number of users: 1
-    * Spawn rate: 1
+    * Ramp up: 1
```
taddes, line 67: Great catch given the UI change we observed ! 👍

tests/load/README.md
```diff
@@ -94,23 +94,13 @@ make load-clean
...
-| USERS PER WORKER  | - This value should be set to the maximum number of users a Locust worker can support given CPU and memory constraints. <br/>- This value is hardcoded in the AutopushLoadTestShape class.                       |
-| LOCUST WORKERS    | - This value is derived by dividing the total number of users needed for the performance test by the 'USERS PER WORKER'. <br>- This value is hardcoded in the AutopushLoadTestShape and the setup_k8s.sh script. |
-
 ### Setup Environment
 
 #### 1. Start a GCP Cloud Shell
 
 The load tests can be executed from the [contextual-services-test-eng cloud shell][15].
+If executing a load test for the first time, the git autopush repository will need to
+be cloned locally.
```
taddes, line 103: Maybe a note that someone should use the HTTPS git clone command? Since SSH isn't enabled in that context.

taddes:
```suggestion
be cloned locally into the root directory of the machine. Use the HTTPS git clone functionality as SSH is not supported in this environment.
```

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
 ./tests/load/setup_k8s.sh
 ```
 
+## Calibration
+
+Following the addition of new features, such as a Locust Task or Locust User, or environmental 
+changes, such as node size it may be necessary to re-establish the recommended parameters of a
```
taddes, line 214:
```suggestion
changes, such as node size or the upgrade of a major dependency like the python version image, it may be necessary to re-establish the recommended parameters of a 
```

thoughts?

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+  kubectl get svc locust-master --watch
+  ```
+
+### Calibrate 
+
+Repeat steps 1 to 3, using a process of elimination to determine the maximum 
+USERS_PER_WORKER. The load tests are considered optimized when CPU and memory resources
+are maximally utilized. This step is meant to determine the maximum user count that a
+node can accommodate by observing CPU and memory usage while steadily increasing or
+decreasing the user count.
```
taddes, line 268: By the way, is this a place you'd want to mention that bisection formula? I think since it provides a clear way for someone to know how to +/- the number of users. Maybe just a short blurb or formula example? 

Perhaps add a little extra too that the CPU and memory can be monitored in the GCS console and the Locust UI:

```suggestion
decreasing the user count. You can monitor the CPU percentage in the Locust UI but also in the Kubernetes engine Workloads tab where both memory and CPU are visualized on charts. 
```

taddes (dismissed):
Splendid! I did add a comment about adding that formula (bisect method?) we discussed being added in the step for calibration. Then there are just a few small things. Feel free to add or not add whatever you prefer, what we have here now is great! Thanks for doing this @Trinaa 🎆

jrconlin (commented):
Please note that link `[4]` goes to the ConServ Google Drive directory, which does not contain Load Test Artifacts.

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
 ./tests/load/setup_k8s.sh
 ```
```
jrconlin, line 209: If we're not demonstrating the `delete` option, do we need to include this `shell` command?

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+      chmod +x tests/load/setup_k8s.sh
+      ```
+
+#### 3. Create the GCP Cluster
+
+* In the `setup_k8s.sh` script, modify the `WORKER_COUNT` variable to equal `1`
+* Execute the `setup_k8s.sh` file from the root directory and select the **create** 
+  option, in order to initiate the process of creating a cluster, setting up the env 
+  variables and building the docker image
+  ```shell
```
jrconlin, line 252: Again, not sure we need to include this if we're not demonstrating selecting the **create** option or specifying an argument to indicate selection of the **create** command.

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+  an external IP is assigned to the `locust_master` node. Monitor the assignment via
+  a watch loop:
+  ```bash
+  kubectl get svc locust-master --watch
+  ```
+
+### Calibrate 
+
+Repeat steps 1 to 3, using a process of elimination to determine the maximum 
+USERS_PER_WORKER. The load tests are considered optimized when CPU and memory resources
```
jrconlin, line 265: Since this is a code or environment variable, it may help to highlight it as such.

```suggestion
`USERS_PER_WORKER`. The load tests are considered optimized when CPU and memory resources
```

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+
+* CPU and Memory usage should be less than 90% of the available capacity
+    * CPU and Memory Resources can be observed in 
+      [Google Cloud > Kubernetes Engine > Workloads][18]
+
+**Log Errors or Warnings**
+
+* Locust will emit errors or warnings if high CPU or memory usage occurs during the 
+  execution of a load test. The presence of these logs is a strong indication that the 
+  USERS_PER_WORKER count is too high
```
jrconlin, line 305:
```suggestion
  `USERS_PER_WORKER` count is too high
```

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+      production because of the use of Shape classes
+
+#### 4. Report Results
+
+* See [Distributed GCP Execution - Report Results](#4-report-results-1)
+* Only client-side measures, provided by Locust, are available for local execution
+
+#### 5. Update Shape and Script Values
+
+* WORKER_COUNT = MAX_USERS/USERS_PER_WORKER
```
jrconlin, line 317:
```suggestion
* `WORKER_COUNT = MAX_USERS/USERS_PER_WORKER`
```

tests/load/README.md
jrconlin:
```suggestion
    * If `MAX_USERS` is unknown, calibrate to determine `WORKER_COUNT` 
```

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+    * If MAX_USERS is unknown, calibrate to determine WORKER_COUNT 
+* Update the USERS_PER_WORKER and WORKER_COUNT values in the following files:
```
jrconlin, line 319:
```suggestion
* Update the `USERS_PER_WORKER` and `WORKER_COUNT` values in the following files:
```

tests/load/README.md
```diff
@@ -219,6 +208,128 @@ Execute the `setup_k8s.sh` file from the root directory and select the **delete*
...
+performance test.
+
+| Parameter         | Description                                                                                                                                                                                                      |
+|-------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [WAIT TIME][13]   | - Changing this cadence will increase or decrease the number of channel subscriptions and notifications sent by an AutopushUser. <br/>- This value can be set in the Locust UI.                                  |
+| [TASK WEIGHT][14] | - Changing this weight impacts the probability of a task being chosen for execution. <br/>- This value is hardcoded in the task decorators of the AutopushUser class.                                            |
+| USERS_PER_WORKER  | - This value should be set to the maximum number of users a Locust worker can support given CPU and memory constraints. <br/>- This value is hardcoded in the AutopushLoadTestShape class.                       |
```
jrconlin, line 221:
```suggestion
| `USERS_PER_WORKER`  | - This value should be set to the maximum number of users a Locust worker can support given CPU and memory constraints. <br/>- This value is hardcoded in the AutopushLoadTestShape class.                       |
```

tests/load/README.md
taddes: fwiw, if nobody told me to do this and if it wasn't documented, I'd have no idea to do it. Given not deleting means a cluster keeps running and it costs us $$, probably doesn't hurt? What do you think?

jrconlin: Oh, I'm not saying that it couldn't be useful, just that it's possibly missing something. 

Maybe it's just me, but I tend to read docs like this in one of two ways: 
1) Thoughtful reflection on a new service.
2) Panicked fury because I need something done NOW before the asteroid hits.

Having just this command doesn't really tell me much other than "This is where the file is".

You could change the `read response` to something like 
```bash
response = ${1:-${COMMAND:-$(read r; echo $r)}}
```
which would let you do things like:

`./tests/load/setup_k8s.sh delete` or `COMMAND=delete ./tests/load/setup_k8s.sh` as well as prompt if no value was provided.

Trinaa: lol to #2, but also so true.

The read response seems like a good idea for an enhancement of the script

tests/load/README.md
Trinaa: I'll look into the read response idea today. I like that.

PR #665, opened 2024-03-22
Chore/update 202403

jrconlin:
**note**: the bump for actix and reqwest appears to have caused a bit of divergence around `StatusCode`. I'm not super happy with the fix, but it works, and we shouldn't get any odd status codes that don't work between versions.

### DIFF
--- a/autoendpoint/Cargo.toml
+++ b/autoendpoint/Cargo.toml
@@ -44,7 +44,7 @@ tokio.workspace = true
 url.workspace = true
 uuid.workspace = true
 
-a2 = { version = "0.8", git = "https://github.com/mozilla-services/a2", branch = "master" }
+a2 = { version = "0.8" }
 bytebuffer = "2.1"
 again = { version = "0.1.2", default-features = false, features = [
     "log",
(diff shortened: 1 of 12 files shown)


autoendpoint/Cargo.toml
```diff
@@ -44,17 +44,17 @@ tokio.workspace = true
 url.workspace = true
 uuid.workspace = true
 
-a2 = { version = "0.8", git = "https://github.com/mozilla-services/a2", branch = "master" }
+a2 = { version = "0.8" }
```
pjenvey, line 47: I guess we finally got all our stuff merged in to a release

autoendpoint/src/routers/mod.rs
```diff
@@ -119,7 +119,7 @@ impl RouterError {
         match self {
             RouterError::Adm(e) => e.status(),
             RouterError::Apns(e) => e.status(),
-            RouterError::Fcm(e) => e.status(),
+            RouterError::Fcm(e) => StatusCode::from_u16(e.status().as_u16()).unwrap_or_default(),
```
pjenvey, line 122: That's annoying, hopefully this will go away at some point (until it maybe happens again) https://github.com/actix/actix-web/pull/3208#issuecomment-2017469785

pjenvey (changes requested):

autoendpoint/src/routers/mod.rs
jrconlin: Yeah, to be honest, though, this is just the latest when it comes to this sort of refactoring. `base64` was the last one to do something like this. I expect some other key library to be the next.

And then I get to deal with protobuf3 fun.

pjenvey (approved):

PR #667, opened 2024-03-23
refactor: upgrade all integration tests to asyncio async await 

taddes:
Closes [SYNC-4120](https://mozilla-hub.atlassian.net/browse/SYNC-4120)

[SYNC-4120]: https://mozilla-hub.atlassian.net/browse/SYNC-4120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taddes:
sentry output test failure only occurring in CI. Investigating why it raises `RuntimeError('Event loop is closed')`

tests/integration/test_integration_all_rust.py
```diff
@@ -584,7 +579,7 @@ def teardown_module():
     kill_process(EP_SERVER)
 
 
-class TestRustWebPush(unittest.TestCase):
+class TestRustWebPush:
     """Test class for Rust Web Push."""
```
jrbenny35, line 583: We can get rid of this class and use pytest fixtures but we can do that later on.

taddes: @jrbenny35  absolutely! That's the next ticket once this is done and am very excited to have fixtures over the class. Thanks for chiming in!

tests/pyproject.toml
```diff
@@ -41,6 +41,9 @@ add-select = ["D212"]
 # D400 First line should end with a period, doesn't work when sentence spans 2 lines
 add-ignore = ["D105","D107","D203", "D205", "D400"]
 
+[tool.pytest.ini_options]
+asyncio_mode = "auto"
```
pjenvey, line 45: It sounds like this mode means we don't need to decorate tests with any `@pytest.mark.asyncio` markers? If so that SGTM: let's get rid of them

tests/integration/test_integration_all_rust.py
```diff
@@ -595,229 +590,259 @@ class TestRustWebPush(unittest.TestCase):
         "sub": "mailto:admin@example.com",
     }
 
+    @staticmethod
+    def clear_sentry_queue():
+        """Clear any values present in Sentry queue."""
+        while not MOCK_SENTRY_QUEUE.empty():
+            MOCK_SENTRY_QUEUE.get_nowait()
+        return
```
pjenvey, line 598:
```suggestion
```

tests/integration/test_integration_all_rust.py
```diff
@@ -595,229 +590,259 @@ class TestRustWebPush(unittest.TestCase):
...
 
     def host_endpoint(self, client):
         """Return host endpoint."""
         parsed = urlparse(list(client.channels.values())[0])
-        return "{}://{}".format(parsed.scheme, parsed.netloc)
+        return f"{parsed.scheme}://{parsed.netloc}"
 
-    @inlineCallbacks
-    def quick_register(self):
+    @pytest.mark.asyncio
```
pjenvey, line 610: regardless of asyncio auto mode or not, this and `shut_down` shouldn't be decorated w/ `@pytest.mark.asyncio` as they're not tests

tests/integration/test_integration_all_rust.py
```diff
@@ -1273,105 +1298,118 @@ def test_big_message(self):
...
 
-    @inlineCallbacks
-    def test_internal_endpoints(self):
+    @pytest.mark.asyncio
+    async def test_ws_ping(self):
+        """Test that the client can send a ping and get expected
+        future that returns the expected float latency value.
+        """
+        client = await self.quick_register()
+        result = await client.ping()
```
pjenvey, line 1402: I wasn't familiar with the [websockets ping API](https://websockets.readthedocs.io/en/stable/reference/asyncio/client.html#websockets.client.WebSocketClientProtocol.ping) so this confused me a bit: could we rename this `result` to `pong_waiter` like their example, and/or note in the method docstring we're awaiting a pong response.

pjenvey (changes requested):

tests/integration/async_push_test_client.py
```diff
@@ -236,7 +237,8 @@ async def send_notification(
         # Pull the sent notification immediately if connected.
         # Calls `get_notification` to get response from websocket.
         if self.ws and self.ws.is_client:  # check back on this after integration
-            return object.__getattribute__(self, "get_notification")(timeout)
+            res = await object.__getattribute__(self, "get_notification")(timeout)
+            return res
```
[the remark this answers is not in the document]
taddes: Good call @pjenvey . Relatedly, was it in convo with you or @jrconlin that it was suggested the `"get_notification"` calls be made explicitly instead of in this 'clever' way, since I know when I first saw it and looked at the resultant return values in the tests that it was a bit baffling at first.

tests/integration/test_integration_all_rust.py
```diff
@@ -595,229 +590,259 @@ class TestRustWebPush(unittest.TestCase):
...
         if os.getenv("SKIP_SENTRY"):
             SkipTest("Skipping sentry test")
             return
         # Ensure bad data doesn't throw errors
-        client = PushTestClient(self._ws_url)
-        yield client.connect()
-        yield client.hello()
-        yield client.send_bad_data()
-        yield self.shut_down(client)
+        TestRustWebPush.clear_sentry_queue()
```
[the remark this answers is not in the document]
taddes: Totally. Will move into the `pytest` setup/teardown fixture in the next PR so it conforms to `pytest` and not this `unittest` artifact.

tests/integration/test_integration_all_rust.py
taddes: Thanks, good catch 👍

tests/pyproject.toml
pjenvey: 👍 I'm fine w/ either way

tests/integration/test_integration_all_rust.py
```diff
@@ -595,229 +590,252 @@ class TestRustWebPush(unittest.TestCase):
...
         assert sorted(values) == ["ERROR:Success", "LogCheck"]
 
+    @pytest.mark.asyncio
     @max_logs(conn=4)
-    def test_no_sentry_output(self):
+    async def test_no_sentry_output(self):
         """Test for no Sentry output."""
+        # NOTE: This test is ordered to run first to avoid possible race conditions that
+        # were observed when the test runs in CI. Use of pytest-order library to
+        # explicitly define order so this test runs first.
```
pjenvey, line 696:
```suggestion
 ```

**Review by pjenvey — CHANGES_REQUESTED** [review]

File: tests/integration/async_push_test_client.py [hunk 9]
```diff
@@ -272,17 +273,27 @@ async def get_broadcast(self, timeout=1):  # pragma: no cover
...
         log.debug("Send: {}")
-        await self.ws.ping("{}")
+        await self.ws.send("{}")
         result = await self.ws.recv()
         log.debug(f"Recv: {result}")
         return result
 
+    async def ping(self):
+        """Test ping/pong request and respose of websocket.
+        A ping may serve as a keepalive or as a check that the remote endpoint received.
```
jrconlin, line 289:
```suggestion
        A ping may serve as a keepalive or as a check that the remote is continuing to respond to websocket protocol messages.
```

tests/integration/async_push_test_client.py
```diff
@@ -272,17 +273,27 @@ async def get_broadcast(self, timeout=1):  # pragma: no cover
...
             raise Exception("WebSocket client not available as expected.")
 
         log.debug("Send: {}")
-        await self.ws.ping("{}")
+        await self.ws.send("{}")
         result = await self.ws.recv()
         log.debug(f"Recv: {result}")
         return result
 
+    async def ping(self):
```
jrconlin, line 287: Not really sure that this is necessary, since it's testing an aspect of the Websocket protocol, not our implementation of Web Push but 🤷

tests/integration/test_integration_all_rust.py
```diff
@@ -595,229 +590,259 @@ class TestRustWebPush(unittest.TestCase):
...
+        client = AsyncPushTestClient(self._ws_url)
+        await client.connect()
+        await client.hello()
+        await client.send_bad_data()
+        await client.disconnect()
 
         # LogCheck does throw an error every time
-        httpx.get(f"http://localhost:{CONNECTION_PORT}/v1/err/crit", timeout=30)
+        async with httpx.AsyncClient() as httpx_client:
+            await httpx_client.get(f"http://localhost:{CONNECTION_PORT}/v1/err/crit", timeout=30)
```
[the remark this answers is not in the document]
jrconlin: I'd prefer to keep the 30s timeout, if only because CI can be a bit flakey or slow and I'm fine if we don't accidentally timeout and fail the test. If  the `client` responds faster, all the better.

tests/integration/test_integration_all_rust.py
taddes: Good catch.

tests/integration/async_push_test_client.py
taddes: Yeah, you are right, don't necessarily need it. I can remove

pjenvey (approved):

jrconlin (approved):

jrconlin:
Thank you for all the work on this. It's awesome.

taddes:
Thanks so much @jrconlin ❤️

pjenvey:
One last thing I forgot to mention: these changes added a few new warnings to the test suite, we should log an issue to take a pass/attempt to quiet them (if possible)

taddes:
Yes @pjenvey , was going to do just that. It's those intermittent socket disconnection warnings which I think are a byproduct of the new library. I'll file the ticket for that and will look at it. Thanks!

PR #668, opened 2024-03-23
refactor: Replace Bottle with FastAPI

taddes:
Closes [SYNC-4188](https://mozilla-hub.atlassian.net/browse/SYNC-4188)

Turns out Docker interface didn't like the IP of `localhost`, yet `127.0.0.1` worked. Maybe an issue in IP resolution since Ipv6 interfaces checked first?

Note: Will need to rebase async changes in https://github.com/mozilla-services/autopush-rs/pull/667 over this before merge. 
- Handles dependency bump raised by dependabot for `black` Closes https://github.com/mozilla-services/autopush-rs/pull/664

[SYNC-4188]: https://mozilla-hub.atlassian.net/browse/SYNC-4188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taddes:
Ugh, clearly an error in the Docker image that does not occur locally, investigating.

jrconlin:
Some docker images do not have any `dnsutils` installed, or the host `/etc/hosts` file doesn't declare `localhost` so they have no way to identify `localhost`. Using an IP address helps. I suppose we could probably try a few 'clever' work around, but it's probably not worth it.

tests/integration/test_integration_all_rust.py
```diff
@@ -1448,9 +1457,35 @@ async def test_broadcast_update_on_connect(self):
         MOCK_MP_SERVICES = {"kinto:123": "ver1"}
         MOCK_MP_POLLED.clear()
         MOCK_MP_POLLED.wait(timeout=5)
+        log.critical("MEGA")
```
jrconlin, line 1460: Why are you logging these as `critical`?

taddes: For the bold red to make it easy to find the debug messages. All those log and err messages are removed now.  Turned out the issue was not with the tests, but autopush throwing an error when the `v1/broadcasts` did not return a response in the time it was setup.

jrconlin: Ah, heh. Sorry. 

FWIW, Lina introduced me to the idea of using leading emoji for debugging lines. Turns out many folk process the colors and shapes faster than text allowing you to quickly skim complex logs for distinct patterns. (It works best for displays that can support emoji fonts, so VT100 style will probably not cut it.)

taddes: Great idea, thx for the suggestion, I have seen it in the repo and it's nice.

tests/integration/dual_test.json
```diff
@@ -7,4 +7,4 @@
     "db_settings": "{\"message_table\": \"message_int_test\",\"router_table\": \"router_int_test\"}",
     "dsn": "http://localhost:8000/"
   }
-}
+}
```
jrconlin, line 10: Was removing this last \n intentional?

jrconlin (approved):

tests/integration/dual_test.json
taddes: My auto format for some reason did so! Probably an IDE "feature". Is there a preference one way or another? Happy to change it if we prefer a trailing line.

jrconlin: no worries. It's a JSON file so not super critical. Just curious if there was some different reason.

PR #670, opened 2024-03-27
feat: Switch to randomized UAID for health check

jrconlin:
Closes: SYNC-4203

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -11,7 +11,7 @@ use cadence::{CountedExt, StatsdClient};
 use futures_util::StreamExt;
 use google_cloud_rust_raw::bigtable::admin::v2::bigtable_table_admin::DropRowRangeRequest;
 use google_cloud_rust_raw::bigtable::admin::v2::bigtable_table_admin_grpc::BigtableTableAdminClient;
-use google_cloud_rust_raw::bigtable::v2::bigtable::{PingAndWarmRequest, ReadRowsRequest};
+use google_cloud_rust_raw::bigtable::v2::bigtable::ReadRowsRequest;
 use google_cloud_rust_raw::bigtable::v2::bigtable_grpc::BigtableClient;
 use google_cloud_rust_raw::bigtable::v2::data::{RowFilter, RowFilter_Chain};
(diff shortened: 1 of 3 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -782,24 +782,29 @@ impl BigtableDb {
...
+                        .read_rows_opt(&req, call_opts(self.health_metadata.clone()))
                 },
                 retryable_error(metrics.clone()),
             )
             .await
             .map_err(|e| DbError::General(format!("BigTable connectivity error: {:?}", e)))?;
 
-        // Since this should return no rows (with the row key set to a value that shouldn't exist)
-        // the first component of the tuple should be None.
+        // Since this may return no rows, the first component of the tuple should be None.
```
pjenvey, line 807: This is an old comment about previous usage of `into_future` here returning a tuple, which we're not doing anymore

```suggestion
```

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -782,24 +782,29 @@ impl BigtableDb {
     /// the tablename.
```
pjenvey, line 782: There's an old comment about `instance_name` here which is no longer relevant

autopush-common/src/db/bigtable/mod.rs
```diff
@@ -115,12 +115,9 @@ impl BigTableDbSettings {
             .map_err(BigTableError::GRPC)
     }
 
+    // Health may require a different metadata declaration.
     pub fn health_metadata(&self) -> Result<Metadata, BigTableError> {
```
pjenvey, line 119: Assuming this new health check works out I don't see the need to distinguish the exact same health vs regular metadata but I won't block on this.

pjenvey (changes requested):

autopush-common/src/db/bigtable/mod.rs
jrconlin: Honestly, I figured I'd leave it in case there's something else we may need to do that's specific to the health check.  (e.g. we need to specify some additional header or some alternate tag because we make the call so frequently or whatever.)

pjenvey: YAGNI IMO

pjenvey (approved):

PR #679, opened 2024-04-15
feat: Ensure that status and labels are propagated for DB errors

jrconlin:
Closes: SYNC-4233


autoendpoint/src/error.rs
```diff
@@ -195,9 +194,9 @@ impl ApiErrorKind {
             ApiErrorKind::General(_) => "general",
             ApiErrorKind::Io(_) => "io",
             ApiErrorKind::Metrics(_) => "metrics",
-            ApiErrorKind::Database(_) => "database",
+            ApiErrorKind::Database(e) => e.metric_label().unwrap_or("database"),
```
pjenvey, line 197: Should probably just return on these?
```suggestion
            ApiErrorKind::Database(e) => return e.metric_label(),
```

jrconlin: I'll also fix up the other one too.

pjenvey (approved):

PR #682, opened 2024-04-19
chore: update per RUSTSEC-2024-0336

pjenvey:
also ignore it because our rustls 0.19/0.20 is only used by tokio-rustls (not affected per the RUSTSEC) and http client lib usage (hyper-alpn and hyper-rustls 0.22)


pjenvey:
Note the RUSTSEC currently describes `rustls-tokio` as not affected but [they meant](https://github.com/rustsec/advisory-db/pull/1946#issuecomment-2067036800) `tokio-rustls`

jrconlin (approved):

taddes (approved):
Thanks @pjenvey , will resolve the `cargo audit` failure in https://app.circleci.com/pipelines/github/mozilla-services/autopush-rs/2462/workflows/a2983cd3-06f2-4886-a6a6-8ecaeadc3465

PR #694, opened 2024-04-25
feat: fixturize pytest

taddes:
Closes [SYNC-4116](https://mozilla-hub.atlassian.net/browse/SYNC-4116)

Summary & Acceptance Criteria Met:
- `unittest` fully removed.
- All tests pulled out of classes and into individual test classes, all references to duplicated class boilerplate in normal tests and broadcast tests moved into their own fixtures.
- Some setup and teardown logic moved into autouse fixtures.
- Test data is injected into individual tests through `pytest.mark.parameterize`. This removes internal test setup and also provides the injected values in the test output, for easier debugging.
- `max_logs` decorator removed and `process_logs` are now implemented though an autouse fixture that runs at the end of each test, allowing the endpoint and connection values to be passed in.
- 🎉 Bonus: resolved the `[   ERROR] Task was destroyed but it is pending!` problem by moving the `test_client` and `registered_test_client` into async fixtures that `yield` the client and then automatically disconnect after each test function.
- 🎉 Bonus: tests run faster! Cut the execution time in half: `65.22s` -> `36.29s`

Note: 
- potential move into conftest would be covered in scope of another task. May not get much benefit from this presently. Also, moving scope of some of the setup would be too much in a single PR given the changes in design here.
- Given UUID data is created as params passed to each test function, I opted for this over a fixture since we assume the push connections are frequently receiving unique UAIDs. Plus annotated above each test function makes it easier to see the input, as there are several test cases that involve a bit of variation. 

[SYNC-4116]: https://mozilla-hub.atlassian.net/browse/SYNC-4116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/test_integration_all_rust.py
+++ b/tests/integration/test_integration_all_rust.py
@@ -788,9 +788,11 @@ class TestRustWebPush:
         assert result["use_webpush"] is True
         await test_client.disconnect()
 
-    async def test_hello_with_bad_prior_uaid(self, test_client):
+    @pytest.mark.parametrize("non_uaid", [uuid.uuid4().hex])
+    async def test_hello_with_bad_prior_uaid(
+        self, test_client: AsyncPushTestClient, non_uaid: str
+    ):
         """Test hello with bad prior uaid."""
-        non_uaid = uuid.uuid4().hex
         await test_client.connect()
         result = await test_client.hello(uaid=non_uaid)
         assert result != {}
@@ -798,33 +800,40 @@ class TestRustWebPush:
         assert result["use_webpush"] is True
         await test_client.disconnect()
 
-    async def test_basic_delivery(self, registered_test_client):
+    @pytest.mark.parametrize("uuid_data", [str(uuid.uuid4())])
+    async def test_basic_delivery(
+        self, registered_test_client: AsyncPushTestClient, uuid_data: str
+    ):
         """Test basic regular push message delivery."""
-        data = str(uuid.uuid4())
-        result = await registered_test_client.send_notification(data=data)
+        result = await registered_test_client.send_notification(data=uuid_data)
         # the following presumes that only `salt` is padded.
         clean_header = registered_test_client._crypto_key.replace('"', "").rstrip("=")
         assert result["headers"]["encryption"] == clean_header
-        assert result["data"] == base64url_encode(bytes(data, "utf-8"))
+        assert result["data"] == base64url_encode(bytes(uuid_data, "utf-8"))
         assert result["messageType"] == "notification"
 
-    async def test_topic_basic_delivery(self, registered_test_client):
+    @pytest.mark.parametrize("uuid_data", [str(uuid.uuid4())])
+    async def test_topic_basic_delivery(
+        self, registered_test_client: AsyncPushTestClient, uuid_data: str
+    ):
         """Test basic topic push message delivery."""
-        data = str(uuid.uuid4())
-        result = await registered_test_client.send_notification(data=data, topic="Inbox")
+        result = await registered_test_client.send_notification(data=uuid_data, topic="Inbox")
         # the following presumes that only `salt` is padded.
         clean_header = registered_test_client._crypto_key.replace('"', "").rstrip("=")
         assert result["headers"]["encryption"] == clean_header
-        assert result["data"] == base64url_encode(data)
+        assert result["data"] == base64url_encode(uuid_data)
         assert result["messageType"] == "notification"
 
-    async def test_topic_replacement_delivery(self, registered_test_client):
+    @pytest.mark.parametrize(
+        ["uuid_data_1", "uuid_data_2"], [(str(uuid.uuid4()), str(uuid.uuid4()))]
+    )
+    async def test_topic_replacement_delivery(
+        self, registered_test_client: AsyncPushTestClient, uuid_data_1: str, uuid_data_2: str
+    ):
         """Test that a topic push message replaces it's prior version."""
-        data = str(uuid.uuid4())
-        data2 = str(uuid.uuid4())
         await registered_test_client.disconnect()
-        await registered_test_client.send_notification(data=data, topic="Inbox", status=201)
-        await registered_test_client.send_notification(data=data2, topic="Inbox", status=201)
+        await registered_test_client.send_notification(data=uuid_data_1, topic="Inbox", status=201)
+        await registered_test_client.send_notification(data=uuid_data_2, topic="Inbox", status=201)
         await registered_test_client.connect()
         await registered_test_client.hello()
         result = await registered_test_client.get_notification()
@@ -832,7 +841,7 @@ class TestRustWebPush:
         # the following presumes that only `salt` is padded.
         clean_header = registered_test_client._crypto_key.replace('"', "").rstrip("=")
         assert result["headers"]["encryption"] == clean_header
-        assert result["data"] == base64url_encode(data2)
+        assert result["data"] == base64url_encode(uuid_data_2)
         assert result["messageType"] == "notification"
         result = await registered_test_client.get_notification()
         assert result is None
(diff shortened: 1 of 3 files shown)


tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
+    process_logs_autouse(max_conn_logs=4)
+
+
+async def test_hello_echo(test_client) -> None:
+    """Test hello echo."""
+    await test_client.connect()
+    result = await test_client.hello()
+    assert result != {}
+    assert result["use_webpush"] is True
+    await test_client.disconnect()
```
Trinaa, line 750: The `test_client` fixture calls `disconnect`, do we also need to do it here?

Question also applies to: 
- test_hello_with_bad_prior_uaid

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
+            data = MOCK_SENTRY_QUEUE.get(timeout=5)
+            assert not data
+    except httpx.ConnectError:
+        pass
+    except Empty:
+        pass
+    process_logs_autouse(max_conn_logs=4)
+
+
+async def test_hello_echo(test_client) -> None:
```
Trinaa, line 744: nit:
```suggestion
async def test_hello_echo(test_client: AsyncPushTestClient) -> None:
```

tests/integration/test_integration_all_rust.py
```diff
@@ -536,22 +524,28 @@ def setup_endpoint_server():
...
-    """Set up module including BigTable or Dynamo
-    and connection, endpoint and megaphone servers.
+@pytest.fixture(autouse=True, scope="session")
+def setup_teardown() -> Generator:
+    """Set up and tear down test resources within module.
+
+    Setup: BigTable or Dynamo and connection, endpoint, mock and megaphone servers.
+
+    Using pytest fixtures for setup/teardown means anything prior to `yield` statement
+    executes at test creation and anything after `yield` is run after tests complete.
```
Trinaa, line 534: nit: I think this comment repeats what is well documented by pytest and can be removed

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
+    await test_client.connect()
+    result = await test_client.hello()
+    assert result != {}
+    assert result["use_webpush"] is True
+    await test_client.disconnect()
+
+
+@pytest.mark.parametrize("non_uaid", [uuid.uuid4().hex])
+async def test_hello_with_bad_prior_uaid(test_client: AsyncPushTestClient, non_uaid: str) -> None:
+    """Test hello with bad prior uaid."""
```
Trinaa, line 755: I agree with your point "annotated above each test function makes it easier to see the input", but using parameterize for a single value causes verbosity in the test method signature and noise/complexity in the test reporting that's unneeded. I suggest instead to declare the input at the top of the test instead, like so

```suggestion
async def test_hello_with_bad_prior_uaid(test_client: AsyncPushTestClient) -> None:
    """Test hello with bad prior uaid."""
    non_uaid = uuid.uuid4().hex
```
\* _This suggestion applies to all tests with singular parameterize values_

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
+    """Return websocket url for megaphone broadcast testing."""
+    return f"ws://localhost:{MP_CONNECTION_PORT}/"
+
 
-    # Max log lines allowed to be emitted by each node type
-    max_endpoint_logs = 8
-    max_conn_logs = 3
-    vapid_payload = {
+@pytest.fixture(name="vapid_payload", scope="session")
+def fixture_vapid_payload() -> dict[str, int | str]:
```
Trinaa, line 599: Maybe define the return value as a type?

```suggestion
VapidPayload = dict[str, int | str]

@pytest.fixture(name="vapid_payload", scope="session")
def fixture_vapid_payload() -> VapidPayload:
```

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
-            SkipTest("Skipping sentry test")
-            return
-        # Ensure bad data doesn't throw errors
-        client = AsyncPushTestClient(self._ws_url)
-        await client.connect()
-        await client.hello()
-        await client.send_bad_data()
+
+@pytest.fixture(name="clear_sentry_queue", autouse=True, scope="function")
+def fixture_clear_sentry_queue():
```
Trinaa, line 608: nit
```suggestion
def fixture_clear_sentry_queue() -> None:
```

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
+    """Clear any values present in Sentry queue.
+    Scoped to run automatically after each test function.
+    """
+    yield
+    while not MOCK_SENTRY_QUEUE.empty():
+        MOCK_SENTRY_QUEUE.get_nowait()
+
+
+@pytest.fixture(name="ws_config", scope="session")
+def fixture_ws_config():
```
Trinaa, line 618: nit:
```suggestion
def fixture_ws_config() -> dict[str, int]:
```

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
+    """Return collection of configuration values."""
+    return {"connection_port": 9150}
+
+
+@pytest.fixture(name="test_client", scope="function")
+async def fixture_test_client(ws_url):
```
Trinaa, line 624: nit:

```suggestion
async def fixture_test_client(ws_url) -> AsyncPushTestClient:
```

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
-        """Test sentry output for autoendpoint."""
-        if os.getenv("SKIP_SENTRY"):
-            SkipTest("Skipping sentry test")
-            return
-        client = await self.quick_register()
-        endpoint = self.host_endpoint(client)
-        await self.shut_down(client)
+
+@pytest.fixture(name="test_client_broadcast", scope="function")
+async def fixture_test_client_broadcast(broadcast_ws_url):
```
Trinaa, line 636: nit:

```suggestion
async def fixture_test_client_broadcast(broadcast_ws_url) -> AsyncPushTestClient:
```

tests/integration/test_integration_all_rust.py
```diff
@@ -590,983 +583,1003 @@ def teardown_module():
...
+    """Return a push test client for use within tests that can have
+    a custom websocket url passed in to initialize.
+    """
+    client = AsyncPushTestClient(broadcast_ws_url)
+    yield client
+    await client.disconnect()
+
+
+@pytest.fixture(name="registered_test_client", scope="function")
+async def fixture_registered_test_client(ws_config):
```
Trinaa, line 646: nit:

```suggestion
async def fixture_registered_test_client(ws_config) -> AsyncPushTestClient:
```

Trinaa (commented):
@taddes I'm really enjoying seeing all the improvements you've made 😸 !

- I agree with your note on conftest and think the right moment to add that in would coincide with splitting this module

- I added some nits on fixture typing, but would actually recommend you're doing a once over and adding them consistently in the test signatures. I suspect this may enable the IDE to highlight some mismatches for us to further improve upon, although if it creates too much blockage in linting I understand if it's omitted for the moment

- Before approving I have some questions for the `fixture_process_logs_autouse` (see inline)

tests/integration/test_integration_all_rust.py
taddes: This actually returns an `AsyncGenerator` because of the use of `yield`, which is not very helpful from a typing context, which is why I left it out, but no harm in adding it if you like though.

tests/integration/test_integration_all_rust.py
taddes: As below, `AsyncGenerator`

tests/integration/test_integration_all_rust.py
taddes: As below, `AsyncGenerator`

tests/integration/test_integration_all_rust.py
taddes: Due to use of `yield`, this is actually a `Generator`, which is fine to know but not super helpful... thanks mypy haha

tests/integration/test_integration_all_rust.py
taddes: Good catch, there are some explicit `disconnect` calls, but these are not actually required in this case since it will auto disconnect.

tests/integration/test_integration_all_rust.py
taddes: Sure!

jrconlin (approved):

tests/integration/test_integration_all_rust.py
```diff
@@ -243,52 +240,40 @@ def print_lines_in_queues(queues, prefix):
                 sys.stdout.write(prefix + line)
 
 
-def process_logs(testcase):
-    """Process (print) the testcase logs (in tearDown).
+@pytest.fixture(name="process_logs_autouse", autouse=True, scope="function")
+def fixture_process_logs_autouse():
```
pjenvey, line 244: Does this need to be a fixture?

I'll argue the decorator version of this is nicer since it didn't require the test to call `process_logs` themselves. Granted the previous decorator accomplished that via `process_logs` in the `tearDown` hook which doesn't exist any longer, but I think if this was still a decorator it could handle that equivalent of `process_logs` by itself?

If so I think that further suggests it shouldn't be a fixture as tests wouldn't require any parameter/object provided from it (such as `process_logs_autouse`).

taddes:
Sorry @Trinaa ... additional push undid approve... grr Git

Trinaa (approved):

tests/integration/test_integration_all_rust.py
taddes: Thanks again @pjenvey , as we discussed offline it's now using the parametrized fixtures that pass in the context much the same way with the decorator, however avoiding the async funkiness... 

Two named fixtures that conditionally return an int repr of that given log value:
```python
@pytest.fixture
def fixture_max_conn_logs(request):
    """Fixture returning altered max_conn_logs value for process_logs fixture."""
    if hasattr(request, "param"):
        return int(request.param)


@pytest.fixture
def fixture_max_endpoint_logs(request):
    """Fixture returning altered max_endpoint_logs value for process_logs fixture."""
    if hasattr(request, "param"):
        return int(request.param)
```
        
Single param invocation:

```python
@pytest.mark.parametrize("fixture_max_conn_logs", [4], indirect=True)
```

Multi param invocation
```python
@pytest.mark.parametrize(
    "fixture_max_endpoint_logs, fixture_max_conn_logs",
    [[4, 1]],
    indirect=["fixture_max_endpoint_logs", "fixture_max_conn_logs"],
)
```

PR #695, opened 2024-04-25
feat: Reduce some of the more common errors from the logs

jrconlin:
Reduces some of the less useful errors from the stdout logs.

* Fixes the Bad Aud to be shorter, clearer

Closes SYNC-4243


autoendpoint/src/extractors/subscription.rs
```diff
@@ -84,7 +84,9 @@ impl FromRequest for Subscription {
                 .fernet
                 .decrypt(&repad_base64(&token_info.token))
                 .map_err(|e| {
-                    error!("🔐 fernet: {:?}", e);
+                    // Since we're decrypting and endpoint, we get a lot of spam links.
+                    // This can fill our logs.
+                    trace!("🔐 fernet: {:?}", e);
```
taddes, line 89: Question: doesn't `trace!` capture more detailed logging, traceback info? I'm just going off the comment in that the intention seems to reduce what is filled in the log, or is that not the intention. I may likely be missing something!

taddes (approved):
👍 Just had one question re: `error!` to `trace!1`, but not blocking! Thanks @jrconlin !

autoendpoint/src/extractors/subscription.rs
taddes: Cool, thanks!

PR #705, opened 2024-05-14
 feat: Add daily TTL refresh for channels

jrconlin:
* Clippy updates
* Defaults to single updates per day.

This does introduce a new field `refreshed_at` which contains the UTC second of the last time a given router record was refreshed. It's tied to the Bigtable configuration option `refresh_period_seconds`, and defaults to 86400. Router updates are only performed once per `refresh_period_seconds`. Route refreshment happens in Bigtable's `update_user()`.

Closes: SYNC-4221


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -849,6 +906,11 @@ impl DbClient for BigTableClientImpl {
             ));
         };
 
+        if self.refresh_router(&user.uaid).await? {
```
jrconlin, line 909: I'm also fine with making the parameter `&mut user` and changing the `refreshed_at` value, but I think this is a bit clearer about what is happening.

jrconlin:
OBE

PR #708, opened 2024-05-20
refactor: autopush-common remove legacy autopush

taddes:
Closes [SYNC-4261](https://mozilla-hub.atlassian.net/browse/SYNC-4261)

This will be somewhat more trivial in scope as [SYNC-3451](https://mozilla-hub.atlassian.net/browse/SYNC-3451) will remove the dependencies and code related to some of these dependencies (usoto_core,  rusoto_dynamodb, and serde_dynamodb)

[SYNC-4261]: https://mozilla-hub.atlassian.net/browse/SYNC-4261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SYNC-3451]: https://mozilla-hub.atlassian.net/browse/SYNC-3451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrconlin (dismissed):

pjenvey (changes requested):
There's a number of `ApcErrorKind` variants that are no longer used (solely by legacy autopush) that could be removed.

I'm not sure if there's an easy way to detect these other than removing all of them then checking where rustc complains

taddes:
Great catch @pjenvey , I removed those referenced in `autopush` by cross referencing the old crate and did some compiler checks to make sure there were no issues. There was an unused `Thread(Box<dyn Any + Send>)` error that wasn't called anywhere that I took out, too.

pjenvey:
@taddes I'm seeing even more unused variants!

Sorry to persist on this but this is a good time to remove all the unnecessary ones (and if enough are removed we might even consider removing ApcError itself).

I've spotted `PongTimeout` and `ExcessivePing` as being from legacy autopush (the new autoconnect has their own versions of these), so there might be others as well.

pjenvey (changes requested):
Still seeing a couple here that could be removed such as the `PongTimeout` and `ExcessivePing` I mentioned

taddes:
@pjenvey I think that accounts for all of the unused variants. BroadcastError, GeneralError, ConfigError, & PayloadError  are all used externally directly, with the others covering errors using the `From` trait, so I left them be. Let me know if that seems sufficient. Thanks!

pjenvey (changes requested):
I think I'm also seeing the `Json` and `Httparse` variants unused, can you double check those?

taddes:
Hm, might those need to remain for the status method?

taddes:
Alright, got em @pjenvey !

pjenvey (approved):

jrconlin:
🎉

PR #709, opened 2024-05-30
fix: jsonwebtoken validation errors

taddes:
Explanation of why this error condition occurred, from looking at the `jsonwebtoken` changelog and source:

v 9.0.0 - _Rejects JWTs containing audiences when the Validation doesn't contain any._ 
- Note: therefore, required to set in`set_required_spec_claims` [source](https://github.com/Keats/jsonwebtoken/blob/afbb44e34dbd721c35dedb68ec13a86ff6db6d8b/src/validation.rs#L152) and explicit `aud` value needs to be set via `set_audience` [source](https://github.com/Keats/jsonwebtoken/blob/afbb44e34dbd721c35dedb68ec13a86ff6db6d8b/src/validation.rs#L137)

v 9.2.0 - _Add an option to not validate aud in the Validation struct_ [source](https://github.com/Keats/jsonwebtoken/blob/afbb44e34dbd721c35dedb68ec13a86ff6db6d8b/src/validation.rs#L74). 
- Note: this could optionally emulate the previous state by setting `validate_aud` to `false`.

From the [docs](https://crates.io/crates/jsonwebtoken/9.3.0): 
_Validation: This library automatically validates the exp claim, and nbf is validated if present. You can also validate the sub, iss, and aud but those require setting the expected values in the Validation struct. **In the case of aud, if there is a value set in the token but not in the Validation, the token will be rejected.**_

After adding the explicit `aud` value and setting the required spec claims, this left a lingering issue with the integration tests where `test_basic_delivery_with_vapid` failed due to a `401 Unauthorized` response during the `send_notification` call.

Solution: setting explicit audience value' only caveat being we need to set the value in the test env as well that is not the push service endpoint:
Ex. validation.set_audience(&["https://push.services.mozilla.org", "http://127.0.0.1:9160"]);
It's probably better then to set the audience value within a configuration variable that can be altered based on context.

Alternatively, `validation.validate_aud = false` would bypass explicit validation of aud, though not sure we'd want that. 

Closes [SYNC-2401](https://mozilla-hub.atlassian.net/browse/SYNC-4201)

[SYNC-2401]: https://mozilla-hub.atlassian.net/browse/SYNC-2401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -28,6 +28,8 @@ use crate::headers::{
 use crate::metrics::Metrics;
 use crate::server::AppState;
 
+use crate::settings::Settings;
+
 const ONE_DAY_IN_SECONDS: u64 = 60 * 60 * 24;
 
 /// Extracts subscription data from `TokenInfo` and verifies auth/crypto headers
@@ -262,11 +264,13 @@ fn validate_vapid_jwt(
     domain: &Url,
     metrics: &Metrics,
 ) -> ApiResult<()> {
+    let settings = Settings::with_env_and_config_file(&None).unwrap();
     let VapidHeaderWithKey { vapid, public_key } = vapid;
 
     let public_key = decode_public_key(public_key)?;
     let mut validation = Validation::new(Algorithm::ES256);
-    validation.set_audience(&["https://push.services.mozilla.org", "http://127.0.0.1:9160"]);
+    let audience: Vec<&str> = settings.vapid_aud.iter().map(|s| s.as_str()).collect();
+    validation.set_audience(&audience);
     validation.set_required_spec_claims(&["exp", "aud", "sub"]);
 
     let token_data = match jsonwebtoken::decode::<VapidClaims>(
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -266,7 +266,7 @@ fn validate_vapid_jwt(
 
     let public_key = decode_public_key(public_key)?;
     let mut validation = Validation::new(Algorithm::ES256);
-    validation.set_audience(&["https://push.services.mozilla.org"]);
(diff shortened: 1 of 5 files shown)


autoendpoint/src/extractors/subscription.rs
```diff
@@ -265,10 +265,14 @@ fn validate_vapid_jwt(
     let VapidHeaderWithKey { vapid, public_key } = vapid;
 
     let public_key = decode_public_key(public_key)?;
+    let mut validation = Validation::new(Algorithm::ES256);
+    validation.set_audience(&["https://push.services.mozilla.org", "http://127.0.0.1:9160"]);
```
jrconlin, line 269: I think these should be passed in as configuration options. I really prefer if we don't hard code things (particularly the test URL.)

jrconlin (changes requested):

autoendpoint/src/extractors/subscription.rs
taddes: Updated the `Settings` struct to hold the vapid_aud value as a `vec`, then pull it in to pass to the `set_audience` method. If you prefer another method with env vars and whatnot, can do that. Do we use the [`cloudops-infra` repo for `autopush`](https://github.com/mozilla-services/cloudops-infra/tree/master/projects/autopush) for env var storage and reading, btw?

jrconlin: No, that should be fine. I've tested locally and things are working fine either if it's specified in a config file 

```toml
vapid_aud = ["https://push.services.mozilla.org", "http://127.0.0.1:8001"]
```

or as an environment variable
```bash
AUTOENDPOINT_VAPID_AUD='["https://push.services.mozilla.org", "http://127.0.0.1:8001"]'
```
(Older versions did not handle parsing the environment string correctly. Looks like that's no longer a problem.)

jrconlin (approved):

PR #711, opened 2024-06-04
 chore: Update to debian bookworm 

jrconlin:
* This updates the build baseline to Debian Bookworm and Rust 1.78

Closes: SYNC-4290


.circleci/config.yml
```diff
@@ -124,14 +124,14 @@ jobs:
...
-            apt install build-essential curl libstdc++6 libstdc++-10-dev libssl-dev pkg-config -y
+            apt install build-essential curl libstdc++6 libstdc++-12-dev libssl-dev pkg-config -y
             apt install cmake -y
             # RUST_VER
-            curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.76 -y
+            curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.78 -y
             export PATH=$PATH:$HOME/.cargo/bin
             echo 'export PATH=$PATH:$HOME/.cargo/bin' >> $BASH_ENV
             rustc --version
-            cargo build --features=emulator --features=dual
```
taddes, line 134: 👍 🥂

autoendpoint/src/routers/apns/router.rs
```diff
@@ -266,31 +266,31 @@ impl ApnsRouter {
         // still needed or used. (I want to get rid of this.)
         if let Some(v) = replacement.get("title") {
             if let Some(v) = v.as_str() {
-                holder.title = v.to_owned();
+                v.clone_into(&mut holder.title);
```
taddes, line 269: Interesting, is this a new clippy and Rust 1.78? I'm not familiar with the pattern. Is it a preferred way of doing things for this context? Just so I understand:

`holder.title = v.to_owned()` clones the content of v into `holder.title`, transferring ownership of the cloned string leaving v intact.

`v.clone_into(& mut holder.title)` : this appears to require a mutable reference, unlike the previous method, so given it's passed into `clone_into`, does it save an allocation by not having to transfer ownership of the `String`? Is that the perf/memory improvement in using this method?

autopush-common/src/db/bigtable/bigtable_client/merge.rs
```diff
@@ -191,7 +191,10 @@ impl RowMerger {
 
         let cell = &mut row.cell_in_progress;
         if chunk.has_family_name() {
-            cell.family = chunk.take_family_name().get_value().to_owned();
+            chunk
+                .take_family_name()
+                .get_value()
+                .clone_into(&mut cell.family);
```
taddes, line 197: Same as above? Reduces ownership transfers by allocating a mutable ref inside the `clone_into` call?

taddes (approved):
Great @jrconlin , thanks for doing this! Approved, but had a general question re the ownership/memory diff in using `clone_into` and `clone_from`.

autoendpoint/src/routers/apns/router.rs
jrconlin: It is indeed a new pattern enforced by clippy under 1.78. Apparently it is more efficient since it doesn't require an intermediary object.

pjenvey: @taddes Note the mutable reference passed to `clone_into` is the destination of the clone, `&mut holder.title` -- `holder.title` must be mutable for the assignment `holder.title = v.to_owned()` as well.

I'll add that `clone_from` is "potentially" more efficient. `Clone` provides a [default implementation of it](https://doc.rust-lang.org/src/core/clone.rs.html#169) that is the same as `clone`, but types can override it to make their own more efficient versions.

Here's a really clear example of one overridden version I found for [<Box[T], A>](https://doc.rust-lang.org/src/alloc/boxed.rs.html#2091): if the destination is the same size, just copy the contents to the already allocated array, otherwise fallback to [allocating a new one](https://doc.rust-lang.org/src/alloc/boxed.rs.html#2087).

taddes: Thanks for the great context @pjenvey & @jrconlin .

PR #712, opened 2024-06-05
refactor: remove User last_connect

taddes:
Closes [SYNC-4248](https://mozilla-hub.atlassian.net/browse/SYNC-4248)

Note a lot of the existing DynamoDB stuff will be removed in upcoming PR, so for now just removed anything related to this and not the surrounding context.

[SYNC-4248]: https://mozilla-hub.atlassian.net/browse/SYNC-4248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/src/architecture.md
```diff
@@ -128,7 +128,6 @@ router record for a `UAID`.
 | router_type  | Router Type (See [`autoendpoint::extractors::routers::RouterType`])              |
 | node_id      | Hostname of the connection node the client is connected to.                      |
 | connected_at | Precise time (in milliseconds) the client connected to the node.                 |
-| last_connect | **global secondary index** - year-month-hour that the client has last connected. |
```
jrconlin, line 131: Since this is already legacy, I wonder if we should keep it? 

The page already discusses things like `curmonth`, which is also a DDB thing back when we did table rotations, so keeping the `last_connect` is about as equally relevant and historic. 

If we just wanted to clean things up, we'd drop all the legacy DDB documentation as well and just keep stuff that reflects the current state.

docs/src/table_rotation.md
```diff
@@ -45,10 +45,9 @@ has read through.
 
 When table rotation is allowed, the router table uses the `curmonth`
 field to indicate the last month the client has read notifications
-through. This is independent of the last_connect since it is possible
```
jrconlin, line 48: Again I can see us keeping this (and maybe the verbs to past tense) since it refers to table rotation, which we stopped doing even for DDB.

jrconlin (dismissed):

docs/src/table_rotation.md
taddes: You know, I had the same thought when I was taking it out. Given it's clearly labeled legacy, we can let it stay.

taddes:
Sry @jrconlin , need to re-approve after adding those blubs back.

pjenvey (approved):

taddes:
Thx @pjenvey

PR #713, opened 2024-06-11
bug: Check to see if the user still exists before registering channels

jrconlin:
there are [reports](https://bugzilla.mozilla.org/show_bug.cgi?id=1486010#c9) of users registering a channel and the channel
immediately returning a 410. This does a simple check to see if the user
record still exists before registering. A user record may have been
removed for some reason while the user connection still persists and
this should catch for that.

Closes: [SYNC-4294](https://mozilla-hub.atlassian.net/browse/SYNC-4294)

[SYNC-4294]: https://mozilla-hub.atlassian.net/browse/SYNC-4294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrconlin:
Thinking about this.
I'm not exactly sure how a device could get into this state. We auto-prune when the bridge rejects a connection, but there may be pre-connected UAs that haven't gotten the message about that. 
For now, instead of creating a potentially invalid endpoint, let's do a check to see if the user record is present before we build the endpoint. (Fortunately, with Bigtable, we're not paying per read, but this might be worth the cost in any case.)

We *might* want to use something like a conditional write instead?

pjenvey:
The [original issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1486010) gives the impression that its root cause is a client side issue, especially with it originally being reported on Desktop which always generates its own `channel_id` (passing it to the server) when registering a new subscription.

So I think we should close this (or at least set as draft) until some digging is done to rule out client side issues being the root cause.

jrconlin:
Yeah, I'm fine making this a draft for now. I can see this being a bit of a problem, but hopefully, it's a fairly rare edge case.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -862,6 +862,15 @@ impl DbClient for BigTableClientImpl {
         Ok(predicate_matched)
     }
 
+    async fn check_user(&self, uaid: &Uuid) -> DbResult<bool> {
```
pjenvey, line 865: if this feature is ever needed: I think it would be better to move the check into `add_channels` itself. Its write would become a `check_and_mutate` depending on the existence of the row key (the same check used by `add_user` w/ a `true`  instead of `false` filter)

PR #715, opened 2024-06-20
refactor : remove dual mode

taddes:
Closes [SYNC-4296](https://mozilla-hub.atlassian.net/browse/SYNC-4296)

[SYNC-4296]: https://mozilla-hub.atlassian.net/browse/SYNC-4296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


autoconnect/autoconnect-settings/src/lib.rs
```diff
@@ -130,7 +130,7 @@ impl Default for Settings {
             // Matches the legacy value
             statsd_label: "autopush".to_owned(),
             statsd_port: 8125,
-            db_dsn: None,
+            db_dsn: Some("grpc://localhost:8086".to_string()),
```
pjenvey, line 133: I think this should continue to be `None` (both here and autoendpoint) as I'd rather a misconfigured autopush fail fast with an error rather than attempting to use this url.

Instead I think `test_settings` should set this `db_dsn` value itself. I'm not sure autoendpoint has any equivalent or needs this, anyway

pjenvey (changes requested):

autoconnect/autoconnect-settings/src/lib.rs
taddes: Good point. Updated 👍

jrconlin (approved):

pjenvey (approved):

PR #720, opened 2024-07-01
refactor: remove dynamodb legacy db impl

taddes:
Closes [SYNC-3451](https://mozilla-hub.atlassian.net/browse/SYNC-3451)

[SYNC-3451]: https://mozilla-hub.atlassian.net/browse/SYNC-3451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
```diff
@@ -181,9 +181,8 @@ impl WebPushClient {
             .or(self.current_timestamp);
         trace!("🗄️ WebPushClient::do_check_storage {:?}", &timestamp);
         // if we're to include topic messages, do those first.
-        // NOTE: DynamoDB would also wind up fetching the `current_timestamp` when pulling these,
-        // Bigtable can't fetch `current_timestamp` so we can't rely on `fetch_topic_messages()`
-        // returning a reasonable timestamp.
+        // NOTE: Bigtable can't fetch `current_timestamp`, so we can't rely on
+        // r `fetch_topic_messages()` eturning a reasonable timestamp.
```
jrconlin, line 185: typo?
```suggestion
        // on `fetch_topic_messages()` returning a reasonable timestamp.
```

autopush-common/build.rs
```diff
@@ -1,5 +1,5 @@
 pub fn main() {
-    if !(cfg!(feature = "dynamodb") || cfg!(feature = "bigtable")) {
-        panic!("No database defined! Please compile with either `features=dynamodb` or `features=bigtable`");
+    if !cfg!(feature = "bigtable") {
```
jrconlin, line 2: Note for Me:
I was debating that we should drop this check, but I can see us creating a dummy storage system in the future for testing purposes.
I still hold that `bigtable` should be included in the `default` feature set, since we have only one now.

autopush-common/src/db/bigtable/mod.rs
```diff
@@ -150,7 +150,7 @@ impl TryFrom<&str> for BigTableDbSettings {
     type Error = DbError;
     fn try_from(setting_string: &str) -> Result<Self, Self::Error> {
         let me: Self = serde_json::from_str(setting_string)
-            .map_err(|e| DbError::General(format!("Could not parse DdbSettings: {:?}", e)))?;
+            .map_err(|e| DbError::General(format!("Could not parse DbSettings: {:?}", e)))?;
```
jrconlin, line 153: heh, and still very true.

autopush-common/src/db/mod.rs
```diff
@@ -89,12 +81,7 @@ impl StorageType {
         }
         let dsn = dsn
             .clone()
-            .unwrap_or(std::env::var("AWS_LOCAL_DYNAMODB").unwrap_or_default());
-        #[cfg(feature = "dynamodb")]
-        if dsn.starts_with("http") {
-            trace!("Found http");
-            return Self::DynamoDb;
-        }
+            .unwrap_or(std::env::var("GOOGLE_APPLICATION_CREDENTIALS").unwrap_or_default());
```
jrconlin, line 84: No, the `GOOGLE_APPLICATION_CREDENTIALS` is a file holding the connection credentials, not the DSN to the database. We'd still need that DSN value. (`AWS_LOCAL_DYNAMODB` was a DSN to the locally running database).

If the DSN is not set, we should fail out with a config error.

jrconlin (changes requested):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
taddes: yuup 🤦 thanks!

autopush-common/build.rs
taddes: Ah good point, so update this to `default` eh?

jrconlin: No, I'd leave it as `feature = "bigtable"` here. I'd just make  sure that both 
autoendpoint/Cargo.toml & 
autoconnect/Cargo.toml:

```toml
default  = ["bigtable"]
```

So that it gets picked up. (I'm guessing that `"emulator"` slipped into `autoconnect/Cargo.toml` due to a bad merge or something. It's semi-harmless, but should be removed for production code.

taddes:
> I don't know why the build failed in the Docker Image. I was able to build locally. (FWIW, you _might_ be able to get around the build error by wrapping the `loop` in an `Ok(loop{...})`) I'm also building with rust 1.79, so I suppose we could bump that too, provided all the images are updated by now.

Yeah, I am building fine locally but having the same issue in CI for autoconnect. I'm playing around with `tokio` to see how it's pulled in for autoconnect to see if that resolves the issue. I will try what you suggested above to see if it resolves in CI.

autopush-common/src/db/mod.rs
taddes: Thanks for the clarification. Ah, so before it was returning an `Option`, so are you saying I should change the function to return a `Result`? I had looked at just calling `expect()` or `unwrap()` at the end of that chain.

pjenvey:
The current build failure looks to be due to the lack of the tokio `macros` feature in autoconnect-ws, I'm not sure why this has begun happening (nor why it's not happening in the ci test job or elsewhere)

taddes:
@pjenvey @jrconlin appears what it needed was annotating the addition of the `macros` feature in the `autoconnect-ws` crate for `tokio`. Not sure why this came up just here, however it's the only place a macro from `tokio` is used.

jrconlin:
I'm moderately curious why local builds are unaffected by that, but 🤷🏻.

autopush-common/Cargo.toml
```diff
@@ -65,7 +62,7 @@ form_urlencoded = { version = "1.2", optional = true }
 [dev-dependencies]
 mockito = "0.31"
 tempfile = "3.2.0"
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "1.38", features = ["macros"] }
```
jrconlin, line 65:
```suggestion
tokio = { workspace=true, features = ["macros"] }
```

This will pull the version from the main Cargo.toml file, so we don't run the risk of discrepancies later.

jrconlin (changes requested):

autopush-common/Cargo.toml
taddes: Great point!

taddes:
> I'm moderately curious why local builds are unaffected by that, but 🤷🏻.

@jrconlin I think I figured out that it was the `dev-dependencies` that has the flagged feature for `macros`, which would explain why local builds are unaffected. At least that's my best guess!

jrconlin (approved):

PR #730, opened 2024-07-03
feat: Trigger an `update_user()` for channel checks and mobile messages

jrconlin:
This turns Message delivery successes to the bridge, and daily "channel checks" as a reasonably proxy for mobile device "liveliness".

This would greatly benefit from #705.

Closes: SYNC-3397


jrconlin:
Gonna sit on this for a bit, because I keep thinking that I should put the liveliness update stuff into it's own function. The con for that is that it adds another specific function and I'm not really a fan of it.

autoendpoint/src/routes/registration.rs
```diff
@@ -157,14 +157,40 @@ pub async fn new_channel_route(
...
+        db.update_user(&mut user).await?;
+        // report the rough user agent so we know what clients are actively pinging us.
+        let os = auth.user_agent.metrics_os.clone();
+        let browser = auth.user_agent.metrics_browser.clone();
+        // use the "real" version here. (these are less normalized)
+        info!("Mobile client check";
+            "os"=>auth.user_agent.os,
+            "os_version"=>auth.user_agent.os_version,
+            "browser"=>auth.user_agent.browser_name,
+            "browser_version"=>auth.user_agent.browser_version);
```
pjenvey, line 184:
```suggestion
            "os" => auth.user_agent.os,
            "os_version" => auth.user_agent.os_version,
            "browser" => auth.user_agent.browser_name,
            "browser_version" => auth.user_agent.browser_version);
```

pjenvey (changes requested):

autopush-common/src/util/user_agent.rs
```diff
@@ -21,6 +21,8 @@ pub struct UserAgentInfo {
     pub metrics_os: String,
     pub os_version: String,
     pub os: String,
+    // Note, Woothee can determine if a user agent is mobile if the
```
taddes, line 24: optional nit: Maybe placing this note under the definition of the struct?

autopush-common/src/util/user_agent.rs
```diff
@@ -55,6 +57,16 @@ impl From<&str> for UserAgentInfo {
     }
 }
 
+impl From<&actix_web::HttpRequest> for UserAgentInfo {
+    fn from(req: &actix_web::HttpRequest) -> UserAgentInfo {
+        if let Some(header) = req.headers().get(&actix_web::http::header::USER_AGENT) {
+            Self::from(header.to_str().unwrap_or("UNKNOWN"))
+        } else {
+            UserAgentInfo::default()
```
taddes, line 65: This will just create an empty default `UserAngentInfo`. Is there any desire to implement a `default` impl to fine grain those values beyond their empty data types?

taddes (approved):
Non blocking, just 2 minor comments 👍

autopush-common/src/util/user_agent.rs
pjenvey: Good point, we should probably use `Self::from("UNKNOWN")` (which I believe causes woothee to fill in all values with "UNKNOWN"?). Maybe that should be the default instead of the empty strings which doesn't seem useful for the typical usage of this (metrics, sentry tags, etc.)

pjenvey: ..and we could punt that change to another issue

pjenvey (approved):

autopush-common/src/util/user_agent.rs
jrconlin: Yeah, let's punt on switching to "UNKNOWN" as the default for UserAgentInfo. ~I'll file a quick ticket for it.~ see https://mozilla-hub.atlassian.net/browse/SYNC-4355

Issue #733, opened 2024-07-18
Enforce a max number of channels (subscriptions) per user


data-sync-user:
[https://mozilla-hub.atlassian.net/browse/SYNC-4221](https://mozilla-hub.atlassian.net/browse/SYNC-4221|smart-link) will begin updating TTL records for every entry in the router column family so we can enable Bigtable’s garbage collection of router data. Under Bigtable we need to first read the contents of the row, then write the contents out again with new TTLs (cell timestamps).

We currently do not enforce any kind of limit of the number of channels (subscriptions) a user can have. Especially with the added work of reading all the channels out and then writing them, we should begin enforcing some kind of maximum limit of channels to limit the work required. Additionally any user exceeding such a likely large limit of subscriptions is likely either misconfigured or abusing our system.

The enforcement of the max should simply drop the user record, forcing them to recreate a fresh one.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-4356)

data-sync-user:
➤ Philip Jenvey commented:

What should be the maximum? 100? 1000?

data-sync-user:
➤ JR Conlin commented:

Hrm. We could probably add a metric to the mobile channel check and the desktop “hello” to get an idea of the median number of channels a given UAID may have.

data-sync-user:
➤ JR Conlin commented:

Thanks to earlier work, we can now see that the majority of users have ~ 1 channel. the 95th percentile have ~ 2.5 (so say 3) The extreme outliers have more than 5 with anything above 10 probably being questionably abusive. No user appears to have more than 20 channels. While probably an edge case, specifying a limit of 20 may suffice.

PR #734, opened 2024-07-19
feat: add implicit updates of channel records to update_user

pjenvey:
in order to refresh TTLs of all the router row data, allowing automatic garbage collection of inactive router records by Bigtable

Closes SYNC-4221

So this isn't too much code and doesn't appear to be generally all that annoying (yet), shall we go with this route? It saves us a couple extra db calls -- as we are already reading (at least requesting from bigtable) all channels in `get_user`.

### DIFF
--- a/autoconnect/autoconnect-common/src/test_support.rs
+++ b/autoconnect/autoconnect-common/src/test_support.rs
@@ -30,12 +30,12 @@ pub fn hello_db() -> MockDbClient {
 pub fn hello_again_db(uaid: Uuid) -> MockDbClient {
     let mut db = MockDbClient::new();
     db.expect_get_user().times(1).return_once(move |_| {
-        Ok(Some(User {
-            uaid,
-            // Last connected 10 minutes ago
-            connected_at: ms_since_epoch() - (10 * 60 * 1000),
-            ..Default::default()
-        }))
+        let user = User::builder()
+            .uaid(uaid)
+            .connected_at(ms_since_epoch() - (10 * 60 * 1000))
+            .build()
+            .unwrap();
+        Ok(Some(user))
     });
 
     db.expect_update_user().times(1).return_once(|_| Ok(true));
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -29,7 +29,7 @@ use crate::db::{
 };
 
 pub use self::metadata::MetadataBuilder;
-use self::row::Row;
+use self::row::{Row, RowCells};
 use super::pool::BigTablePool;
 use super::BigTableDbSettings;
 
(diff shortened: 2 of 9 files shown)


autoconnect/autoconnect-common/src/test_support.rs
```diff
@@ -30,12 +30,12 @@ pub fn hello_db() -> MockDbClient {
 pub fn hello_again_db(uaid: Uuid) -> MockDbClient {
     let mut db = MockDbClient::new();
     db.expect_get_user().times(1).return_once(move |_| {
-        Ok(Some(User {
-            uaid,
-            // Last connected 10 minutes ago
-            connected_at: ms_since_epoch() - (10 * 60 * 1000),
-            ..Default::default()
-        }))
+        let user = User::builder()
```
jrconlin, line 33: Sad that we can't call `..Default::default()` because the `_channels` item isn't technically writable, but very happy to have a builder method instead.

autopush-common/src/db/mod.rs
```diff
@@ -161,6 +163,19 @@ pub struct User {
...
+    /// automatically have their TTL (cell timestamp) reset during
+    /// [DbClient::update_user].
+    ///
+    /// This is solely used for the sake of that update thus private.
+    /// [DbClient::get_channels] is preferred for reading the latest version of
+    /// the channel ids (partly due to historical purposes but also is a more
+    /// flexible API that might benefit different, non Bigtable [DbClient]
+    /// backends that don't necessarily store the channel ids in the router
+    /// record).
+    _channels: HashSet<Uuid>,
```
jrconlin, line 178: The only problem I have with this is that rust has the syntax that things prefixed with an `_` are unused, not private (like python). This is already private because we don't declare it `pub`, so maybe we should go with something like `priv_channels` or `channel_cache` or something so some future person isn't confused about why we're keeping this unused variable in the structure.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -201,6 +202,46 @@ fn to_string(value: Vec<u8>, name: &str) -> Result<String, DbError> {
     })
 }
 
+/// Parse the "set" (see [DbClient::add_channels]) of channel ids in a bigtable Row.
```
jrconlin, line 205: +1 on extracting these into functions. Thanks!

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -201,6 +202,46 @@ fn to_string(value: Vec<u8>, name: &str) -> Result<String, DbError> {
...
+        result.insert(Uuid::from_str(chid).map_err(|e| DbError::General(e.to_string()))?);
+    }
+    Ok(result)
+}
+
+/// Convert the [HashSet] of channel ids to cell entries for a bigtable Row
+fn channels_to_cells(channels: Cow<HashSet<Uuid>>, expiry: SystemTime) -> Vec<cell::Cell> {
+    let channels = channels.into_owned();
+    let mut cells = Vec::with_capacity(channels.len().min(100_000));
+    for (i, channel_id) in channels.into_iter().enumerate() {
```
taddes, line 229: Is there a desire to check the size of `challens` prior to iteration, or is it intended to be a processing limit at runtime while looping through? Just curious in case a check before iterating was valuable.

taddes (approved):
Lgtm, just had a question about the check for the cell count.

jrconlin (approved):

autopush-common/src/db/bigtable/bigtable_client/mod.rs
pjenvey: Yes, we're going to begin enforcing a max limit of channels shortly, which will probably end up changing this check https://github.com/mozilla-services/autopush-rs/issues/733

PR #736, opened 2024-07-19
feat: Add metrics to return the number of channels per UAID

jrconlin:
This introduces two new metrics: *.busi.channel_count, which return the number of known or declared channels for a given UAID. I'd expect that these metrics will be short lived, but should provide insight into the approximate number of channels that our users have.

Closes: SYNC-4357


autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -54,7 +54,9 @@ impl UnidentifiedClient {
     ) -> Result<(WebPushClient, impl IntoIterator<Item = ServerMessage>), SMError> {
         trace!("❓UnidentifiedClient::on_client_msg");
         let ClientMessage::Hello {
-            uaid, broadcasts, ..
+            uaid,
+            broadcasts,
+            channel_ids,
```
pjenvey, line 59: We ignore this because it's essentially deprecated: Desktop no longer passes in `channel_ids` here, so there's no point in metricing it.

autoendpoint/src/routes/registration.rs
```diff
@@ -192,6 +192,12 @@ pub async fn get_channels_route(
     }
     let channel_ids = db.get_channels(&uaid).await?;
 
+    app_state
+        .metrics
+        .count_with_tags("busi.channel_count", channel_ids.len() as u64)
```
pjenvey, line 197: I don't see how count gets us per uaid stats?

pjenvey (changes requested):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -92,6 +94,13 @@ impl UnidentifiedClient {
                 }
             })
             .send();
+        if let Some(chids) = channel_ids {
+            self.app_state
+                .metrics
+                .count_with_tags("busi.channel_count", chids.len() as u64)
```
taddes, line 100: My general question about this is documenting and making clear what "busi" means, as described in the Jira ticket given it's a business, not performance concern.  However, even calling it a business concern remains a little opaque for me. Is there a plan for there to be more categorical hierarchies that fall under this sub-root in the future? I suppose we could  put out a core metric and add that sub root down the road once there are more within the category, or if more are expected this is a good first pass. 

At the least, I would suggest `business` instead of `busi`, given it's just more explicit for those who might look at it and not know what exactly it's pertaining to.  What do you think?

taddes (changes requested):
Just some things regarding the `busi` sub-root and the general idea around this. See comment. Thx!

autoendpoint/src/routes/registration.rs
jrconlin: We are doing a count for each UAID (since it's done as part of the daily check-in). I expect that the count will be somewhat "noisy" in the way that a number of our metrics are, however the approximate mean should be fairly easy to discern when displayed on a graph. Likewise, significant outliers should also be noticeable, although, I presume specifying the UAID would add too much cardinality. We may be able to later refine that by logging UAIDs that fall outside of the mean with an appropriate buffer, but for now, we don't know what that might be.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
jrconlin: Fair. It's a new category that relates more to the business logic than anything else. Using the longer form might help. Since Autopush has two main components, and therefore two different roots, there must be two different metrics.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -97,7 +97,7 @@ impl UnidentifiedClient {
         if let Some(chids) = channel_ids {
             self.app_state
                 .metrics
-                .count_with_tags("busi.channel_count", chids.len() as u64)
+                .count_with_tags("business.channel_count", chids.len() as u64)
```
taddes, line 100: 👍

taddes (approved):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
jrconlin: Argh. right. Forgot about that. Fortunately, looks like we can take advantage of the tty PR to do this reasonably efficiently.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
```diff
@@ -289,6 +289,8 @@ pub struct ClientFlags {
     pub check_storage: bool,
     /// Flags the need to drop the user record
     pub old_record_version: bool,
+    /// First time a user has connected "today"
+    pub user_gm: bool,
```
taddes, line 293: Just curious about the name `user_gm`. What does the `gm` stand for? Forgive my ignorance if it's a metrics-related acronym I don't know. If not, I'm curious maybe if a more imperative variable name.

taddes (approved):
LGTM, just had a question about `user_gm`, see in rev.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
jrconlin: "good morning". It's fairly meaningless but I needed something to hold the fact that this is the first connection of the day.

I'm very open to better names.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -93,6 +95,26 @@ impl UnidentifiedClient {
...
+            // Return the number of channels for the user using the internal channel_count.
+            // NOTE: this metric can be approximate since we're sampling to determine the
+            // approximate average of channels per user for business reasons.
+            self.app_state
+                .metrics
+                .incr_with_tags("ua.connection.check")
+                // TODO: These should reflect the actual os and browser, but getting those
+                // values at this point would be very complicated. For now, just use "desktop"
+                .with_tag("os", "desktop")
+                .with_tag("browser", "desktop")
```
pjenvey, line 109: Ignore this comment if we're killing all this but it could potentially apply to the `.with_tag_value("desktop")` on the histogram:

If we delay emitting this metric until after creating `WebPushClient` a few lines below we can get the parsed `os`/`browser_name`/etc. values from the pub `wpclient.ua_info`

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
pjenvey: With it solely being used for emitting the new metric I suggest something like `emit_channel_metrics`.

pjenvey (changes requested):
Sorry for the delay

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -93,6 +95,26 @@ impl UnidentifiedClient {
...
+                .send();
```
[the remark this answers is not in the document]
jrconlin: Yeah, it might be a bit extra. The thought was that since the histogram is more about collecting `channel_count()` having a different one that noted the check might capture instances where `channel_count()` was zero. I'm guessing it's probably not that much of a concern, so I'm fine dropping it.

PR #739, opened 2024-07-24
feat: Add ability to track messages based on known VAPID keys

jrconlin:
_*Note*_: This introduces a new setting for autoendpoint: `tracking_keys`
This is a JSON formatted list of the "raw"/x962 formatted VAPID public keys that should be monitored for Push tracking and follows the same format as other key data fields. 

The newly added `script/convert_pem_to_x962.py` script aids by reading a VAPID public key PEM file and outputing a x962 formatted string. 

Closes: [SYNC-4349](https://mozilla-hub.atlassian.net/browse/SYNC-4349)

[SYNC-4349]: https://mozilla-hub.atlassian.net/browse/SYNC-4349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -35,6 +35,10 @@ pub struct Subscription {
     pub user: User,
     pub channel_id: Uuid,
     pub vapid: Option<VapidHeaderWithKey>,
+    /// Should this subscription update be tracked internally?
+    /// (This should ONLY be applied for messages that match known
+    /// Mozilla provided VAPID public keys.)
+    pub trackable: bool,
 }
 
 impl FromRequest for Subscription {
@@ -69,6 +73,11 @@ impl FromRequest for Subscription {
                 .transpose()?;
 
             trace!("raw vapid: {:?}", &vapid);
+            let trackable = if let Some(vapid) = &vapid {
+                app_state.settings.is_trackable(&vapid)
+            } else {
+                false
+            };
 
             // Capturing the vapid sub right now will cause too much cardinality. Instead,
             // let's just capture if we have a valid VAPID, as well as what sort of bad sub
@@ -127,6 +136,7 @@ impl FromRequest for Subscription {
                 user,
                 channel_id,
                 vapid,
+                trackable,
             })
         }
         .boxed_local()
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -38,7 +38,7 @@ pub struct Subscription {
     /// Should this subscription update be tracked internally?
     /// (This should ONLY be applied for messages that match known
     /// Mozilla provided VAPID public keys.)
-    pub trackable: bool,
+    pub tracking_id: Option<String>,
 }
 
 impl FromRequest for Subscription {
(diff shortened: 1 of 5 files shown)


autoendpoint/src/extractors/subscription.rs
```diff
@@ -35,6 +35,10 @@ pub struct Subscription {
     pub user: User,
     pub channel_id: Uuid,
     pub vapid: Option<VapidHeaderWithKey>,
+    /// Should this subscription update be tracked internally?
+    /// (This should ONLY be applied for messages that match known
+    /// Mozilla provided VAPID public keys.)
+    pub tracking_id: Option<String>,
```
jrconlin, line 41: This was changed from a simple bool to an `Option<String>` since that's most likely the final form of this. For now, we're just using a "filler" UUID4, but I strongly suspect that https://mozilla-hub.atlassian.net/browse/SYNC-3400 will introduce a more formalized tracking_id generator.

pjenvey: Seems like the comment still reflects the bool -- so it needs an update

pjenvey (changes requested):

autoendpoint/src/extractors/subscription.rs
jrconlin: Well, it _is_ an option, so it has bool like properties (None vs. Some). I'll clarify the comment.

autoendpoint/src/settings.rs
```diff
@@ -31,6 +32,17 @@ pub struct Settings {
...
 
+    /// A stringified JSON list of VAPID public keys which should be tracked internally.
+    /// This should ONLY include Mozilla generated and consumed messages (e.g. "SendToTab", etc.)
+    /// These keys should be specified in stripped, b64encoded, X962 format (e.g. a single line of
+    /// base64 encoded data without padding).
+    /// You can use `scripts/convert_pem_to_x962.py` to easily convert EC Public keys stored in
+    /// PEM format into appropriate x962 format.
+    pub tracking_keys: String,
+    /// Cached, parsed tracking keys.
+    //TODO: convert to decoded Vec<u8>?
```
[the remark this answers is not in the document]
jrconlin: It kinda is. 
Ideally, the key we track should be the bytearray of the decoded public key. This would prevent things like case confusion or formatting errors. Since we are using a consistently formatted, well known key, we can short cut things for now and just do string matches, but that's very much a cheat.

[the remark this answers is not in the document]
pjenvey: Ok, but I think we can still kill this `tracking_vapid_pubs` field entirely since we're only calling `tracking_keys()` once at startup now. Then the comment should probably move to that method

autoendpoint/src/settings.rs
```diff
@@ -169,10 +199,39 @@ impl Settings {
...
+    /// Extract the message Id from the headers (if present), otherwise just make one up.
+    pub fn get_tracking_id(&self, headers: &HeaderMap) -> String {
+        headers
+            .get("X-MessageId")
+            .map(|s| {
+                s.to_str()
+                    .unwrap_or(uuid::Uuid::new_v4().as_simple().to_string().as_str())
+                    .to_string()
+            })
+            .unwrap_or(uuid::Uuid::new_v4().as_simple().to_string())
```
pjenvey, line 222: These chains can be a bit much but I'd argue utilzing `and_then`/`ok` improves this since we'll only need `new_v4` at the very end (also surprised clippy didn't force `unwrap_or_else` here).
```suggestion
            .and_then(|val| val.to_str().ok())
            .map(|s| s.to_owned())
            .unwrap_or_else(|| uuid::Uuid::new_v4().as_simple().to_string())
```

pjenvey (changes requested):

pjenvey (approved):

PR #741, opened 2024-07-25
feat: add APNS timeouts for a2

taddes:
Per [`ClientConfig`](https://github.com/WalletConnect/a2/blob/master/src/client.rs#L78) struct in a2 library, updated autoconnect Settings struct to hold `request_timeout_secs` and `pool_idle_timeout_secs`, define `default()` struct to pass into `ClientConfig`, with defaults set to what are the current defaults in a2. We can of course change them going forward. This way the settings are untethered from whatever a2 may change them to.

Some contextual comments are included for clarity.

Closes [SYNC-4307](https://mozilla-hub.atlassian.net/browse/SYNC-4307)

[SYNC-4307]: https://mozilla-hub.atlassian.net/browse/SYNC-4307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -46,6 +46,12 @@ pub struct Settings {
     pub fcm: FcmSettings,
     pub apns: ApnsSettings,
 
+    // These values correspond to the a2 library ClientConfig struct.
+    // https://github.com/WalletConnect/a2/blob/master/src/client.rs#L65-L71.
+    // Utilized by autoendpoint router config in creating the client.
+    pub request_timeout_secs: Option<u64>,
+    pub pool_idle_timeout_secs: Option<u64>,
+
     #[cfg(feature = "stub")]
     pub stub: StubSettings,
 }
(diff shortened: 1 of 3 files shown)


autoendpoint/src/settings.rs
```diff
@@ -46,6 +46,12 @@ pub struct Settings {
     pub fcm: FcmSettings,
     pub apns: ApnsSettings,
 
+    // These values correspond to the a2 library ClientConfig struct.
```
jrconlin, line 49: These settings are specific to APNS only. They should be placed in `ApnsSettings`.

jrconlin (changes requested):

autoendpoint/src/settings.rs
taddes: Np, I'll move them. The ticket had mentioned creating Autoendpoint settings and then passing those values to a2 during router configuration, so I put it in the autoendpoint settings. Does make more sense for them to be with apns, indeed!

jrconlin (approved):

PR #742, opened 2024-07-26
feat: Normalize and explain TTLs maximums

jrconlin:
This normalizes to a max channel TTL of 30 days and a max router TTL of 2 * max channel TTL.

See doc comment in autopush-common for details.

This also adds two convenience functions to convert SystemTime and UTC millisecond offsets into formatted DateTime strings. These were useful to identify the bug and may also be useful for various logging statements.

There was also a bug in one of the unit tests, where we were taking the cell timestamp for `connected_at` instead of the cell value. How this did not fail before is a mystery.


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1832,13 +1833,20 @@ mod tests {
         };
 
         // Ensure the initial expiry (timestamp) of all the cells in the row
-        let expiry = row.take_required_cell("connected_at").unwrap().timestamp;
+        let connected_at = SystemTime::UNIX_EPOCH
```
jrconlin, line 1836: Sigh. Yeah, this was using cell timestamp instead of the actual value. No idea how this was passing originally.

pjenvey: But that's the point of the test: ensuring ttls are updated. The `connected_at` value isn't important here, I only chose it because it's the first field read/write in the serialization methods

jrconlin: Ah, if that's the case, then we should make sure that all records, including the `chid:*` items are updated whenever we call `increment_storage`, since that's currently failing. Thanks!

taddes (approved):
LGTM. Thanks for the useful BigTable context. Just had one suggestion for readability but totally optional. Probably best to have @pjenvey look too given better experience with BigTable

pjenvey (changes requested):
The problem with the test here is due to treating the channel records TTL differently from the rest of router's.

I think the channel record ttl in the router CF should match the other router CF ttls, whereas the `MAX_CHANNEL_TTL` of 30 days should be solely for notifications stored in the message CFs (probably renaming it to `MAX_NOTIFICATION_TTL`)

jrconlin:
Adding a discussion note here so that I don't lose the thought again:

I've added a comment to the top of `update_user()` that encapsulates how and when the ROUTER record has it's `cell.timestamp` value updated. Hopefully, this will help me remember in the future.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1853,16 +1868,17 @@ mod tests {
             panic!("Expected row");
         };
 
-        let expiry2 = row.take_required_cell("connected_at").unwrap().timestamp;
-        assert!(expiry2 > expiry);
+        let ca_expiry2 = row.take_required_cell("connected_at").unwrap().timestamp;
+
+        assert!(ca_expiry2 >= ca_expiry);
```
pjenvey, line 1873: I think this should still be `>` to ensure it was touched. If it's causing you any flake let's just add a small sleep instead.
```suggestion
        assert!(ca_expiry2 > ca_expiry);
```

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1853,16 +1868,17 @@ mod tests {
...
         for mut cells in row.cells.into_values() {
             let Some(cell) = cells.pop() else {
                 continue;
             };
-            assert_eq!(
-                cell.timestamp, expiry2,
-                "{} cell timestamp should match connected_at's",
+            assert!(
+                cell.timestamp >= ca_expiry2,
+                "{} cell timestamp expiry should exceed connected_at's",
```
pjenvey, line 1881: Is this change necessary? IMHO I prefer the stricter check (ensuring everything was touched).

autopush-common/src/lib.rs
```diff
@@ -19,3 +19,32 @@ pub mod test_support;
...
+/// [60 days](https://searchfox.org/mozilla-central/search?q=OFFER_PROFILE_RESET_INTERVAL_MS),
+/// the User Agent suggest "refreshing Firefox", which essentially throws away one's
+/// current profile. This would include all subscriptions a user may have had.
+///
+/// To that end, messages left unread for more than 30 days should be considered
+/// "abandoned" and any router info assigned to a User Agent that has not contacted
+/// Autopush in 60 days can be discarded.
+///
+/// The maximum TTL for channels, 30 days in seconds
+pub const MAX_CHANNEL_TTL: u64 = 30 * 24 * 60 * 60;
```
pjenvey, line 48: I still think this needs some kind of rename with `channels` now living in the router record: they now exist for 60 days as they're in the router record.
```suggestion
pub const MAX_NOTIFICATION_TTL: u64 = 30 * 24 * 60 * 60;
```

pjenvey (changes requested):

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: It was, I'll add a sleep.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
jrconlin: Since the timestamp here is the time that bigtable will be (potentially) cleaning up the cell, we want to make sure that doesn't happen until after the expiration timestamp for the data. So, yes, the time we clean up the data should happen either at or later than the time that we want the data to live.

autopush-common/src/lib.rs
jrconlin: Yeah, agree about the naming, but I didn't want to add more crap into this PR.
I'll open a cleanup task to do that.

autopush-common/src/lib.rs
```diff
@@ -19,3 +19,32 @@ pub mod test_support;
...
+/// a user has not used the User Agent in more than
+/// [60 days](https://searchfox.org/mozilla-central/search?q=OFFER_PROFILE_RESET_INTERVAL_MS),
+/// the User Agent suggest "refreshing Firefox", which essentially throws away one's
+/// current profile. This would include all subscriptions a user may have had.
+///
+/// To that end, messages left unread for more than 30 days should be considered
+/// "abandoned" and any router info assigned to a User Agent that has not contacted
+/// Autopush in 60 days can be discarded.
+///
+/// The maximum TTL for channels, 30 days in seconds
```
pjenvey, line 47:
```suggestion
/// The maximum TTL for notifications, 30 days in seconds
```

pjenvey (changes requested):

pjenvey (approved):

PR #744, opened 2024-08-02
feat: [SYNC-4365]: Report test results and coverage on CI

b4handjr:
This PR splits the current `test` job, which runs both rust and integration-tests, to 2 individual jobs. I have also added a job just for the `cargo format` step. The 2 new jobs use a new cache which will need to be generated on the first run so the builds will seem long but should shorten down significantly once the cache is built.

The `Rust Test` job now reports code coverage. The file is quite large but it is JSON and is uploaded as an artifact. This is done via [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov). The Integration test job also reports its test results but it is stored using `store_test_results`. I can also add this as an artifact if needed.

This should get us the test metrics we want and also shorten our overall test time on CI a bit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ commands:
             else
               echo "${PASS}" | docker login -u="${USER}" --password-stdin
             fi
-  setup-rust:
+  setup_rust:
     steps:
       - run:
           name: Set up Rust and build applications
@@ -47,6 +47,52 @@ commands:
             echo 'export PATH=$PATH:$HOME/.cargo/bin' >> $BASH_ENV
             rustc --version
             cargo build --features=emulator
+  setup_bigtable:
+    steps:
+      - run:
+          name: Setup Bigtable
+          command: scripts/setup_bt.sh
+  setup_cbt:
+    steps:
+      - run:
+          name: Set up cbt
+          command: |
+            echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+            curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | gpg --dearmor -o /usr/share/keyrings/cloud.google.gpg
+            apt-get update -y
+            apt install google-cloud-cli-cbt -y
+  create_test_result_workspace:
+    steps:
+      - run:
+          name: Create Workspace
+          command: mkdir -p workspace/test-results
+  restore_test_cache:
+    parameters:
+      cache_key:
+        type: string
+    steps:
+      - restore_cache:
+          name: Restoring Rust cache
+          key: <<parameters.cache_key>>
+  save_test_cache:
+    parameters:
+      cache_key:
+        type: string
+    steps:
+      - save_cache:
+          name: Save Rust cache
+          key: <<parameters.cache_key>>
+          paths:
+            - target
+            - ~/.cargo/registry
+            - ~/.cargo/git
+  setup_python:
+    steps:
+      - run:
+          name: Set up Python
+          command: |
+            pip install --upgrade pip
+            pip install poetry
 
 jobs:
   audit:
@@ -104,38 +150,23 @@ jobs:
           username: $DOCKER_USER
           password: $DOCKER_PASS
         command: gcloud beta emulators bigtable start --host-port=localhost:8086
-    resource_class: xlarge
+    resource_class: medium
     environment:
       BIGTABLE_EMULATOR_HOST: localhost:8086
     steps:
       - checkout
-      - restore_cache:
-          name: Restoring Rust cache
-          key: rust-v1-integration-test-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
-      - run:
-          name: Create Workspace
-          command: mkdir -p workspace
-      - setup-rust
+      - restore_test_cache:
+          cache_key: rust-v1-integration-test-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
+      - create_test_result_workspace
+      - setup_rust
       - run:
           name: Set up system
           command: |
             apt update
             apt install libssl-dev apt-transport-https ca-certificates gnupg curl cmake -y
-      - run:
-          name: Set up cbt
-          command: |
-            echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
-            curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | gpg --dearmor -o /usr/share/keyrings/cloud.google.gpg
-            apt-get update -y
-            apt install google-cloud-cli-cbt -y
-      - run:
-          name: Set up Python
-          command: |
-            pip install --upgrade pip
-            pip install poetry
-      - run:
-          name: Setup Bigtable
-          command: scripts/setup_bt.sh
+      - setup_cbt
+      - setup_python
+      - setup_bigtable
       - run:
           name: Integration tests (Bigtable)
           command: make integration-test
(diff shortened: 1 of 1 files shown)


taddes:
Thanks for this @jrbenny35 , going to look now. For some reason I'm having a memory of a caching issue back on merino-py where cached images caused test failures due to a variance between the application and test builds. I'm going to try and look back to see what that was in case it could possibly manifest here, but maybe not.  It'll be great to have those metrics!

jrbenny35:
> Thanks for this @jrbenny35 , going to look now. For some reason I'm having a memory of a caching issue back on merino-py where cached images caused test failures due to a variance between the application and test builds. I'm going to try and look back to see what that was in case it could possibly manifest here, but maybe not. It'll be great to have those metrics!

I can into a little of this but it was because I changed the docker image and didn't change the cache key. I have bumped the cache keys here, and I also created one specific for the integration tests.

.circleci/config.yml
```diff
@@ -160,12 +148,102 @@ jobs:
...
       - save_cache:
           name: Save Rust cache
-          key: rust-v1-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
+          key: rust-v1-integration-test-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
           paths:
             - target
             - ~/.cargo/registry
             - ~/.cargo/git
 
+  test-rust:
```
Trinaa, line 157: Could we call this job `test-unit`?

.circleci/config.yml
```diff
@@ -160,12 +148,102 @@ jobs:
...
+          destination: cov.json
+      - save_cache:
+          name: Save Rust cache
+          key: rust-v2-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
+          paths:
+            - target
+            - ~/.cargo/registry
+            - ~/.cargo/git
+
+  check-rust-formatting:
```
Trinaa, line 217: Suggestion to call this job `rust-checks` to be more in line with the python static analysis job called `python-checks`.

.circleci/config.yml
```diff
@@ -160,12 +148,102 @@ jobs:
...
+    docker:
+      - image: python:3.12-slim-bookworm
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+        environment:
+          RUST_BACKTRACE: 1
+          RUST_TEST_THREADS: 1
+    resource_class: xlarge
```
Trinaa, line 226: I wonder if we could use the medium or large resource class for static analysis and save some CircleCI credits?

.circleci/config.yml
```diff
@@ -160,12 +148,102 @@ jobs:
...
+      - run:
+          name: Setup Bigtable
+          command: scripts/setup_bt.sh
+      - run:
+          name: Install cargo-llvm-cov
+          command: cargo install cargo-llvm-cov
+      - run:
+          name: Report tests and coverage
+          command: |
+            cargo llvm-cov --json --output-path workspace/test-results/cov.json test --features=emulator --features=bigtable --jobs=2
```
Trinaa, line 205: The following comment was previously present around this step, and i wonder if it should be preserved?

```
# Note: This build can potentially exceed the amount of memory availble to the CircleCI instance.
# We've seen that limiting the number of jobs helps reduce the frequency of this. (Note that
# when doing discovery, we found that the docker image `meminfo` and `cpuinfo` often report
# the machine level memory and CPU which are far higher than the memory allocated to the docker
# instance. This may be causing rust to be overly greedy triggering the VM to OOM the process.)
```

Trinaa (commented):
imo there is opportunity still to reduce duplication and increase comprehensibility at the job level by abstracting common steps into reusable commands. 
Suggested common steps to be converted to command:

- "Set up cbt" (L125 & L186)
- "Set up Python" (L86 & L132)
- "Restoring Rust cache" (L12, L178 & L233) (parameter required)
- "Save Rust cache" (L149 & L209) (parameter required)
- "Create Workspace" (L116 & L182) --> Maybe rename to "Create Test Result Workspace"
- "Setup Bigtable" (L137 & L197)


In the name of saving CPU cycles and reducing costs, is it possible to execute the common setup steps in `test-rust` and `test-integration` in a shared requisite job, persisting the outcomes, like '~/.cargo', in a workspace and re-using them?

jrbenny35:
>  is it possible to execute the common setup steps in test-rust and test-integration in a shared requisite job, persisting the outcomes, like '~/.cargo', in a workspace and re-using them?

I tried this, the major concern is that it adds extra time. Having to persist to the workspace as well as saving the cache takes time, and then the `attach_workspace` also takes time. I found it was quicker to just restore the cache once per job and then do the `cargo` steps. It should use the cache and the build times are usually less than a min. `persist_to_workspace` would sometimes take up to 5 min, and the requisite `attach_workspace` would also take at least 2-3 min.

We have explored this on experimenter before and came to the same conclusion. Having the individual jobs build and test is usually quicker than 1 job building and having to the store those files (especially if it is a lot of files, for autopush it was 3.5gb), and then extracting those files in the following jobs.

jrbenny35:
> imo there is opportunity still to reduce duplication and increase comprehensibility at the job level by abstracting common steps into reusable commands.

I can do this, it's pretty straightforward.

.circleci/config.yml
jrbenny35: Looks like we need at least large or we run out of memory.

https://stackoverflow.com/questions/46259776/collect2-fatal-error-ld-terminated-with-signal-9-killed

https://app.circleci.com/pipelines/github/mozilla-services/autopush-rs/2787/workflows/217086e8-0c5d-47f7-a2d3-80da9c6087a1/jobs/15479/parallel-runs/0/steps/0-104

Trinaa (commented):
Your changes make the build a couple minute faster, that's pretty awesome 🚀 

I noticed for the unit tests, we are getting the coverage report alright, but the test results aren't being reported ([link](https://app.circleci.com/pipelines/github/mozilla-services/autopush-rs/2788/workflows/f3ff4cbe-2b92-4609-937f-9ed573070a8c/jobs/15491))?
image

jrbenny35:
> Your changes make the build a couple minute faster, that's pretty awesome 🚀
> 
> I noticed for the unit tests, we are getting the coverage report alright, but the test results aren't being reported ([link](https://app.circleci.com/pipelines/github/mozilla-services/autopush-rs/2788/workflows/f3ff4cbe-2b92-4609-937f-9ed573070a8c/jobs/15491))? image

I think this is because it's not an XML which is what circle needs. I think this goes towards what I sent you on Monday about generating an XML report.

Trinaa:
@jrbenny35 Aww

> > Your changes make the build a couple minute faster, that's pretty awesome 🚀
> > I noticed for the unit tests, we are getting the coverage report alright, but the test results aren't being reported ([link](https://app.circleci.com/pipelines/github/mozilla-services/autopush-rs/2788/workflows/f3ff4cbe-2b92-4609-937f-9ed573070a8c/jobs/15491))? image
> 
> I think this is because it's not an XML which is what circle needs. I think this goes towards what I sent you on Monday about generating an XML report.

Okay I understand, so in Contile we are using `cargo2junit` to accomplish this need, but you are suggesting to perhaps use [cargo-nextest](https://docs.trunk.io/flaky-tests/frameworks/cargo-nextest). @jrconlin, @pjenvey or @taddes any thoughts on this?

.circleci/config.yml
```diff
@@ -90,81 +150,113 @@ jobs:
...
-      # Need to download the poetry.lock files so we can use their
-      # checksums in restore_cache.
-      - restore_cache:
-          name: Restoring Rust cache
-          key: rust-v1-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
-      - run:
-          name: Create Workspace
-          command: mkdir -p workspace
+      - restore_test_cache:
+          cache_key: rust-v1-integration-test-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "Cargo.lock" }}
```
jrconlin, line 159: What is the difference between the `rust-v1` we use here and the `rust-v2` we use later?

jrconlin (approved):

jrconlin (approved):

.circleci/config.yml
jrbenny35: This one is specific to the integration tests. The `v2` of the other one is because I am using a different image and the paths aren't the same

jrbenny35:
> Okay I understand, so in Contile we are using `cargo2junit` to accomplish this need, but you are suggesting to perhaps use [cargo-nextest](https://docs.trunk.io/flaky-tests/frameworks/cargo-nextest). @jrconlin, @pjenvey or @taddes any thoughts on this?

I want to say I tried this but there was some weird behavior because the `cargo2junit` library needs a "nightly" version of rust and that didn't allow the autopush project to compile correctly.

Trinaa (approved):

PR #748, opened 2024-08-20
Add junit reporter for unit tests

b4handjr:
This adds Junit XML reporting for the unit tests on circleci, and it uploads them to `test_artifacts`. This allows us to see a report of any failures on circleci in the `Tests` tab.

I needed to increase the circleci `resource_class` as it seems to be running out of memory more often now, but I also reduced the amount of builds a job does, and even removed a build for the `rust_checks` job. This should result in faster jobs too (just by a little).

I also changed the cache key as it was using the branch as a part of the key and this won't really help us as each PR will be on a different branch, so it is now setup to just use a checksum of  `Cargo.lock`.

Fixes https://mozilla-hub.atlassian.net/browse/SYNC-4369

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ commands:
   setup_rust:
     steps:
       - run:
-          name: Set up Rust and build applications
+          name: Set up Rust
           command: |
             apt update
             apt install build-essential curl libstdc++6 libstdc++-12-dev libssl-dev pkg-config -y
@@ -46,7 +46,11 @@ commands:
             export PATH=$PATH:$HOME/.cargo/bin
             echo 'export PATH=$PATH:$HOME/.cargo/bin' >> $BASH_ENV
             rustc --version
-            cargo build --features=emulator
+  build_applications:
+    steps:
+      - run:
+          name: Build Applicatins
+          command: cargo build --features=emulator
   setup_bigtable:
     steps:
       - run:
@@ -159,6 +163,7 @@ jobs:
           cache_key: rust-v1-integration-test-{{ checksum "Cargo.lock" }}
       - create_test_result_workspace
       - setup_rust
+      - build_applications
       - run:
           name: Set up system
           command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -222,12 +230,14 @@ jobs:
       - run:
           name: Report tests and coverage
           command: |
-            cargo llvm-cov --json --output-path workspace/test-results/cov.json test --features=emulator --features=bigtable --jobs=2
+            cargo llvm-cov --json --output-path workspace/test-results/cov.json nextest --features=emulator --features=bigtable --jobs=2 --profile=ci
       - store_artifacts:
           path: ~/project/workspace/test-results/cov.json
           destination: cov.json
```
Trinaa, line 236: We need to have the xml report stored as an artifact as well in this step so that we can download it from the API. The `store_test_results` makes things observable in CircleCI, but we also want to scrape the xml atm.

I wonder if you'd be willing to make this change for the integration tests too as part of this PR? It's a bit of scope creep, but I noticed this is missing there too.

Trinaa (commented):
This looks good to me so far, I've requested one tweak and a tiny scope creep 🙇‍♀️

If Sync confirm that the 2xlarge resource class is alright in today's meeting, I think we can merge. I've added the topic to today's [notes](https://docs.google.com/document/d/1IlWgBHr_sSguxUpZPZutvU27Th49_tf2xjognnMEAJg/edit)

\*\* **update**  \*\* We got confirmation from Sync that 2xlarge is approved.

.circleci/config.yml
b4handjr: I thought of this but wasn't sure. I can make the change to both, i think thats fine

b4handjr: See https://github.com/mozilla-services/autopush-rs/pull/748/commits/60c6c0c14dc3a8f5ae650df65d2832994cb645e5

Trinaa (approved):
Awesome! Thanks so much for this work @b4handjr !!! 🚀

PR #749, opened 2024-08-26
chore: Simplify publish_docs workflow

jrconlin:
(note: this is a re-submit of #747 to resolve CI issues.)

from [janbrasna](https://github.com/janbrasna)

The current docs build takes several minutes, the toolchain setup doesn't really do anything, and the workflow is triggered from unexpected events (probably not doing what it was believed to do) — so this is an attempt to simplify, and speed up. The main issues are:

    the curl doesn't run due to -y error; OTOH neither rustup nor cargo
needs any manual setting up here…
    the on:check_run doesn't imply successful build, but more
importantly triggers also outside of master/PRs basically on any branch anywhere commit checks are run, publishing docs on pretty much any push to any ref unnecessarily.
    something in the build of validator_derive (either validator, grpcio
or google-cloud-rust-raw) takes ~6-7mins to build, basically 99% of the whole workflow run, so question is whether to switch to binaries or caching?
    The used Node 16 actions raise deprecation warnings, need updating.

Logs:

    before: runs/10327553562 (triggered from RRM181 PR check that hasn't
landed yet)
    after: job/28971156482#step:6:1
    after cache: job/28972053092#step:6:1

More detailed explainers are in [the changeset](https://github.com/mozilla-services/autopush-rs/pull/747).

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -1,57 +1,37 @@
 name: Publish docs to pages
 run-name: ${{ github.actor }} is publishing docs
 on:
-    # on a successful push
+    workflow_dispatch:  # Allow manual triggering
     push:
-        # only run on pushes to these branches
-        branches:
-            - 'master'
-            - 'main'
-            - 'docs'
-        # or run if pushes to these files.
-        # paths:
-        #    - 'docs/**.md'
-    check_run:
-        types:
-            # Only run if we have a successfully completed branch
-            - completed
-    # if desired, we can also use cron like rules to trigger these
-    # schedule:
-        #   - cron: 0 0 * * *
-    # Allow for manual triggering
-    workflow_dispatch:
+        branches: [master,main]
+        # paths: ['docs/**.md']  # API docs need building always
 
-# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
-permissions:
-    contents: read
-    pages: write
-    id-token: write
-
-# Allow only one concurrent deployment. Cancel any in-progress.
 concurrency:
-    group: "pages"
-    cancel-in-progress: true
+    group: github-pages
+    cancel-in-progress: false  # Skip any intermediate builds but finish deploying
 
 jobs:
     build:
         runs-on: ubuntu-latest
         env:
-            MDBOOK_ENV: 0.4.24
-            MERMAID_ENV: 0.12.6
+            MDBOOK_ENV: 0.4.40
+            MERMAID_ENV: 0.13.0
             DEST_DIR: /home/runner/.cargo/bin
         steps:
-            # these are other job descriptions to call.
-            - uses: actions/checkout@v3
-            - name: Configure rust  # TODO: can we export building rust and installing mdbook as an artifact and reuse it?
-              run: |
-                curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf -y | sh
-                rustup update
-                export PATH=$PATH:$DEST_DIR
+            - uses: actions/checkout@v4
+            - uses: actions/cache@v4
+              with:
+                path: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_docs.yml
```diff
@@ -1,57 +1,37 @@
...
-            MERMAID_ENV: 0.12.6
+            MDBOOK_ENV: 0.4.40
+            MERMAID_ENV: 0.13.0
             DEST_DIR: /home/runner/.cargo/bin
         steps:
-            # these are other job descriptions to call.
-            - uses: actions/checkout@v3
-            - name: Configure rust  # TODO: can we export building rust and installing mdbook as an artifact and reuse it?
-              run: |
-                curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf -y | sh
```
jrconlin, line 47: See [comment](https://github.com/mozilla-services/autopush-rs/pull/747#discussion_r1729953076)
`cargo` is now a part of the distribution core and we no longer need to import it directly.

.github/workflows/publish_docs.yml
```diff
@@ -1,57 +1,37 @@
...
     push:
-        # only run on pushes to these branches
-        branches:
-            - 'master'
-            - 'main'
-            - 'docs'
-        # or run if pushes to these files.
-        # paths:
-        #    - 'docs/**.md'
-    check_run:
```
jrconlin, line 14: See [comment](https://github.com/mozilla-services/autopush-rs/pull/747#discussion_r1723323064)

.github/workflows/publish_docs.yml
```diff
@@ -1,57 +1,37 @@
...
-    contents: read
-    pages: write
-    id-token: write
-
-# Allow only one concurrent deployment. Cancel any in-progress.
 concurrency:
-    group: "pages"
-    cancel-in-progress: true
+    group: github-pages
+    cancel-in-progress: false  # Skip any intermediate builds but finish deploying
```
taddes, line 11: Just a question: do we for sure want a successful deploy should something in docs build fail? I wonder if this would result in us missing failed doc builds. However, it's not critical to the service health so I get the rationale here.

.github/workflows/publish_docs.yml
taddes: Should we still possibly run `rustup update`? There may be a variance between the distro version and what's currently available.

taddes (approved):
LGTM, just added 2 comments/questions that aren't blocking.

.github/workflows/publish_docs.yml
janbrasna: it's kept uptodate in the runners https://github.com/mozilla-services/autopush-rs/pull/747#discussion_r1729953076 ~ and trying to update it manually only triggers warnings https://github.com/mozilla-services/autopush-rs/pull/747#discussion_r1723311791 because rustup isn't managing parts of the toolchain: [#step:3:16](https://github.com/mozilla-services/autopush-rs/actions/runs/10564554553/job/29267171191#step:3:16)

.github/workflows/publish_docs.yml
janbrasna: If this is related to the concurrency lines directly, then they don't change anything about success of the builds. They only manage how many docs workflow runs can queue in succession, always skipping anything between the first and the last run for this concurrency group. It only changes one thing: don't cancel the one already running and wait for it to finish before starting the most recent.
(the rationale is: you never know if the newer build will be {more} successful, so chances are if the running job can deploy, let it deploy on success… in case the newer run won't build etc.)

This won't change any failed builds behaviour at all. These still fail loudly, and notify accordingly https://github.com/mozilla-services/autopush-rs/pull/747#discussion_r1729952533 — and won't deploy if the build fails.

PR #746, opened 2024-08-09
feat: push Docker images to GAR

emaydeck-mozilla:
I took a stab at updating the CircleCI config to push images to GAR after they're pushed to Docker Hub. This approach requires the [circleci/gcp-gcr orb](https://circleci.com/developer/orbs/orb/circleci/gcp-gcr).

Requirements:

- [x] New environment variables added to CircleCI project
- [x] Workload Identity configured for service account
- [ ] New workflows tested

PR #750, opened 2024-08-27
feat: use GHA to build and push images to GAR

emaydeck-mozilla:
This creates a GitHub Actions workflow that builds and pushes Docker images to Google Artifact Registry (GAR). This can either be triggered in either of the following ways:

- Automatically through successful Circle CI tests when a new tag is created
- Manually as a workflow_dispatch action

This PR does not break or change the existing CircleCI workflow that builds Docker images and pushes them to Docker Hub. Once we know that this GHA workflow is behaving as expected, we can remove those steps from the CircleCI workflow and rename that to something else (ie. "tests" instead of "build-test-deploy").

Unfortunately, it seems GitHub won't let me test the new workflow until it's present in `master`. Once it's there, I can test/refine more as needed from a branch.

This replaces https://github.com/mozilla-services/autopush-rs/pull/746.

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -1,6 +1,6 @@
 name: Build and push image to GAR
 
-run-name: Build and push image "${inputs.image}:${inputs.tag}" to GAR
+run-name: Build and push image "${{ inputs.image }}:${{ inputs.tag }}" to GAR
 
 on:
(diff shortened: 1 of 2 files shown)


jrconlin (approved):

dlactin (approved):
lgtm!

pjenvey (approved):

PR #752, opened 2024-08-28
fix: fixed workflow_dispatch runs for docker.yml GHA workflow

emaydeck-mozilla:
This includes fixes for the GHA workflow that builds and pushes images to GAR.

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -73,9 +73,7 @@ jobs:
         with:
           context: .
           file: Dockerfile
-          tags:
-            - ${{ env.GAR_LOCATION }}-docker.pkg.dev/${{ env.GCP_PROJECT_ID }}/${{ env.GAR_REPOSITORY }}/${{ env.IMAGE }}:latest
-            - ${{ env.GAR_LOCATION }}-docker.pkg.dev/${{ env.GCP_PROJECT_ID }}/${{ env.GAR_REPOSITORY }}/${{ env.IMAGE }}:${{ env.TAG }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-push-image.yml
```diff
@@ -12,15 +12,15 @@ on:
         required: true
         type: string
       gar_location:
-        required: true
+        required: false
```
jrconlin, line 15: What happens if these are not specified? I see that the tags below still draw from `env.GAR_LOCATION` for instance. I presume that those are just being inherited?

jrconlin (commented):
Also don't know if you want to silence the formatting error in the Dockerfile.

.github/workflows/build-and-push-image.yml
emaydeck-mozilla: I encountered issues when I required inputs even with defaults supplied here: https://github.com/mozilla-services/autopush-rs/actions/runs/10605118999

I figured it would make the most sense to specify the defaults here instead of including them as inputs upstream in docker.yml.

In this case, since I specify default values when `required: false` is set, we should be fine. This particular workflow may only be triggered by another workflow, so there should be no chance we see any errors.

jrconlin: Ah, cool. We probably ought to document that somewhere so that future us don't do dumb things, but that can be a different PR.

emaydeck-mozilla: I don't mind throwing in a comment now!

emaydeck-mozilla: Added a comment to that file now

pjenvey (approved):

PR #753, opened 2024-08-29
fix: change logic for image build triggers

emaydeck-mozilla:
It turns out `check_run` and `check_suite` events only occur on the default branch, so I can't verify tags the way I was before. In this case, I can leverage the `release/<tag>` branch name in `.check_suite.branch_name` to determine that tagging should occur and to identify the tag name.

In the future, we will probably want to create a workflow that initiates a tagging event, runs tests, and then builds/pushes the image.


jrconlin (approved):

pjenvey (approved):

emaydeck-mozilla:
Let's pursue the approach in https://github.com/mozilla-services/autopush-rs/pull/746 first. There's value in keeping CI workflows together and moving everything to GHA together.

PR #754, opened 2024-08-30
fix: use tag pushes as trigger for image builds

emaydeck-mozilla:
This simplifies the Build and deploy Docker images to GAR workflow to use tag pushes as a trigger instead of waiting for CircleCI tests to pass for that new tag. This workflow assumes that we are satisfied with the tests that are executed during pull requests, using a tag push as the source of truth to dev and stage deployments.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,7 +1,8 @@
 name: Build and deploy Docker images to GAR
 on:
   push:
-    tags: *
+    tags:
+      - '*'
   workflow_dispatch:
     inputs:
       git_tag:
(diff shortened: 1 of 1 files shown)


jrconlin (approved):

jrconlin (approved):

.github/workflows/docker.yml
```diff
@@ -1,7 +1,8 @@
 name: Build and deploy Docker images to GAR
 on:
-  check_run:
-    types: [completed]
+  push:
+    tags:
+      - '*'
```
jrconlin, line 5: Ah, yeah, makes sense that they'd be an array. Sorry for missing that.

jrconlin (approved):

emaydeck-mozilla:
Let's pursue the approach in https://github.com/mozilla-services/autopush-rs/pull/746 first. There's value in keeping CI workflows together and moving everything to GHA together.

PR #758, opened 2024-09-04
doc: Update docs to indicate that `sub` is required.

jrconlin:
We require the `sub` so that we can contact folk that are possibly doing questionable things. While we don't validate the contact URI, if we see a problem and we can't contact folk, it probably means we're just going to block them.

Closes #SYNC-3678

### DIFF
--- a/autoendpoint/src/headers/vapid.rs
+++ b/autoendpoint/src/headers/vapid.rs
@@ -11,6 +11,14 @@ use autopush_common::util::{sec_since_epoch, ONE_DAY_IN_SECONDS};
 
 pub const ALLOWED_SCHEMES: [&str; 3] = ["bearer", "webpush", "vapid"];
 
+/*
+The Assertion block for the VAPID header.
+
+Please note: We require the `sub` claim in addition to the `exp` and `aud`.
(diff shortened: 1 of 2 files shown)


taddes (approved):
👍 Thanks for adding this useful context.

autoendpoint/src/headers/vapid.rs
```diff
@@ -11,6 +11,14 @@ use autopush_common::util::{sec_since_epoch, ONE_DAY_IN_SECONDS};
...
 pub const ALLOWED_SCHEMES: [&str; 3] = ["bearer", "webpush", "vapid"];
 
+/*
+The Assertion block for the VAPID header.
+
+Please note: We require the `sub` claim in addition to the `exp` and `aud`.
+See [HTTP Endpoints for Notficiations::Lexicon::{vapid_key}](https://mozilla-services.github.io/autopush-rs/http.html#lexicon-1)
+for details.
+
+ */
```
ddurst, line 21: Actually took me a minute to realize I had to click through to the IETF site to get the definitions for `aud`, `exp`, and `sub`, but I think that's reasonable for someone who would be in here.

ddurst (approved):
Ticks the box. 👍

PR #761, opened 2024-09-04
chore: added missing use_oidc gcr-auth param

emaydeck-mozilla:
This value was defaulting to `false`, causing the orb to expect a service account key to be present.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -429,6 +429,7 @@ jobs:
           google-project-id: GCP_PROJECT_ID
           registry-url: <<parameters.registry-url>>
           service_account_email: GCP_SERVICE_ACCOUNT_EMAIL
+          use_oidc: true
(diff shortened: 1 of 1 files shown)


taddes (approved):
Good catch 👍 , will take note for the future in case there are future issues.

jrconlin:
We should probably add a comment pointing folk to where all these values are defined (and what the defaults are). I can see this coming back to haunt us eventually, so leaving a trail might be useful.

emaydeck-mozilla:
> We should probably add a comment pointing folk to where all these values are defined (and what the defaults are). I can see this coming back to haunt us eventually, so leaving a trail might be useful.

Done! https://github.com/mozilla-services/autopush-rs/pull/761/commits/3805e699969009037d495b5e9e1c8dbccb1dce27

jrconlin (approved):

PR #764, opened 2024-09-04
fix: minor fixes for GAR deploy CircleCI jobs

emaydeck-mozilla:
This now mounts the workspace so we can load the built image from that. Also, I fixed the image name in the `gcp-gcr/push-image` step.

This was tested successfully in https://app.circleci.com/pipelines/github/mozilla-services/autopush-rs/2873/workflows/9adebe2f-13e8-4e5d-a853-e78b7f6ca368

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -446,8 +446,6 @@ jobs:
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
               echo 'export GAR_TAG=master' >> $BASH_ENV
-            elif [ "${CIRCLE_BRANCH}" == "RRM-181" ]; then
-              echo 'export GAR_TAG=debug' >> $BASH_ENV
             elif [ ! -z "${CIRCLE_TAG}" ]; then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -436,6 +436,11 @@ jobs:
           use_oidc: true
           workload_identity_pool_id: GCP_OIDC_WIP_ID
           workload_identity_pool_provider_id: GCP_OIDC_WIP_PROVIDER_ID
+      - attach_workspace:
+          at: /tmp/cache
+      - run:
+          name: Restore Docker image cache
+          command: docker load -i /tmp/cache/docker.tar
```
taddes, line 443: Just a quick question: is there any concern of caching-related issues with docker image artifacts? I recall some issues in Merino when we'd make changes where there would be an out of date docker image that would break things in CI.  Not super specific, but curious of any possibilities of issues, since you know more about this than I do!

taddes (approved):
👍 Looks good! Just had a question about Docker image caching.

pjenvey (approved):

PR #769, opened 2024-09-25
feat: Add database tracking and report for Push Reliability

jrconlin:
This PR introduces tracking throughput for the database.

It also introduces the PushReliability reporting skeleton. This will be
fleshed out with full reporting later.

Closes: #[SYNC-4324](https://mozilla-hub.atlassian.net/browse/SYNC-4324)

[SYNC-4324]: https://mozilla-hub.atlassian.net/browse/SYNC-4324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


autoconnect/autoconnect-web/Cargo.toml
```diff
@@ -34,3 +34,5 @@ ctor.workspace = true
 tokio.workspace = true
 
 autoconnect_common = { workspace = true, features = ["test-support"] }
+
```
jrconlin, line 37: these features will be used in the next PR.

autoendpoint/src/settings.rs
```diff
@@ -310,7 +321,7 @@ mod tests {
     #[test]
     fn test_tracking_keys() -> ApiResult<()> {
         let settings = Settings{
-            tracking_keys: r#"["BLMymkOqvT6OZ1o9etCqV4jGPkvOXNz5FdBjsAR9zR5oeCV1x5CBKuSLTlHon-H_boHTzMtMoNHsAGDlDB6X7vI"]"#.to_owned(),
+            tracking_keys: r#"["BLMymkOqvT6OZ1o9etCqV4jGPkvOXNz5FdBjsAR9zR5oeCV1x5CBKuSLTlHon-H_boHTzMtMoNHsAGDlDB6X7"]"#.to_owned(),
```
jrconlin, line 324: Modified to check for padding stripping.

tests/integration/test_integration_all_rust.py
```diff
@@ -808,6 +814,41 @@ async def test_basic_delivery_with_vapid(
...
+    await registered_test_client.send_notification(data=uuid_data, vapid=vapid_info)
+    await registered_test_client.connect()
+    await registered_test_client.hello()
+    result = await registered_test_client.get_notification()
+
+    # the following presumes that only `salt` is padded.
+    clean_header = registered_test_client._crypto_key.replace('"', "").rstrip("=")
+    assert result["headers"]["encryption"] == clean_header
+    assert result["data"] == base64url_encode(uuid_data)
+    assert result["messageType"] == "notification"
```
taddes, line 849: We have an enum of `ClientMessageType` in the [async_push_test_client.py](https://github.com/mozilla-services/autopush-rs/blob/3cc5a565f550d482f83d1fc9534cac12bd4cbde0/tests/integration/async_push_test_client.py#L20) which you could add this as a variant, should you wish. Can alter this to simply be `MessageType` as well, since the "client" does double duty in these tests and isn't strictly the client...

tests/integration/test_integration_all_rust.py
```diff
@@ -184,6 +186,10 @@ def base64url_encode(value: bytes | str) -> str:
     message_table_name=MESSAGE_TABLE,
     human_logs="true",
     crypto_keys="[{}]".format(CRYPTO_KEY),
+    # convert to x692 format
+    tracking_keys="[{}]".format(
+        base64.urlsafe_b64encode((b"\4" + TRACKING_PUB_KEY.to_string())).decode()
+    ),
```
taddes, line 192:
```suggestion
    tracking_keys=f"[{base64.urlsafe_b64encode(b'\4' + TRACKING_PUB_KEY.to_string()).decode()}]",
```

taddes (changes requested):
👍 thanks for the work on this. Some thoughts on adjustments

autoendpoint/src/settings.rs
jrconlin: If we were decoding the key from base64, yes. We're currently not doing that so just tweaking the string values to ensure that they match (with and without the padding) should be fine. 

If we ever decide to decode these strings and do a byte comparison of the decoded pairs, then we would have to revisit this test, but that's out of scope for this PR.

autoendpoint/src/extractors/notification.rs
```diff
@@ -171,6 +172,12 @@ impl Notification {
         map.insert("ttl", serde_json::to_value(self.headers.ttl).unwrap());
         map.insert("topic", serde_json::to_value(&self.headers.topic).unwrap());
         map.insert("timestamp", serde_json::to_value(self.timestamp).unwrap());
+        if let Some(reliability_id) = &self.subscription.reliability_id {
+            map.insert(
+                "reliability_id",
+                serde_json::to_value(reliability_id).unwrap(),
```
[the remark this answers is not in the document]
jrconlin: Fair. I didn't want to modify a bunch of extra things, but I think you're right. I'll update `ApiErrorKind`

autoendpoint/src/extractors/notification.rs
```diff
@@ -160,25 +161,28 @@ impl Notification {
...
     /// we can't simply convert this notification type to that one and serialize
     /// via serde.
-    pub fn serialize_for_delivery(&self) -> HashMap<&'static str, serde_json::Value> {
+    pub fn serialize_for_delivery(&self) -> ApiResult<HashMap<&'static str, serde_json::Value>> {
         let mut map = HashMap::new();
 
         map.insert(
             "channelID",
-            serde_json::to_value(self.subscription.channel_id).unwrap(),
+            serde_json::to_value(self.subscription.channel_id)?,
```
pjenvey, line 169: I might be overthinking this but: usually killing unwraps is an improvement but with this moving the unwrap to the caller -- IMHO the original unwraps here were clearer that the individual serializations were infallible (probably because the line doing the unwrap hides the method details, somehow I don't even trust it even w/ the name `serialize_for_delivery`). Maybe renaming it to `serialize_for_delivery_infallible` would improve this, or better yet (IMHO) continue having the method not return a `Result` but without the unwraps w/ [what I did in this branch](https://github.com/mozilla-services/autopush-rs/compare/feat/SYNC-4324_track_db...feat/SYNC-4324_track_db-document-infallible)

autoendpoint/src/settings.rs
```diff
@@ -169,9 +169,11 @@ impl Settings {
     // public key, but that may not always be true.
     pub fn tracking_keys(&self) -> Vec<String> {
         let keys = &self.tracking_keys.replace(['"', ' '], "");
-        Self::read_list_from_str(keys, "Invalid AUTOEND_TRACKING_KEYS")
-            .map(|v| v.to_owned())
-            .collect()
+        let result = Self::read_list_from_str(keys, "Invalid AUTOEND_TRACKING_KEYS")
+            .map(|v| v.to_owned().replace("=", ""))
+            .collect();
+        trace!("🔍 tracking_keys: {:?}", result);
```
pjenvey, line 175:
```suggestion
        trace!("🔍 tracking_keys: {result:?}");
```

pjenvey (changes requested):

autoconnect/autoconnect-common/src/protocol.rs
```diff
@@ -80,11 +80,19 @@ impl FromStr for ClientMessage {
...
 #[derive(Debug, Deserialize)]
 pub struct ClientAck {
+    // The channel_id which received messages
     #[serde(rename = "channelID")]
     pub channel_id: Uuid,
+    // The corresponding version number for the message.
     pub version: String,
+    // The Reliability ID for the messages (if present)
+    #[serde(default)]
+    pub reliability_id: Option<String>,
```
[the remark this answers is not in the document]
jrconlin: Yeah, thanks! forgot to clean this up.

autoendpoint/src/extractors/notification.rs
jrconlin: Maybe I'm just being dense. 
`serialize_for_delivery()` gets unwrapped by `send_notification()` because `send_notification()` wants to return a `reqwest::Error` that gets checked by `route_notification` to see which error metric needs to be updated.

That doesn't seem super clear either. Plus, the function isn't infallible, since it very much returns an error that we pay attention to. 

`route_notification()` is part of the `Router` impl, which returns an `ApiResult`. Isn't the problem that we're basically eating the error?  Wouldn't it be better to add `ApiErrorKind::ReqwestError(_)` and then handle the error without doing any unwraps?

Going to update with that as a patch.

pjenvey: I think `serialize_for_delivery` is still infallible but yea, that definitely works (and I probably got too lost in the weeds)

autoendpoint/src/routers/webpush.rs
```diff
@@ -177,11 +179,11 @@ impl WebPushRouter {
         &self,
         notification: &Notification,
         node_id: &str,
-    ) -> Result<Response, reqwest::Error> {
+    ) -> ApiResult<Response> {
         let url = format!("{}/push/{}", node_id, notification.subscription.user.uaid);
-        let notification = notification.serialize_for_delivery();
+        let notification = notification.serialize_for_delivery().unwrap();
```
pjenvey, line 184: Forgot the last bit:
```suggestion
        let notification = notification.serialize_for_delivery()?;
```

pjenvey (changes requested):

autoendpoint/src/settings.rs
pjenvey: Forgot this one FYI (but I also don't care)

jrconlin: Meh, if i am fixing the other thing...

pjenvey (approved):

taddes (approved):
👍 great stuff, lgtm!

Issue #770, opened 2024-09-30
Autoendpoint ignores configuration file settings (including `vapid_aud`) when validating VAPID claims


ncalexan:
`validate_vapid_jwt` creates a new `Settings` object solely from the environment -- ignoring any configuration file settings -- at https://github.com/mozilla-services/autopush-rs/blob/9afbe422c6c46a9babca3a4fcb4ef180480c3790/autoendpoint/src/extractors/subscription.rs#L292

In practice, this object defaults to having `vapid_aud` as set [here](https://github.com/mozilla-services/autopush-rs/blob/9afbe422c6c46a9babca3a4fcb4ef180480c3790/autoendpoint/src/settings.rs#L73-L76). This is both incorrect and inefficient.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-4422)

ncalexan:
Worth a test that has a domain that _is not_ `push.services.mozilla.com`, I think.

jrconlin:
I'm not so sure.

The unit test submits a header that uses and aud of `example.com` where the settings uses `push.services.mozilla.com`. The actual validation of the `aud` is performed in `jsonwebtoken::decode` within [subscription.rs](https://github.com/mozilla-services/autopush-rs/blob/3cc5a565f550d482f83d1fc9534cac12bd4cbde0/autoendpoint/src/extractors/subscription.rs#L296-L304). I feel we can presume that their unit tests work correctly and that strings that do not match would be properly flagged. 

The local `aud` is specified as a setting, since any hard-coded value would be flagged in review. Is there a specific error case you're thinking of?

ncalexan:
I saw that one, but AFAICT the tests don't actually exercise the situation that I hit: a custom `vapid_aud` that _does_ validate a custom `aud`.  Shouldn't the unit tests ensure there isn't a future regression?

jrconlin:
I'm not 100% sold on the value, but I do understand the desire. Adding the test is cheap. I'll do a follow up PR that includes it.

jrconlin:
Ah! Sure enough, there was a bug, and it was a sneaky one too.
There was some legacy code that was double-checking the aud against the `endpoint_url` setting. 

Since the jwt does the validation already, no need to double check (and the associated error metrics would never fire anyway).

PR #772, opened 2024-10-02
bug: Do not double-check VAPID aud

jrconlin:
We were currently double-checking the VAPID aud via some legacy code which may produce invalid results if the declared `vapid_aud` setting and the `endpoint_url` did not match.

Closes: #770

### DIFF
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -1,6 +1,5 @@
 use std::borrow::Cow;
 use std::error::Error;
-use std::str::FromStr;
 
 use actix_web::{dev::Payload, web::Data, FromRequest, HttpRequest};
 use autopush_common::{
@@ -12,7 +11,6 @@ use cadence::{CountedExt, StatsdClient};
 use futures::{future::LocalBoxFuture, FutureExt};
 use jsonwebtoken::{Algorithm, DecodingKey, Validation};
 use openssl::hash::MessageDigest;
-use url::Url;
 use uuid::Uuid;
 
 use crate::error::{ApiError, ApiErrorKind, ApiResult};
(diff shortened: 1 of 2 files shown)


autoendpoint/src/extractors/subscription.rs
```diff
@@ -389,27 +389,6 @@ fn validate_vapid_jwt(
...
-        Err(_) => {
-            error!("Bad Aud: Invalid audience {:?}", &token_data.claims.aud);
-            metrics.clone().incr("notification.auth.bad_vapid.aud");
-            return Err(VapidError::InvalidAudience.into());
-        }
-    };
-
-    let domain = &settings.endpoint_url();
-
-    if domain != &aud {
```
pjenvey, line 403: I gather jsonwebtoken will now do an equivalent check against our `settings.vapid_aud` (via `validation.set_audience()`) -- however I don't see us assigning this new `AUTOEND__VAPID_AUD` setting [anywhere](https://github.com/mozilla-it/webservices-infra/blob/main/autopush/k8s/autoendpoint/templates/configmap.yaml), which I think means it's [defaulting to incorrect values](https://github.com/mozilla-services/autopush-rs/blob/master/autoendpoint/src/settings.rs#L73) (not `settings.endpoint_url()`). 

Do we even need `settings.vapid_aud` -- might we always set something along the lines of `validation.set_audience(&[settings.endpoint_url()])` instead?

pjenvey: and prod's basically `endpoint_url()` for reference [here](https://github.com/mozilla-it/webservices-infra/blob/fb647cb8d278b0ea8dde65d56a9452986678ab1e/autopush/k8s/autoendpoint/values-prod.yaml#L26)

pjenvey (approved):
I think the original check we did would be a little more lax with trailing `/`, since it compared `Url`s? But that doesn't matter, especially if we already have the jsonwebtoken's validation already occurring on prod (I can't remember if that's the case). Just noting it in case we see more validation errors or something.

PR #778, opened 2024-10-17
feat: Add redis based reliability reporting

jrconlin:
This adds a feature flag `reliable_report` that optionally enables Push
message reliablity reporting. The report is done in two parts.
The first part uses a Redis like storage system to note message states.
This will require a regularly run "cleanup" script to sweep for expired
messages and adust the current counts, as well log those states to some
sequential logging friendly storage (e.g. common logging or steamed to
a file). The clean-up script should be a singleton to prevent possible
race conditions.

The second component will write a record of the state transition
times for tracked messages to a storage system that is indexed by the
tracking_id. This will allow for more "in depth" analysis by external
tooling.

The idea being that reporting will be comprised of two parts:
One part which shows active states of messages (with a log of prior
states to show trends over time), and an optional "in-depth" record
that could be used to show things like length of time in storage,
overall success rates, survivability rates, etc.

This patch also:
* fixes a few typos
* changes several methods that should consume Notifications, actually  consume them.
* convert from `tracking_id` to `reliability_id`
* convert instance of specialized `Metrics` to generic Cadence (to make calls more consistent)
* adds a `RELIABLE_REPORT` flag to testing.

Closes: [SYNC-4327](https://mozilla-hub.atlassian.net/browse/SYNC-4327)

### DIFF
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
@@ -163,6 +163,31 @@ impl WebPushClient {
         Ok(smsgs)
     }
 
+    #[cfg(feature = "reliable_report")]
+    /// Record and transition the state for trackable messages.
+    async fn record_state(
+        &self,
+        messages: &mut Vec<Notification>,
+        state: autopush_common::reliability::PushReliabilityState,
+    ) {
+        // *Note* because `.map()` is sync
+        // we can't call the async func without additional hoops.
+        // I'm guessing that there's a more elegant way to do this, but this works for now.
+        for message in messages {
+            let expiry = message.timestamp + message.ttl;
+            message.reliable_state = self
+                .app_state
+                .reliability
+                .record(
+                    &message.reliability_id,
+                    state,
+                    &message.reliable_state,
+                    Some(expiry),
+                )
+                .await;
+        }
+    }
+
     /// Read a chunk (max count 10 returned) of Notifications from storage
     ///
     /// This alternates between reading Topic Notifications and Timestamp
@@ -186,10 +211,20 @@ impl WebPushClient {
         let topic_resp = if self.flags.include_topic {
             trace!("🗄️ WebPushClient::do_check_storage: fetch_topic_messages");
             // Get the most recent max 11 messages.
-            self.app_state
+            #[allow(unused_mut)]
+            let mut messages = self
+                .app_state
                 .db
                 .fetch_topic_messages(&self.uaid, 11)
-                .await?
+                .await?;
+            #[cfg(feature = "reliable_report")]
+            // Since we pulled these from storage, mark them as "retrieved"
+            self.record_state(
+                &mut messages.messages,
+                autopush_common::reliability::PushReliabilityState::Retreived,
+            )
+            .await;
+            messages
         } else {
             Default::default()
         };
@@ -226,7 +261,8 @@ impl WebPushClient {
             "🗄️ WebPushClient::do_check_storage: fetch_timestamp_messages timestamp: {:?}",
             timestamp
         );
-        let timestamp_resp = self
+        #[allow(unused_mut)]
+        let mut timestamp_resp = self
             .app_state
             .db
             .fetch_timestamp_messages(&self.uaid, timestamp, 10)
@@ -244,6 +280,13 @@ impl WebPushClient {
                 )
                 .with_tag("topic", "false")
                 .send();
+            #[cfg(feature = "reliable_report")]
+            // Since we pulled these from storage, mark them as "retrieved"
+            self.record_state(
+                &mut timestamp_resp.messages,
+                autopush_common::reliability::PushReliabilityState::Retreived,
+            )
+            .await;
         }
 
         Ok(CheckStorageResponse {
--- a/scripts/setup_bt.sh
+++ b/scripts/setup_bt.sh
@@ -12,11 +12,15 @@ TABLE_NAME=${TABLE_NAME:-"autopush"}
 MESSAGE_FAMILY=${MESSAGE_FAMILY:-"message"}
 MESSAGE_TOPIC_FAMILY=${MESSAGE_TOPIC_FAMILY:-"message_topic"}
 ROUTER_FAMILY=${ROUTER_FAMILY:-"router"}
+RELIABILITY_FAMILY=${RELIABILITY_FAMILY:-"reliability"}
 
 cbt -project $PROJECT -instance $INSTANCE createtable $TABLE_NAME
 cbt -project $PROJECT -instance $INSTANCE createfamily $TABLE_NAME $MESSAGE_FAMILY
 cbt -project $PROJECT -instance $INSTANCE createfamily $TABLE_NAME $MESSAGE_TOPIC_FAMILY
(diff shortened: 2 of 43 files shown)


pjenvey (dismissed):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
```diff
@@ -163,6 +163,31 @@ impl WebPushClient {
...
+    #[cfg(feature = "reliable_report")]
+    /// Record and transition the state for trackable messages.
+    async fn record_state(
+        &self,
+        messages: &mut Vec<Notification>,
+        state: autopush_common::reliability::PushReliabilityState,
+    ) {
+        // *Note* because `.map()` is sync
+        // we can't call the async func without additional hoops.
+        // I'm guessing that there's a more elegant way to do this, but this works for now.
```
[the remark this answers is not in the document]
jrconlin: Perhaps, but when I was writing this, I had to keep reminding myself of the sync nature of `.map()`. I'm guessing that others (including me) might forget that in the future.

jrconlin:
Well, that was less than entertaining, but at least educational.

Makefile
```diff
@@ -2,7 +2,19 @@ SHELL := /bin/sh
...
+# The following is taken from [the makefile manual](https://www.gnu.org/software/make/manual/make.html#Syntax-of-Functions)
+NOOP :=
+SPACE :=$(NOOP) $(NOOP)
+PYTEST_AND :=" and "
+# specify the tests to NOT run
+# NOTE: passing `-m not` to pytest will cause an error be sure at least one of these is set.
+NEG_PYTEST_ARGS ?= $(if $(SKIP_SENTRY),sentry) $(if $(TEST_STUB),,stub) $(if $(TEST_RELIABILITY),,reliable_report) # Stub tests do not work in CI
+# Now compose the pytest args by gluing the above items together
+NEG_PYTEST_ARGS := "not $(subst $(SPACE), and not ,$(strip $(NEG_PYTEST_ARGS)))"
+PYTEST_ARGS := -m $(NEG_PYTEST_ARGS)
```
Trinaa, line 17: I hadn't realized that multiple -m flags are not additive TIL. My thought to overcome this lookes like this, it's less efficient but imo easier to understand. 

```
PYTEST_MARKERS := $(if $(SKIP_SENTRY),not sentry,)
PYTEST_MARKERS += $(if $(TEST_STUB),,$(if $(strip $(PYTEST_MARKERS)),and ,)not stub)
PYTEST_MARKERS += $(if $(TEST_RELIABILITY),,$(if $(strip $(PYTEST_MARKERS)),and ,)not reliable_report)
PYTEST_ARGS ?= $(if $(strip $(PYTEST_MARKERS)),-m "$(strip $(PYTEST_MARKERS))",)
```

That said, perhaps controlling markers with environment variables is overall a poor choice and we should perhaps stick to setting their values in the configuration, as @taddes suggested. We could add the line `addopts = -m "not stub and not reliable_report"` under `markers` in the [pyproject.toml](https://github.com/mozilla-services/autopush-rs/blob/master/tests/pyproject.toml#L100-L103)


**Also some reminders:**

nit: please add the reliable_report marker to the list in [pyproject.toml](https://github.com/mozilla-services/autopush-rs/blob/master/tests/pyproject.toml#L100-L103)
image

nit: please include the reliable_report marker in the [testing.md documentation](https://github.com/mozilla-services/autopush-rs/blob/master/docs/src/testing.md?plain=1#L91-L94)

jrconlin: I'm fine with less efficient / more readable. I'm not sure adding them to `pyproject.toml` is a good idea, for a couple of reasons. I would prefer if test configuration was done at one location, so we don't have to chase around if we want to change things later. It would be nice to have some sort of "combination" function like `standard = "not stub and not reliable_report and not sentry"` but I can see that getting equally complex in the future and I'd like to not make test configurations Turing complete.

jrconlin: Related: https://github.com/pytest-dev/pytest/issues/7229

Makefile
```diff
@@ -2,7 +2,21 @@ SHELL := /bin/sh
...
 TESTS_DIR := tests
 TEST_RESULTS_DIR ?= workspace/test-results
-PYTEST_ARGS ?= $(if $(SKIP_SENTRY),-m "not sentry") $(if $(TEST_STUB),,-m "not stub") # Stub tests do not work in CI
+# Markers are "clever", and work by including multiple markers with " and ".
+# This uses makefile magic to construct the set of items.
+# technically `$(eval )` would work here, but I've not had it work consistently.
+# The following is taken from [the makefile manual](https://www.gnu.org/software/make/manual/make.html#Syntax-of-Functions)
+NOOP :=
+SPACE :=$(NOOP) $(NOOP)
+PYTEST_AND :=" and "
```
Trinaa, line 11: I think these are no longer needed?

Trinaa (commented):
The integration test changes LGTM!

pjenvey (changes requested):
I have a few nitpicks and cleanup to the recording code in this branch:

https://github.com/mozilla-services/autopush-rs/compare/feat/SYNC-4327_redis...feat/SYNC-4327_redis-pjenvey

I'm still thinking about some of the state changes we discussed last week but that can happen in a later PR.

scripts/setup_bt.sh
```diff
@@ -12,11 +12,15 @@ TABLE_NAME=${TABLE_NAME:-"autopush"}
...
 cbt -project $PROJECT -instance $INSTANCE createtable $TABLE_NAME
 cbt -project $PROJECT -instance $INSTANCE createfamily $TABLE_NAME $MESSAGE_FAMILY
 cbt -project $PROJECT -instance $INSTANCE createfamily $TABLE_NAME $MESSAGE_TOPIC_FAMILY
 cbt -project $PROJECT -instance $INSTANCE createfamily $TABLE_NAME $ROUTER_FAMILY
+cbt -project $PROJECT -instance $INSTANCE createfamily $TABLE_NAME $RELIABILITY_FAMILY
+
 cbt -project $PROJECT -instance $INSTANCE setgcpolicy $TABLE_NAME $MESSAGE_FAMILY "maxage=1s or maxversions=1"
 cbt -project $PROJECT -instance $INSTANCE setgcpolicy $TABLE_NAME $MESSAGE_TOPIC_FAMILY "maxage=1s or maxversions=1"
 cbt -project $PROJECT -instance $INSTANCE setgcpolicy $TABLE_NAME $ROUTER_FAMILY maxversions=1
+cbt -project $PROJECT -instance $INSTANCE setgcpolicy $TABLE_NAME $MESSAGE_TOPIC_FAMILY "maxage=60d or maxversions=1"
```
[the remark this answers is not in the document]
jrconlin: This part is a bit confusing, and I may not be clarifying it, but here's the thinking behind it:

`record_reliability()` [sets](https://github.com/mozilla-services/autopush-rs/blob/6aaf27fc00311f70fa7f044f01083b0a8c39bdf1/autoendpoint/src/extractors/notification.rs#L235) the cell's `expr` = the message's timestamp of receipt plus the TTL header value. We want to retain the message for `60d` past that, so we specify the `setgcpolicy` here. That's different from what we do elsewhere, where we set the cell`s expr pre-emptively to the receipt timestamp + offset + TTL. My thinking is that the reliability timestamps are probably a bit more fungible when it comes to retention, so altering the gc policy makes more sense.

pjenvey (approved):

PR #789, opened 2024-10-25
tests: Add automated notification tests.

b4handjr:
This adds a suite of automated notification tests that simply compare a screenshot of before and after a notification has been sent. It uses docker, pytest and selenium, and generate an XML report.

Using the `makefile` we can pass in which servers we want to use: `make ENV="dev" notification-test`

Closes: https://mozilla-hub.atlassian.net/browse/SYNC-4371

Base Screenshot: 
base_screenshot

Screenshot with notification: 
screenshot


tests/notification/conftest.py
```diff
@@ -0,0 +1,38 @@
+"""Conftest file for notification tests."""
+
+from typing import Any
```
jrconlin, line 3: nit: `Any` is the duct tape for python typing. We really should try to identify the types for cleanliness, security, and to hint to the IDEs for code completion.

FWIW:
```python
from selenium.webdriver.firefox.webdriver import WebDriver # for `selenium`
from pytest import Config, Parser # for `pytestconfig` & `firefox_options`
```

tests/notification/test_notification.py
```diff
@@ -0,0 +1,118 @@
+"""Module containing the Notification test files for autopush-rs."""
+
+import logging
+import time
+from typing import Any
```
jrconlin, line 5: nit: FWIW:

```python
from PIL.Image import Image # for `setup_page`
from pytest import Testdir # for `tmpdir`
from selenium.webdriver.firefox.webdriver import WebDriver # for `selenium`
```

tests/notification/dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    libdrm2 \
+    libgbm1 \
+    libatspi2.0-0 \
+    bzip2 \
+    libglib2.0-0 \
+    libnss3 \
+    libgconf-2-4 \
+    libfontconfig1 \
+    libdbus-glib-1-2 \
+    && apt-get clean
```
jrconlin, line 42: add a `-y` just in case something asks? Might help prevent a hang or timeout.

tests/notification/test_notification.py
```diff
@@ -0,0 +1,118 @@
...
+    """Directory to store the screenshots for testing."""
+    return tmpdir.mkdir("images")
+
+
+@pytest.fixture(autouse=True)
+def setup_page(selenium: Any, images_dir: str) -> ImageGrab:
+    """Fixture to setup the test page and take the base screenshot."""
+    selenium.get("localhost:8201")
+    selenium.find_element(By.CSS_SELECTOR, ".container").click()
+    time.sleep(5)  # wait a bit to take the base screenshot
```
pjenvey, line 24: Do you think a selenium explicit wait could potentially replace this sleep? I'm fine with the sleep for now (I'm guessing if the explicit wait is possible it might take some effort to determine which element to wait on) but if so, might be worth including a comment about that potential improvement

tests/notification/conftest.py
b4handjr: I concur, I didn't put much effort here as these tests are pretty low risk, but I can add all the types if you would like that.

tests/notification/test_notification.py
b4handjr: This is outside of the selenium context, I am just trying to prevent getting a screenshot with no notification by taking it immediately after I hit click because we know computers will do it quite fast and that's unrealistic.

tests/notification/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  server:
+    image: "firefoxtesteng/autopush-e2e-test"
```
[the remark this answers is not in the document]
b4handjr: They could be viewed as both, we could combine them. And that image you're referring too is quite old as seen by the `firefoxtesteng` docker repo.

Makefile
```diff
@@ -41,6 +42,13 @@ integration-test:
 		--junit-xml=$(TEST_RESULTS_DIR)/integration_test_results.xml \
 		-v $(PYTEST_ARGS)
 
+notification-test:
+	$(DOCKER_COMPOSE) -f $(NOTIFICATION_TEST_DIR)/docker-compose.yml build
+	$(DOCKER_COMPOSE) -f $(NOTIFICATION_TEST_DIR)/docker-compose.yml up -d server
+	ENV=$(ENV) $(DOCKER_COMPOSE) -f $(NOTIFICATION_TEST_DIR)/docker-compose.yml run -it --name notification-tests tests
+	docker cp notification-tests:/code/notification-tests.xml $(NOTIFICATION_TEST_DIR)
+	docker rm notification-tests
+
```
Trinaa, line 51: The tests aren't passing for me locally... I get the feeling that part of the problem is we don't incorporate poetry in this command. Could we try something like:

```suggestion
notification-test: $(INSTALL_STAMP)
	$(DOCKER_COMPOSE) -f $(NOTIFICATION_TEST_DIR)/docker-compose.yml build
	$(DOCKER_COMPOSE) -f $(NOTIFICATION_TEST_DIR)/docker-compose.yml up -d server
	$(POETRY) run docker compose -f $(NOTIFICATION_TEST_DIR)/docker-compose.yml run -it --name notification-tests tests
	docker cp notification-tests:/code/notification-tests.xml $(NOTIFICATION_TEST_DIR)
	docker rm notification-tests

```

Errors:
image

tests/notification/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[tool.poetry]
```
Trinaa, line 1: We should be trying to use the central[ pyproject.toml](https://github.com/mozilla-services/autopush-rs/blob/master/tests/pyproject.toml) instead of having an independent one.

b4handjr: This is mainly because the tests are run inside of docker. I could use that file, ill have to make a group just for these tests.

Makefile
b4handjr: They have to be run inside of docker, see the `README`.

b4handjr: Oh wait, I think those steps there are wrong, let me fix.

b4handjr: Please try again, I think I fixed it

b4handjr:
@Trinaa I removed the added poetry files and added the needed dependencies to a group. Had to do some trickery with the docker image but it all works

Makefile
```diff
@@ -75,7 +83,6 @@ lint:
 	$(POETRY) install --no-root
 	$(POETRY) run isort --sp $(PYPROJECT_TOML) -c $(TESTS_DIR)
 	$(POETRY) run black --quiet --diff --config $(PYPROJECT_TOML) --check $(TESTS_DIR)
-	$(POETRY) run flake8 --config $(FLAKE8_CONFIG) $(TESTS_DIR)
```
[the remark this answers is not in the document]
jrconlin: FWIW, I'm ok with dropping flake8. It's redundant and often conflicts with black's formatting. 
if anything, we're probably not excepting [all the recommended styles](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#pycodestyle).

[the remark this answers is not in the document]
Trinaa: Perhaps we drop it in a different PR then and include removing the Makefile commands and FLAKE8_CONFIG file.

Makefile
Trinaa: Okay I've cleared the amd/arm issue see https://github.com/mozilla-services/autopush-rs/pull/789/commits/0e0372be76bc25c8418045f272fefcd46289f3ea

But still the tests don't pass, new errors:
image

b4handjr:
@Trinaa so this doesn't work? `ENV="dev" docker compose run -it tests`

b4handjr:
Okay I think I got it. I forgot to actually pass the env to the docker command at run time.

tests/notification/README.md
```diff
@@ -0,0 +1,33 @@
+# Autopush Notification Integration Tests
```
Trinaa, line 1: nit: perhaps this content should be in [testing.md](https://github.com/mozilla-services/autopush-rs/blob/master/docs/src/testing.md)

cc: @taddes

tests/notification/Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM python:3-slim
+
+ENV POETRY_HOME="/opt/poetry" \
+    POETRY_VIRTUALENVS_IN_PROJECT=1 \
+    POETRY_NO_INTERACTION=1 \
+    GECKODRIVER="0.35.0"
+
+ENV PATH="$POETRY_HOME/bin:$PATH"
+
+ENV ENV="dev"
```
Trinaa, line 10: nit: 'ENV' is very generic, perhaps `NOTIFICATION_TEST_ENV` or `TEST_ENV` instead?

Trinaa (commented):
The tests are still not passing for me, I have the following errors: 
image

However this isn't a blocker for me as long as they pass on the CD environment. I would like to merge this and move onto building and publishing the docker image so that SRE can integrate these tests in CD. 

The sticking point for me, why I'm not approving, is the flake8 removal. I don't think this should be part of the PR scope.

Makefile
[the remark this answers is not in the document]
b4handjr: Oh this is a mistake, I must've accidentally added this

b4handjr:
> The tests are still not passing for me, I have the following errors: image
> 
> However this isn't a blocker for me as long as they pass on the CD environment. I would like to merge this and move onto building and publishing the docker image so that SRE can integrate these tests in CD.
> 
> The sticking point for me, why I'm not approving, is the flake8 removal. I don't think this should be part of the PR scope.

I'll fix the flake8 removal, that was a mistake. 

This error looks like Firefox is not starting. Is the docker build happening okay?

Trinaa (approved):
Thank you for fixing the flake8 line! Lets merge and keep things moving.

PR #795, opened 2024-11-01
feat: Delete ACK'd messages from storage to prevent some dupes

jrconlin:
Moved the MessageID from autoendpoint to autopush-common so that I can use the `decrypt()` and `sort_key()` functions.

The stub is there to keep the file change count low.

Closes: ~SYNC-4466~ PUSH-330


jrconlin:
This is in response to https://bugzilla.mozilla.org/show_bug.cgi?id=1925631

PR #798, opened 2024-11-12
refactor(integration): replace pytest.skip with pytest markers

Trinaa:
Proposing to use [pytest markers](https://docs.pytest.org/en/stable/example/markers.html) instead of `pytest.skip + environment variables` for filtering integration tests. This will:

- Allow for more accurate skip rate reporting in [test metrics](https://docs.google.com/spreadsheets/d/1abjtg2e-PHm8JDP5A629KFVA-eBc8I5VtFPcz_UCDQs/edit?gid=1981049601#gid=1981049601)
- Will reduce cyclomatic complexity in the test cases
- Will likely scale better if more filters are needed in future

As a workaround, I've added continued support for the environment variables in the Makefile. 

Note: Tests that are filtered out are noted as 'deselected' in cmd output

image
image

Closes: #SYNC-4490

An alternative could be introducing a marker like `@pytest.mark.no_ci`. We could then leave the `pytest.skip + environment variables` in place and still report skip rates accurately, but I find this less ideal because we would have multiple mechanisms for filtering tests. This complexity seems unnecessary. 

Add-on: I removed the `no-root` option in the Makefile because we use package-mode = false and should no longer need it according to this warning.
image

### DIFF
--- a/tests/integration/test_integration_all_rust.py
+++ b/tests/integration/test_integration_all_rust.py
@@ -634,12 +634,12 @@ async def fixture_registered_test_client(ws_config) -> AsyncGenerator:
     log.debug("🐍 Test Client Disconnected")
 
 
+@pytest.mark.sentry
+# Selectively include or exclude this test based on `mark` See
+# https://docs.pytest.org/en/7.1.x/example/markers.html
 @pytest.mark.parametrize("fixture_max_conn_logs", [4], indirect=True)
 async def test_sentry_output_autoconnect(test_client: AsyncPushTestClient) -> None:
     """Test sentry output for autoconnect."""
-    if os.getenv("SKIP_SENTRY"):
-        log.debug("Skipping test_sentry_output_autoconnect")
-        pytest.skip("Skipping test_sentry_output_autoconnect")
     # Ensure bad data doesn't throw errors
     await test_client.connect()
     await test_client.hello()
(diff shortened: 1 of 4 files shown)


tests/integration/test_integration_all_rust.py
```diff
@@ -634,12 +634,10 @@ async def fixture_registered_test_client(ws_config) -> AsyncGenerator:
     log.debug("🐍 Test Client Disconnected")
 
 
+@pytest.mark.sentry
```
jrconlin, line 637: Since the team was unfamiliar with this feature of pytest, let's add a quick document reference so that others can also benefit. We only really need to add this once. We could add it before the first test or to the `pyproject.toml` file. 

(e.g. 
```suggestion
# Selectively include or exclude this test based on `mark` See
# https://docs.pytest.org/en/7.1.x/example/markers.html
@pytest.mark.sentry
```

jrconlin (changes requested):

jrconlin (approved):

taddes (approved):
Thanks for doing this @Trinaa ! Not to be too picky, but to pick up from @jrconlin 's comment, I thought it might not hurt to also add a line in the test docs https://github.com/mozilla-services/autopush-rs/blob/master/docs/src/testing.md for clarity. I'll leave it up to you, but I kind of like us being in the habit of adding to the core service docs as well when we have a procedural or technical norm we want to establish.

taddes (approved):
🚀

PR #803, opened 2024-11-20
bug: Address insecure sub header fetch. 

jrconlin:
When insecurely getting the `sub` for tracking and errors, we were not
properly decoding. This was resulting in a VAPID error being generated.
This should not impact processing, but was generating a lot of logging
messages.

* FCM would reject TTLS > that 4 weeks, which is less than
our 30 day max. Added a specific filter for that.

* fixed some spelling mistakes.

Closes: [SYNC-4514](https://mozilla-hub.atlassian.net/browse/SYNC-4514)

[SYNC-4514]: https://mozilla-hub.atlassian.net/browse/SYNC-4514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -332,6 +332,9 @@ fn validate_vapid_jwt(
             jsonwebtoken::errors::ErrorKind::InvalidAudience => {
                 return Err(VapidError::InvalidAudience.into());
             }
+            jsonwebtoken::errors::ErrorKind::MissingRequiredClaim(e) => {
+                return Err(VapidError::InvalidVapid(format!("Missing required {}", e)).into());
+            }
             _ => {
                 // Attempt to match up the majority of ErrorKind variants.
                 // The third-party errors all defer to the source, so we can
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -87,7 +87,7 @@ impl FromRequest for Subscription {
             if let Some(ref header) = vapid {
                 let sub = header
                     .vapid
-                    .sub()
+                    .insecure_sub()
                     .map_err(|e: VapidError| {
                         // Capture the type of error and add it to metrics.
                         let mut tags = Tags::default();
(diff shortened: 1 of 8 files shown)


taddes (changes requested):
LGTM, thanks for doing this! Just a few minor things.

autoendpoint/src/extractors/subscription.rs
```diff
@@ -87,7 +87,7 @@ impl FromRequest for Subscription {
             if let Some(ref header) = vapid {
                 let sub = header
                     .vapid
-                    .sub()
+                    .insecure_sub()
                     .map_err(|e: VapidError| {
```
pjenvey, line 91: `sub`/`insecure_sub` logs its errors, and here we emit a metric when they fail (in map_err). However we ultimately eat the error via `unwrap_or_default()` a few lines below: so I'm not seeing how this issue would lead to messages being rejected as invalid?

pjenvey (changes requested):

autoendpoint/src/extractors/subscription.rs
jrconlin: Honestly, me neither. Part of the goal here is to just remove them as noise from our logs, as well as fix grabbing the right value.

autoendpoint/src/routers/webpush.rs
```diff
@@ -209,12 +211,14 @@ impl WebPushRouter {
...
-                            .vapid
-                            .as_ref()
-                            .map(|vapid| vapid.vapid.claims().map(|c| c.sub).unwrap_or_default()),
+                        // try to extract the `sub` from the VAPID claims.
+                        notification.subscription.vapid.as_ref().map(|vapid| {
+                            vapid
+                                .vapid
+                                .claims()
+                                .map(|c| c.sub.unwrap_or_default())
+                                .unwrap_or_default()
```
pjenvey, line 220:
```suggestion
                                .ok()
                                .and_then(|c| c.sub)
                                .unwrap_or_default()
```

autopush-common/src/db/bigtable/mod.rs
```diff
@@ -166,9 +166,25 @@ impl TryFrom<&str> for BigTableDbSettings {
...
+                "default".to_owned()
+            } else {
+                me.app_profile_id
+            },
+            // Python uses a string based on a ms timestamp, but since we deploy A LOT of these at the same time,
+            // I'm wondering if there's a risk for collision. Thus I'm using higher entropy UUIDv4.
+            // Forcing hyphenation to stay consistent. (Hopefully there's no string length limits, but since
+            // we aren't getting direction and the emulator doesn't care, YOLO!)
+            uuid::Uuid::new_v4().hyphenated()
+        );
```
pjenvey, line 187: let's revert this back

pjenvey (changes requested):

pjenvey (approved):

PR #808, opened 2024-12-09
feat: Call out the "RESOURCE_EXHAUSTED" error metric

jrconlin:
Introduces notification.bridge.error.resource_exhausted

Closes: SYNC-4551


autoendpoint/src/routers/fcm/error.rs
```diff
@@ -91,6 +89,9 @@ impl ReportableError for FcmError {
             FcmError::InvalidAppId(_) | FcmError::NoAppId => {
                 Some("notification.bridge.error.fcm.badappid")
             }
+            FcmError::Upstream { status, .. } if status == "RESOURCE_EXHAUSTED" => {
+                Some("notification.bridge.error.fcm.resource_exhausted")
+            }
```
pjenvey, line 94: We still have this distinction between how we handle router errors (described in [SYNC-3695](https://mozilla-hub.atlassian.net/browse/SYNC-3695), not sure why we closed this one out?)

The common `handle_error` uses the same metric name for all `RouterErrors`: `notification.bridge.error` w/ tags (most importantly `platform: fcmv1`, `reason: eg_resource_exhausted`). It doesn't handle any more specific kind of errors (like `FcmError`), but even outside of it, I think we should be emitting metrics with the same name and as many as the same tags as possible -- via the `ReportableError` is fine but we probably can't emit `app_id` here without a bunch of other annoying work. That way they're all plotted on the same dashboard panel.

autoendpoint/src/routers/mod.rs
```diff
@@ -202,11 +202,14 @@ impl ReportableError for RouterError {
     fn metric_label(&self) -> Option<&'static str> {
         // NOTE: Some metrics are emitted for other Errors via handle_error
         // callbacks, whereas some are emitted via this method. These 2 should
-        // be consoliated: https://mozilla-hub.atlassian.net/browse/SYNC-3695
+        // be consolidated: https://mozilla-hub.atlassian.net/browse/SYNC-3695
         match self {
             RouterError::Apns(e) => e.metric_label(),
             RouterError::Fcm(e) => e.metric_label(),
             RouterError::TooMuchData(_) => Some("notification.bridge.error.too_much_data"),
+            RouterError::Upstream { status, .. } if status == "RESOURCE_EXHAUSTED" => {
```
pjenvey, line 210: Can we kill `RouterError::Upstream` entirely now (is it still used anywhere)?

pjenvey (changes requested):

autoendpoint/src/routers/fcm/error.rs
pjenvey: This applies to `notification.bridge.error.fcm.badappid` above as well -- we don't plot this one anywhere because of its different naming

autoendpoint/src/routers/fcm/error.rs
```diff
@@ -86,11 +86,8 @@ impl ReportableError for FcmError {
 
     fn metric_label(&self) -> Option<&'static str> {
         match &self {
-            FcmError::InvalidAppId(_) | FcmError::NoAppId => {
-                Some("notification.bridge.error.fcm.badappid")
-            }
-            FcmError::Upstream { status, .. } if status == "RESOURCE_EXHAUSTED" => {
-                Some("notification.bridge.error.fcm.resource_exhausted")
+            FcmError::InvalidAppId(_) | FcmError::NoAppId | FcmError::Upstream { .. } => {
```
pjenvey, line 89: I don't think we want `Upstream` here with `handle_error` already emitting a metric for it: as we'd get a metric emitted twice for it.

I think all `FcmErrors` will be passed through `handle_error` (or going to sentry, which `InvalidAppId`/`NoAppId` are set to do)? If so I'd argue we shouldn't return any metric labels here at all (for now)

jrconlin: Ah, you're right, I forgot that `handle_error()` records the metric and then returns the original error for later processing.
That said, we do have some duplication already. I'll add a comment so I can remember that later.
I note that `RouterError::TooMuchData(_)` produces `notification.bridge.error {too_much_data}` [🔗](https://github.com/mozilla-services/autopush-rs/blob/559b6537eb8b70bed4dd702403af8cc0dde4f57b/autoendpoint/src/routers/common.rs#L120-L121)  and `notification.bridge.error.too_much_data` [🔗](https://github.com/mozilla-services/autopush-rs/blob/559b6537eb8b70bed4dd702403af8cc0dde4f57b/autoendpoint/src/routers/mod.rs#L195-L196) Should I remove that as well so that we only have one? (I can add that as a different PR.)
I'll also note that we might want to have a `ReportableError::metric_tags()` function similar to how we have `extras()` to provide the additional information, since that seems to be a pattern we're hitting.

pjenvey: I'm fine with leaving it for now. 

We already have `ReportableError::tags` (it defaults to returning nothing, so you just have to implement it to override)

.cargo/audit.toml
```diff
@@ -1,4 +1,5 @@
 [advisories]
 ignore = [
-    "RUSTSEC-2021-0124", # Bound by tokio restrictions, cargo, reqwest, hyper...    
+    "RUSTSEC-2021-0124", # Bound by tokio restrictions, cargo, reqwest, hyper...
+    "RUSTSEC-2024-0402", # Bound by config 0.14, h2 0.4, reqwest 0.12-
```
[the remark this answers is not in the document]
pjenvey: In fact neither is RUSTSEC-2021-0124 (am I doing this wrong?)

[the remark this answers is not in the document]
jrconlin: Sigh, I had removed that, but from the other branch. 
Let me merge those and fix this up.

Yes, for a moment, that CVE was not being handled, but the updates fixed it.

autoendpoint/src/routers/common.rs
```diff
@@ -136,6 +122,15 @@ pub async fn handle_error(
                 error.errno(),
             );
         }
+        RouterError::Fcm(FcmError::Upstream { status, .. }) => incr_error_metric(
+            metrics,
+            platform,
+            app_id,
+            &format!("upstream_{}", status),
```
pjenvey, line 129: Sorry, one last thing: `status` here is either `FcmErrorResponse::status` or `StatusCode::to_string()`. I know the latter will contain whitespace (e.g. "200 OK") and the former probably does too? Which makes this an invalid metric tag value

pjenvey (changes requested):

autoendpoint/src/routers/common.rs
jrconlin: Sigh, you're right, with the various back & forth and merges `status` is not the same. I'll modify.

jrconlin: Wait, no, That is right.

The `status` here is from `Upstream{status: String, ..}` and reflects the returned `FcmErrorResponse.status`, which is a set of enums. There _was_ as similar error, though, and I've corrected it. I've also added some comments because this was a source of confusion.

autoendpoint/src/routers/fcm/client.rs
```diff
@@ -153,15 +153,20 @@ impl FcmClient {
...
                 }
-                (status, None) => RouterError::Upstream {
-                    status: status.to_string(),
-                    message: "Unknown reason".to_string(),
-                },
+                // In this case, we've gotten an error, but FCM hasn't returned a body.
+                // (This may happen in the case where FCM terminates the connection abruptly
+                // or a similar event.) Treat that as an INTERNAL error.
+                (_, None) => FcmError::Upstream {
+                    error_code: "INTERNAL".to_string(),
```
pjenvey, line 166: nit: how about distinguishing between their `INTERNAL` type w/ the response's status code (e.g. `reason: UNKNOWN_502`)
```suggestion
                    error_code: format!("UNKNOWN_{}", status.as_str()),
```

pjenvey (dismissed):

autoendpoint/src/routers/fcm/client.rs
jrconlin: I'm fine using "UNKNOWN" but I don't think adding the status code will get us much more. 
I'm going to add a `warn!()` logger message to report the error, status and the raw response, since that will probably be more useful when we see a bunch of these, since they should be (hopefully) very rare.

pjenvey (approved):

PR #812, opened 2024-12-18
chore(ci): Add capability to build and push test images to GCR.

b4handjr:
Adds the steps needed to build and push our containerized tests to GCR.

[SYNC-4559](https://mozilla-hub.atlassian.net/browse/SYNC-4559)

[SYNC-4559]: https://mozilla-hub.atlassian.net/browse/SYNC-559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -344,6 +344,7 @@ jobs:
   build-integration-test-container:
     docker:
       - image: cimg/base:2024.06
+    resource_class: large
     parameters:
       image:
         type: string
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -479,12 +479,10 @@ workflows:
           filters:
             tags:
               only: /.*/
-            branches:
-              only: master
       - build-test-container:
-          name: Build Notification Test Image
-          image: autopush-notification-tests
-          path: notification
+          name: Build End-To-End Test Image
+          image: autopush-end-to-end-tests
+          path: end-to-end
           filters:
             tags:
               only: /.*/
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -357,6 +340,7 @@ jobs:
         default: build
       image:
         type: string
+        default: docker
```
emaydeck-mozilla, line 343: This probably isn't necessary since we're not specifying a default for either of the build jobs and we're passing this parameter from every job that calls it.

emaydeck-mozilla (approved):
This looks good to me. If you want, you can test this out by having it temporarily filter against your branch instead of `master` in the job definitions. Once your push up a commit in your branch, CircleCI should try to build and deploy the images and I can confirm whether or not they were successfully pushed to GAR.

jrconlin (approved):

.circleci/config.yml
```diff
@@ -470,18 +472,31 @@ workflows:
...
+          name: Build Load Test Image
+          image: autopush-load-tests
+          path: load
           filters:
             tags:
               only: /.*/
-
-      - build-integration-test:
+            branches:
+              only: master
```
Trinaa, line 483: This is a departure from the way we have the PR builds setup currently. We usually build the load-test image in PR builds without deploying, so developers can be signaled before they merge if they have introduced a breaking change. I suggest to remove this filter and perhaps the one for the notification test container as well for the same reason.

Trinaa (dismissed):
Pre-approving so that you are not blocked over the holidays @b4handjr.

I suggest to take this opportunity to rename the 'notification' tests, which is not a commonly recognized test category or level , to 'end-to-end' tests.

PR #815, opened 2024-12-30
Feat: Add urgency support

p1gp1g:
It allows clients to define the minimum urgency of messages they wish to receive. For example, a client with low battery can request a high urgency message only until their battery is good again.

### DIFF
--- a/autoconnect/autoconnect-common/src/protocol.rs
+++ b/autoconnect/autoconnect-common/src/protocol.rs
@@ -14,6 +14,9 @@ use uuid::Uuid;
 
 use autopush_common::notification::Notification;
 
+#[cfg(feature = "urgency")]
+use autopush_common::db::Urgency;
+
 #[derive(Debug, Eq, PartialEq, Serialize)]
 #[serde(untagged)]
 pub enum BroadcastValue {
@@ -64,6 +67,11 @@ pub enum ClientMessage {
     Nack {
         code: Option<i32>,
         version: String,
+    },
+
+    #[cfg(feature = "urgency")]
+    Urgency {
+        min: Urgency,
     },
 
     Ping,
@@ -122,6 +130,11 @@ pub enum ServerMessage {
     },
 
     Notification(Notification),
+
+    #[cfg(feature = "urgency")]
+    Urgency {
+        status: u32,
+    },
 
     Ping,
 }
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
@@ -9,6 +9,9 @@ use autoconnect_common::{
 };
 use autopush_common::{endpoint::make_endpoint, util::sec_since_epoch};
 
+#[cfg(feature = "urgency")]
+use autopush_common::{db::Urgency, util::ms_since_epoch};
+
 use super::WebPushClient;
 use crate::error::{SMError, SMErrorKind};
 
@@ -38,6 +41,8 @@ impl WebPushClient {
                 self.nack(code);
                 Ok(vec![])
             }
+            #[cfg(feature = "urgency")]
+            ClientMessage::Urgency { min } => Ok(self.change_min_urgency(min).await?),
             ClientMessage::Ping => Ok(vec![self.ping()?]),
         }
     }
@@ -328,6 +333,42 @@ impl WebPushClient {
             Err(SMErrorKind::UaidReset.into())
         } else {
             Ok(vec![])
+        }
+    }
+
+    /// Update minimum urgency for the user and the flag
+    ///
+    /// If the new urgency is lower than the previous one,
+    /// We check pending messages, to send messages that were
+    /// retained because of their urgency
+    #[cfg(feature = "urgency")]
+    async fn change_min_urgency(
+        &mut self,
+        new_min: Urgency,
+    ) -> Result<Vec<ServerMessage>, SMError> {
+        // Change the min urgency
+        self.flags.min_urgency = new_min;
+
+        if let Some(mut user) = self.app_state.db.get_user(&self.uaid).await? {
+            // If the user hasn't set a minimum urgency yet, they receive all messages,
+            // which is equivalent to setting very-low as a minimum
+            let current_urgency = user.urgency.unwrap_or(Urgency::VeryLow);
+
+            // We update the user
+            user.urgency = Some(new_min);
+            user.connected_at = ms_since_epoch();
+            self.app_state.db.update_user(&mut user).await?;
+
+            let mut res = vec![ServerMessage::Urgency { status: 200 }];
+            // if new urgency < previous: fetch pending messages
+            if new_min < current_urgency {
+                self.ack_state.unacked_stored_highest = None;
+                self.current_timestamp = None;
+                res.append(&mut self.check_storage().await?);
+            }
+            Ok(res)
+        } else {
+            Ok(vec![ServerMessage::Urgency { status: 404 }])
         }
     }
 }
--- a/autopush-common/src/db/mod.rs
+++ b/autopush-common/src/db/mod.rs
@@ -12,6 +12,8 @@ use std::collections::{HashMap, HashSet};
 use std::result::Result as StdResult;
 
 use derive_builder::Builder;
+use lazy_static::lazy_static;
+use regex::RegexSet;
 use serde::Serializer;
 use serde_derive::{Deserialize, Serialize};
 use uuid::Uuid;
@@ -144,6 +146,9 @@ pub struct User {
     /// Last node/port the client was or may be connected to
     #[serde(skip_serializing_if = "Option::is_none")]
     pub node_id: Option<String>,
+    /// Last minimum urgency set by the client
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub urgency: Option<Urgency>,
     /// Record version
     #[serde(skip_serializing_if = "Option::is_none")]
     pub record_version: Option<u64>,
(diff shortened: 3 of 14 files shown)


autoconnect/autoconnect-common/src/protocol.rs
```diff
@@ -66,6 +66,10 @@ pub enum ClientMessage {
         version: String,
     },
```
jrconlin, line 68: Introducing new protocol messages should be kept behind a feature flag until we can properly verify that there are no problems with current or legacy UserAgents.

autoconnect/autoconnect-common/src/protocol.rs
```diff
@@ -123,6 +127,10 @@ pub enum ServerMessage {
 
     Notification(Notification),
```
jrconlin, line 129: Introducing new protocol messages should be kept behind a feature flag until we can properly verify that there are no problems with current or legacy UserAgents.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -330,4 +335,42 @@ impl WebPushClient {
...
+    /// retained because of their urgency
+    async fn change_min_urgency(
+        &mut self,
+        new_min: Urgency,
+    ) -> Result<Vec<ServerMessage>, SMError> {
+        // Change the min urgency
+        self.flags.min_urgency = new_min;
+
+        let status = if let Some(mut user) = self.app_state.db.get_user(&self.uaid).await? {
+            let current_urgency = user.urgency.unwrap_or(Urgency::VeryLow);
```
jrconlin, line 352: Wouldn't the default urgency be `Urgency::Normal`?

I presume that any message without an explicit urgency flag would be prioritized over any message that had specified a lower than normal urgency.

autopush-common/src/db/mod.rs
```diff
@@ -203,6 +207,27 @@ impl User {
...
+impl From<&str> for Urgency {
+    fn from(value: &str) -> Self {
+        match value.to_lowercase().as_str() {
+            "high" => Urgency::High,
+            "low" => Urgency::Low,
+            "very-low" => Urgency::VeryLow,
+            _ => Urgency::Normal,
+        }
+    }
+}
```
jrconlin, line 229: Considering that the two times you're using a `from` here, you might want to consider adding a `impl From<Option<actix_web::http::HeaderValue>> for Urgency`

jrconlin (changes requested):
Looks interesting, but my major concern is that we're changing the client protocol, and we'd want to make sure that we don't break anything on the current or legacy UA. I'd be more comfortable landing this if it was behind a feature flag so that we could test it in isolation from production or stage. 

Obviously, this change would also require updates to the UA code to actually consume and utilize the new `Urgency` message, and I don't know what priority that might be with that team.

p1gp1g:
Thank you for your review, I've introduced a new feature for the urgency and addressed 2 other comments

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
p1gp1g: I've added a comment about the fallback. It uses VeryLow as a default because if the user doesn't set a minimum, it means they want all messages

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -330,4 +335,42 @@ impl WebPushClient {
...
+            // We update the user
+            user.urgency = Some(new_min);
+            user.connected_at = ms_since_epoch();
+
+            // if new urgency < previous: fetch pending messages
+            if self.app_state.db.update_user(&mut user).await.is_ok() {
```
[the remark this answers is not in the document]
p1gp1g: I have done it to return a 404 if the user isn't saved yet (if they send urgency before registering to a channel)

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -330,4 +335,42 @@ impl WebPushClient {
...
+                if new_min < current_urgency {
+                    self.ack_state.unacked_stored_highest = None;
+                    self.current_timestamp = None;
+                    let mut res = vec![ServerMessage::Urgency { status: 200 }];
+                    res.append(&mut self.check_storage().await?);
+                    // We return the Urgency Ack + pending messages
+                    return Ok(res);
+                }
+                200
```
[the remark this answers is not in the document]
p1gp1g: Should I change something ?

autopush-common/src/db/mod.rs
p1gp1g: I've added `From<Option<&String>>`

jrconlin:
Thanks! Looks good. I'm going to do a quick bit of research on the client code to see if there are any potential impacts, but I like the approach you took. 

Sigh, Also just noted that this commit is not signed, (see our [Contributing doc](https://github.com/mozilla-services/autopush-rs/blob/master/CONTRIBUTING.md#sending-pull-requests). I also just realized that we never linked to [what "sign" means](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-verification), which might explain a few things ( _sighs_ ). Unfortunately, we cannot land unsigned commits, so you will need to close this PR in favor of one that is signed, or once we get approval, we can resubmit this for you. My apologies for not catching that earlier.

p1gp1g:
I have signed the commits. Let me know how your tests go with the client, there shouldn't be any issue since the ServerMessage::Urgency is only send after a ClientMessage::Urgency

FYI, I've written a [UnifiedPush](https://unifiedpush.org/) (~webpush for mobile applications) distributor that uses autopush, and I did the tests with it. By the way, I'll try to add UnifiedPush support to Firefox-Android soon

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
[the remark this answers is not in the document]
jrconlin: The problem is that if `update_user(..)` fails, we ought to know why (e.g. there may be a database issue happening that we should track. If the update fails, we should record the error and pass back a result that's meaningful to the caller. That's mostly handled by the existing error processing, but `is_ok()` interrupts that.

Granted, having this function return a context free status `500` doesn't really help much either. (How did we get this? How do we fix the problem? etc.)

If we get a 500, I'd say that all bets are off and that we return a proper 500 that closes the connection. 

That leaves either a list of pending `ServerMessage`s, or a single `ServerMessage` with a 404 status. 

I'd probably do something like (the following is me just writing code. Please correct any glaring bugs):

```
if let Some(mut user) = self.app_state.db.get_user(&self.uaid).await? {
    // If the user hasn't set a minimum urgency yet, they receive all messages,
    // which is equivalent to setting very-low as a minimum
    let current_urgency = user.urgency.unwrap_or(Urgency::VeryLow);

    // We update the user
    user.urgency = Some(new_min);
    user.connected_at = ms_since_epoch();

    // if new urgency < previous: fetch pending messages
    self.app_state.db.update_user(&mut user).await?;

    let mut res = vec![ServerMessage::Urgency { status: actix_web::http::StatusCode::OK }];
    // if new urgency < previous: fetch pending messages
    if new_min < current_urgency {
          self.ack_state.unacked_stored_highest = None;
          self.current_timestamp = None;
          res.append(&mut self.check_storage().await?);
      }
      // We return the Urgency Ack + pending messages
      return Ok(res); 
} else {
      return Ok(vec![ServerMessage::Urgency { status: actix_web::http::StatusCode::NOT_FOUND }])
}
```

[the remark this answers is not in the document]
p1gp1g: OK I see, I've pushed a patch. Regarding the status code, I prefer not mixing types, all other `ServerMessage`'s status are `u32`, and I think they should migrated to `StatusCode` in the same time.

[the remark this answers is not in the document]
jrconlin: You're right. That was more me thinking out loud that we should be consistent about using `StatusCode` when possible, but that should _not_ be part of this PR.

jrconlin (approved):
I'm going to give one of the other dev's a chance to look this over before I land it.

Thank you for this!

p1gp1g:
@jrconlin For information, how much time do you give them to review it ?

jrconlin:
Well, there's an outstanding conflict in the PR right now, so it's not going to land until that's resolved. Since this is isolated anyway, I'll land it after that's done and I give it a quick re-review.

p1gp1g:
I have rebased the branch

jrconlin (approved):

jrconlin:
Sigh. Just noticed that these commits aren't signed. [We require all commits to be signed](https://github.com/mozilla-services/autopush-rs/blob/master/CONTRIBUTING.md#sending-pull-requests) before we can accept them. (I've tried to make that very clear in the docs and PR templates, but I guess there's still work to do)

Sadly, that means we can't accept this PR as is (even one unsigned commit corrupts things.)

You can either close this PR and resubmit it or I can submit it on your behalf, but I'll need one of the other core folk to sign off on it.

p1gp1g:
I have squashed the branch and signed the commit

jrconlin (approved):

jrconlin:
Closing this in favor of https://github.com/mozilla-services/autopush-rs/pull/841 because of CI and signature issues.

PR #817, opened 2025-01-07
fix SYNC-4567: Pin poetry and update makefile to use new commands.

b4handjr:
This pins poetry on CI to use version 2.0.0 and updates our make file to make use of the `--project` flag. I also added a make command to install the pinned version.


b4handjr:
No specific reason besides that is what the current version is.

jrconlin (approved):

PR #818, opened 2025-01-07
doc: Update CONTRIBUTING to emphasize commit signing

jrconlin:
While previously mentioned in the CONTRIBUTING.md document, we should emphasize and assist folk in understanding the importance of signing commits for PRs. We cannot accept unsigned code, which can lead to frustration with folk very kindly providing support for us.


b4handjr:
Your poetry fix is due to a version bump. I have a fix for it here: https://github.com/mozilla-services/autopush-rs/pull/817

taddes (approved):
Thanks @jrconlin 👍

PR #830, opened 2025-01-22
Accept VAPID without sub

### LINKED ISSUE #814: Requirement for VAPID sub ?
I have a few app sending webpush messages with VAPID. They are designed for people self-hosting their own (application) server and to do some tests, I followed the spec and since they don't require the sub field, I haven't implemented it to reduce set up requirement for users.

I got some 401 with autopush when I realized you require `sub`. What's the reason to require it when VAPID isn't required at all ? Are you OK to make it optional ? I will set dummy mailto for the moment

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-4564)

p1gp1g:
Improve compatibility with RFC8292:
If the application server wishes to provide contact details, it MAY include a "sub" (Subject) claim in the JWT. So we don't require it.

Fix #814

---

I've also moved the check for empty sub before the bad format, because it was impossible to get SubEmpty error (it can't be empty and have one of the defined schemes)

### DIFF
--- a/autoendpoint/src/headers/vapid.rs
+++ b/autoendpoint/src/headers/vapid.rs
@@ -9,19 +9,13 @@ use crate::headers::util::split_key_value;
 use autopush_common::util::{sec_since_epoch, ONE_DAY_IN_SECONDS};
 
 pub const ALLOWED_SCHEMES: [&str; 3] = ["bearer", "webpush", "vapid"];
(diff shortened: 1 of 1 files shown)


jrconlin:
Closed in favor of https://github.com/mozilla-services/autopush-rs/pull/831

PR #831, opened 2025-01-22
bug: Make VAPID sub assertion optional

jrconlin:
Only reports `sub` if it is present and non-empty.

Closes: [SYNC-4564](https://mozilla-hub.atlassian.net/browse/SYNC-4564)

[SYNC-4564]: https://mozilla-hub.atlassian.net/browse/SYNC-4564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/extractors/subscription.rs
+++ b/autoendpoint/src/extractors/subscription.rs
@@ -94,10 +94,10 @@ impl FromRequest for Subscription {
             // let's just capture if we have a valid VAPID, as well as what sort of bad sub
             // values we get.
             if let Some(header) = &vapid {
-                let sub = header
+                if let Some(sub) = header
                     .vapid
                     .insecure_sub()
-                    .map_err(|e: VapidError| {
+                    .inspect_err(|e: &VapidError| {
                         // Capture the type of error and add it to metrics.
                         let mut tags = Tags::default();
                         tags.tags
@@ -108,10 +108,12 @@ impl FromRequest for Subscription {
                             .with_tag("error", e.as_metric())
                             .send();
                     })
-                    .unwrap_or_default();
+                    .unwrap_or_default()
+                {
+                    info!("VAPID sub: {sub}");
+                };
                 // For now, record that we had a good (?) VAPID sub,
                 app_state.metrics.incr("notification.auth.ok")?;
-                info!("VAPID sub: {:?}", sub)
             };
 
             match token_info.api_version {
@@ -304,7 +306,7 @@ fn validate_vapid_jwt(
     // Set the audiences we allow. This obsoletes the need to manually match
     // against values later.
     validation.set_audience(&[settings.endpoint_url().origin().ascii_serialization()]);
-    validation.set_required_spec_claims(&["exp", "aud", "sub"]);
+    validation.set_required_spec_claims(&["exp", "aud"]);
 
     let token_data = match jsonwebtoken::decode::<VapidClaims>(
         &vapid.token,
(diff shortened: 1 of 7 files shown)


autoendpoint/src/extractors/subscription.rs
```diff
@@ -111,7 +111,9 @@ impl FromRequest for Subscription {
                     .unwrap_or_default();
                 // For now, record that we had a good (?) VAPID sub,
                 app_state.metrics.incr("notification.auth.ok")?;
-                info!("VAPID sub: {:?}", sub)
+                if sub.is_some() {
+                    info!("VAPID sub: {:?}", sub);
+                }
```
pjenvey, line 116: Nit probably instead?
```suggestion
                if let Some(sub) = sub {
                    info!("VAPID sub: {sub}");
                }
```

autoendpoint/src/headers/vapid.rs
```diff
@@ -149,26 +149,24 @@ impl VapidHeader {
...
-            if !sub.starts_with("mailto:") && !sub.starts_with("https://") {
-                info!("🔐 Vapid: Bad Format {:?}", sub);
-                return Err(VapidError::SubBadFormat);
+        let sub = match data.sub {
+            Some(sub) => {
+                if !sub.starts_with("mailto:") && !sub.starts_with("https://") {
+                    info!("🔐 Vapid: Bad Format {:?}", sub);
+                    return Err(VapidError::SubBadFormat);
+                }
+                Some(sub)
```
pjenvey, line 164: let's do a `let else` --

I'll note [RFC8292](https://www.rfc-editor.org/rfc/rfc8292) says sub **SHOULD** be a mailto/https, so maybe we shouldn't even consider other schemes as errors (but it's only for our logging/debugging purposes anyway). However I'd argue the empty check should be kept if we're continuing to consider other schemes as errors.

```suggestion
        let Some(sub) = data.sub else {
            return Ok(None);
        };
        if !sub.starts_with("mailto:") && !sub.starts_with("https://") {
            info!("🔐 Vapid: Bad Format {sub:?}");
            return Err(VapidError::SubBadFormat);
        }
        if sub.is_empty() {
            info!("🔐 Empty Vapid sub");
            return Err(VapidError::SubEmpty);
        }
        info!("🔐 Vapid: sub: {sub}");
        Ok(Some(sub))
```

pjenvey (changes requested):
@jrconlin The wrong issue's in the commit message, this is https://mozilla-hub.atlassian.net/browse/SYNC-4564

autoendpoint/src/extractors/subscription.rs
```diff
@@ -94,24 +94,21 @@ impl FromRequest for Subscription {
...
+                    // Capture the type of error and add it to metrics.
+                    let mut tags = Tags::default();
+                    tags.tags
+                        .insert("error".to_owned(), e.as_metric().to_owned());
+                    app_state
+                        .metrics
+                        .incr_with_tags("notification.auth.error")
+                        .with_tag("error", e.as_metric())
+                        .send();
+                })? {
```
pjenvey, line 107: This was eating the error previously w/ `unwrap_or_default`

jrconlin: Ah, right. I was thinking we wanted it to be raised, but not here. We actually want to eat the error after we add the metric. (sigh)

pjenvey (approved):

autoendpoint/src/headers/vapid.rs
```diff
@@ -149,26 +149,23 @@ impl VapidHeader {
...
-            }
-            info!("🔐 Vapid: sub: {:?}", sub);
-            return Ok(sub.to_owned());
+        let Some(sub) = data.sub else { return Ok(None) };
+        if !sub.starts_with("mailto:") && !sub.starts_with("https://") {
+            info!("🔐 Vapid: Bad Format {sub:?}");
+            return Err(VapidError::SubBadFormat);
+        };
+        if sub.is_empty() {
+            info!("🔐 Empty Vapid sub");
```
p1gp1g, line 164: @jrconlin : This is unreachable code, sub can't start with "mailto:" or "https://" and be empty. The check for empty sub must be moved before the other one.

You can use tests from #830 to correctly check that

jrconlin: Ah! Good catch. I'll submit a follow up.

PR #837, opened 2025-02-07
feat: normalize durations and timestamps

taddes:
I really went back and forth a lot for the definitions of `SystemTime` in the BigTable implementation and ended up setting everything back to using `SystemTime`. There isn't a good monotonic stand in from `chrono` and it seemed silly to be converting all the `chrono` duration back to things that would play nice with Diesel and the spanner impls.  Happy to revisit that, but it just added bloat that seemed to offer little value.  Lmk what you think about that @pjenvey & @jrconlin . Fundamentally, this addresses the need to have solid datetime methods defining times and duration opposed to integers. Conversions to integers happen at invocations that need that, given their signatures, which I think was the key problem we needed to fix from the incident.

Closes [SYNC-4521](https://mozilla-hub.atlassian.net/browse/SYNC-4521) [SYNC-4465](https://mozilla-hub.atlassian.net/browse/SYNC-4465)

[SYNC-4521]: https://mozilla-hub.atlassian.net/browse/SYNC-4521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SYNC-4465]: https://mozilla-hub.atlassian.net/browse/SYNC-4465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Cargo.toml
```diff
@@ -64,7 +64,7 @@ log = { version = "0.4", features = [
 ] }
 mockall = "0.12"
 mozsvc-common = "0.2"
-openssl = { version = "0.10" }
+openssl = { version = "0.10.70" }
```
pjenvey, line 67: We tend to not specify minor revisions here. It's a good balance of not having to touch `Cargo.toml` for every minor bump + allows cargo to figure out the latest version for us when needed -- which shouldn't matter to our code as minor revisions should be backwards compat/not change APIs.

To handle CVE/RUSTSECs in this way you can _usually_ just do an e.g. `cargo update -p openssl` to bump to the latest `0.10` only in `Cargo.lock`

jrconlin: IIRC, [Rust is a bit odd about versions](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax), so 
`0.10.70` translates to "use anything between 0.10.70 and 0.11.0" (which is the same as `~0.10.70` and because there's a leading `0` it's special cased.)

If we need to lock to a specific version, we should use `^0.10.70`

autopush-common/src/lib.rs
```diff
@@ -46,10 +48,10 @@ pub mod util;
 /// "abandoned" and any router info assigned to a User Agent that has not contacted
 /// Autopush in 60 days can be discarded.
 ///
-const ONE_DAY_IN_SECONDS: u64 = 24 * 60 * 60;
+pub const ONE_DAY_IN_SECONDS: u64 = Duration::days(1).num_seconds() as u64;
```
jrconlin, line 51: Huh. I thought the idea was that we'd make these `Duration` values, and only convert to seconds when we need to.

e.g. 
`.map(|ttl| min(ttl, MAX_NOTIFICATION_TTL.num_seconds() as i64)`

Basically, strongly type TTL as a `Duration` so that we don't muck up how we work with them accidentally.

pjenvey: I believe Taddes is working on that (so this is currently a WIP)

jrconlin: Ah, sorry! Hope you don't mind, but I converted this to a draft for now so I don't do more dumb drive-by's. 

My apologies.

Cargo.toml
taddes: Cool, thx for that! Will do.

autopush-common/src/lib.rs
taddes: Yeah, sry, should have moved back to WIP. Wrestling with Prometheus all day and didn't put it back.

jrconlin (dismissed):
So much cleaner. Thank You!

autopush-common/src/reliability.rs
```diff
@@ -118,7 +125,11 @@ impl PushReliability {
                     .unwrap_or_else(|| "None".to_owned()),
                 new
             );
-            match client.get_connection_with_timeout(CONNECTION_EXPIRATION) {
+            match client.get_connection_with_timeout(
+                CONNECTION_EXPIRATION
+                    .to_std()
+                    .expect("Could not convert const CONNECTION_EXPIRATION."),
```
pjenvey, line 131: I think the usage of `map_err` vs `expect` should be consistent in the 3 places in this module. However, more importantly, w/ `to_std` returning a `Result`, I'd argue this is one of those cases where the usage is specifically limited to `get_connection_with_timeout`'s argument that we should just keep the original `std::time::Duration` to avoid `to_std`'s noise (sigh).

pjenvey (changes requested):
> I really went back and forth a lot for the definitions of SystemTime in the BigTable implementation and ended up setting everything back to using SystemTime. 

Agreed there's a fine line here. autopush might have more cases that look like they're asking to be converted fully to `chrono`, like e.g. in the Bigtable impl where we're using both std time and chrono (and u64 casting on top of all that)

```rust
let expiry =
            std::time::SystemTime::now() + Duration::from_secs(MAX_ROUTER_TTL.num_seconds() as u64);
```

But like you say in other cases such as when you ultimately need a `std::time::Duration`, having to handle a `to_std()` -> `Result` doesn't always seem worth it. 

This definitely solves the original issue from the incident though 👍

pjenvey (approved):

PR #840, opened 2025-02-13
feat: Add reliability tracking of Error state.

jrconlin:
This is a debatable work in progress that records when an message goes into an ERROR state. This does **NOT** do any fine grain about _what_ error. If you want to know that, it's more on you. This is just noting that "ok, tick that this message failed for some reason" for the eventual collection report phase.

Pro: punting errored messages to an explicit state provides cleaner snapshots of final message states.
Con: this obscures where the message hit an error, meaning that it's easier to ignore potential bottlenecks.


jrconlin:
closed in favor of https://github.com/mozilla-services/autopush-rs/pull/845

PR #845, opened 2025-02-18
feat: Add "long term" report summary generator for push reliability

jrconlin:
This adds a daily (or longer) summary generator for the push report.

Issue: SYNC-4610


pjenvey (changes requested):

scripts/reliability_report.py
```diff
@@ -0,0 +1,272 @@
...
+                args = parser.set_defaults(**toml.load(f))
+    args = parser.parse_args()
+    # fixup the bigtable settings so that they're easier for this script to deal with.
+    if args.db_settings is not None:
+        bt_settings = json.loads(args.db_settings)
+        parts = bt_settings.get("table_name").split("/")
+        for i in range(0, len(parts), 2):
+            # remember: the `tablename` dsn uses plurals for
+            # `projects`, `instances`, & `tables`
+            bt_settings[parts[i].rstrip("s")] = parts[i + 1]
```
pjenvey, line 245: You're saying "projects/etc/instances/etcetc" isn't compatible with this client? Isn't that the standard? e.g. `_INSTANCE_NAME_RE` [here](https://gcloud.readthedocs.io/en/latest/_modules/google/cloud/bigtable/instance.html)

jrconlin: It is. 
What we're doing here is converting the plural names to singular keys e.g. `projects/foo/instances/bar/...` become `"project"=>"foo", "instance"=>"bar"` we use this later in `BigtableScanner.collect()` 
Granted, we could have just as easily used the plural and instead gone with 
```python
table = self.bigtable.get_table(
    self.settings.bigtable.get("instances"),
    self.settings.bigtable.get("tables"),
)
```
but that parses a bit strange, since these are singular values.

pjenvey (changes requested):

scripts/reliability_report.py
```diff
@@ -0,0 +1,267 @@
...
+        success_count = fail_count = row_count = expired_count = 0
+
+        # Iterate over the rows and collect interesting bits of info
+        for row_count, row in enumerate(raw_rows, start=1):
+            milestones = OrderedDict()
+            successful = expired = False
+            total_time = 0
+            for cell in row.cells:
+                milestone = bytearray(cell.qualifier).decode()
+                timestamp = int.from_bytes(cell.value) * 0.000000001
```
pjenvey, line 91: Also I'm confused by the timestamp value: we're storing our `ms_since_epoch` in the cell value,  -- wouldn't it be `* 1000` for comparison against python's `time.time`? 

Since we're using the bigtable recognized/standard 64 bit be type, optionally you could use `int(cell)` which they have for the conversion from bytes.

```suggestion
                timestamp = int.from_bytes(cell.value) * 0.000000001
```

jrconlin: This is pulling the time value stored in bigtable, which is a millisecond value stored as a u64 int vale.  We need to convert the integer to an appropriately scaled float to represent the milliseconds as the remainder. Currently, the UTC is expressed as a 10 digit number (e.g. `1741896409`). Multiplying this by the 10 digit `0.000000001` offsets the decimal point such that the remainder of the int64 value can provide the millisecond value compatible with python. Multiplying the u64 value by 1000 would only increase the integer base and would not resolve the remainder.

I'm not sure what the suggested change entails, since it appears that it is the same code. Is there some white-space difference that is not apparent?

pjenvey: oops I meant .001 (not `*1000`) as python's `time.time` is seconds (w/ more precision in the fractional part) - this is now resolved

pjenvey (approved):

PR #847, opened 2025-02-20
chore: Update to rust 1.85

jrconlin:
Minor code update for clippy, and a `cargo update` sweep for dependencies.


pjenvey (changes requested):
This should also bump the versions in Dockerfile/CI (which JR was waiting intentionally waiting for the version's new docker tags to land)

pjenvey (approved):

taddes (approved):
👍

PR #853, opened 2025-03-05
feat: record a couple more state changes, particularly Delivered

pjenvey:
and generalize report_handler's error reporting

Closes SYNC-4644

### DIFF
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
@@ -214,16 +214,15 @@ impl WebPushClient {
                     "version" => &notif.version
                 );
                 // Get the stored notification record.
-                let n = &self.ack_state.unacked_stored_notifs[pos];
+                let n = &mut self.ack_state.unacked_stored_notifs[pos];
+                let is_topic = n.topic.is_some();
                 debug!("✅ Ack notif: {:?}", &n);
-                // TODO: Record "ack'd" reliability_id, if present.
                 // Only force delete Topic messages, since they don't have a timestamp.
                 // Other messages persist in the database, to be, eventually, cleaned up by their
                 // TTL. We will need to update the `CurrentTimestamp` field for the channel
                 // record. Use that field to set the baseline timestamp for when to pull messages
                 // in the future.
-                // Topic/legacy messages have no sortkey_timestamp
-                if n.sortkey_timestamp.is_none() {
+                if is_topic {
                     debug!(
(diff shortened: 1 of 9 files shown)


autopush-common/src/reliability.rs
```diff
@@ -285,30 +285,12 @@ pub fn gen_report(values: HashMap<String, i32>) -> Result<String> {
...
-                .content_type("text/plain")
-                .body(format!("# ERROR: {e}\n"))
-        }
-    }
+pub async fn report_handler(reliability: &Arc<PushReliability>) -> Result<HttpResponse> {
+    reliability.gc().await?;
+    let report = gen_report(reliability.report().await?)?;
+    Ok(HttpResponse::Ok()
+        .content_type("application/openmetrics-text; version=1.0.0; charset=utf-8")
+        .body(report))
```
jrconlin, line 293: Have we checked to see what Prometheus does if it encounters non-text content for reports?

jrconlin: Just checked. Prometheus rejects the content of the page as an error and records no data for that sample.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -232,8 +231,21 @@ impl WebPushClient {
                         .db
                         .remove_message(&self.uaid, &n.chidmessageid())
                         .await?;
+                    // NOTE: timestamp messages may still be in state of flux: they're not fully
```
jrconlin, line 234: We should really only mark this as `Delivered` if the `ACK` code is [100](https://searchfox.org/mozilla-central/source/dom/push/PushServiceWebSocket.sys.mjs#27). The others aren't really 'ACK`s and probably should be considered Errors.

jrconlin (changes requested):

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
pjenvey: How about separate states for these instead of Errors (especially since we have always treated them as normal ACKs)? 100: Delivered, 101: DecryptionError, 102: NotDelivered

jrconlin: I think I'm ok with that. I guess I was more thinking that we'd only really consider a message as "Delivered" if the UA returned a code of 100, since that means that the webapp actually received the message. We could consider receipt by the User Agent to be "Delivered" but we should explain why the webapp might not have processed it correctly (a bit out of scope for us, but whatever)

jrconlin (approved):

PR #857, opened 2025-03-11
feat: add reliability states for different ack type codes

pjenvey:
- cleanup state serialized names to match their variant names
- ignore RUSTSEC-2024-0437 for now

Issue SYNC-4644

### DIFF
--- a/autopush-common/src/reliability.rs
+++ b/autopush-common/src/reliability.rs
@@ -14,7 +14,7 @@ use prometheus_client::{
 use redis::{Commands, ConnectionLike};
 
 use crate::db::client::DbClient;
-use crate::errors::{ApcError, ApcErrorKind, Result};
+use crate::errors::{ApcErrorKind, Result};
 use crate::util::timing::sec_since_epoch;
 
 // Redis Keys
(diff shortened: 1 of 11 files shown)


autopush-common/src/reliability.rs
```diff
@@ -26,75 +26,32 @@ const CONNECTION_EXPIRATION: Duration = Duration::from_secs(10);
...
+    strum::EnumString,
+)]
+#[serde(rename_all = "snake_case")]
+#[strum(serialize_all = "snake_case")]
+pub enum ReliabilityState {
+    Received,        // Message was received by the Push Server
+    Stored,          // Message was stored because it could not be delivered immediately
+    Retrieved,       // Message was taken from storage for delivery
+    IntTransmitted,  // Message was handed off between autoendpoint and autoconnect
+    IntAccepted,     // Message was accepted by autoconnect from autopendpoint
```
pjenvey, line 48: @jrconlin I assume the previous, mismatched serialized names for these were a typo/aren't necessary, but if they're still needed we can restore them easily enough with one off procmacros on these fields

jrconlin (approved):

autopush-common/src/reliability.rs
jrconlin: wfm. So long as their differentiated, I don't really care.

PR #858, opened 2025-03-12
fix: typo in cron script name

pjenvey:
Issue SYNC-4327


jrconlin (approved):

jrconlin:
We may drop this file eventually. 
The "reaper" function is incorporated into the Prometheus  reporting, so this file is somewhat ancillary.

jrconlin (approved):
this is why I added a spellchecker to my instance as of late.

PR #859, opened 2025-03-14
bug: Autoendpoint should be less strict about config options

jrconlin:
Closes: [SYNC-4672](https://mozilla-hub.atlassian.net/browse/SYNC-4672)

### DIFF
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -16,7 +16,6 @@ pub const ENV_PREFIX: &str = "autoend";
 
 #[derive(Clone, Debug, Deserialize)]
 #[serde(default)]
-#[serde(deny_unknown_fields)]
 pub struct Settings {
     pub scheme: String,
     pub host: String,
@@ -135,6 +134,7 @@ impl Settings {
             }
         })?;
 
+        dbg!(&built);
         Ok(built)
     }
 
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -134,7 +134,6 @@ impl Settings {
             }
         })?;
 
-        dbg!(&built);
         Ok(built)
     }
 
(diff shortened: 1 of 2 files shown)


autoendpoint/src/settings.rs
```diff
@@ -16,7 +16,6 @@ pub const ENV_PREFIX: &str = "autoend";
 
 #[derive(Clone, Debug, Deserialize)]
 #[serde(default)]
-#[serde(deny_unknown_fields)]
```
pjenvey, line 19: How about allowing the reliability setting even when the feature's disabled instead?

jrconlin: This is just the latest time we've had this problem occur. Ideally, we should raise a warning if we encounter an unknown config variable, but instead `config` raises the single error from serde which causes the app to bail. We really shouldn't do that, particularly for a config change like this.

Instead, we should just ignore the unknown config change like we do with `autoconnect`. This change makes `autoendpoint` work the same way.

pjenvey: I don't recall the last time this specifically occurred (I probably forgot all about it) but in theory I prefer the failing fast behavior so we immediately know the config is broken vs it quietly occurring without notice. I think warnings are too easy to ignore but they're better than nothing for e.g. deprecating old config knobs.

Though as you say we don't `deny_unknown_fields` in autoconnect or apparently anywhere else, such as syncstorage, so I see your point

pjenvey (approved):

autoendpoint/src/settings.rs
jrconlin: Yeah, it's definitely tricky, and I'd prefer if `config` had more options around this. (e.g. it didn't eat the error, but raised it so it became catchable, or if there was an option to raise the warning.) Granted, this may become more of a moot point as we start taking ownership over more of the deployment systems, but sometimes getting things timed correctly can be challenging.

I may submit a PR to `config` to do some of this, since I can definitely see the benefits to other projects that hit this sort of thing.

PR #863, opened 2025-03-19
bug: compare tracking keys in raw form.

jrconlin:
It is possible for the tracking key to be specified in Standard format vs. URL safe. In addition, there are other formatting concerns when comparing as encoded strings. To be more accurate, we should compare the raw public key string.

NOTE: I'm going to fix the config file for the servers. This patch is lower priority and is only to catch future misconfiguration problems.

### DIFF
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -173,10 +173,10 @@ impl Settings {
             .collect()
     }
 
-    /// Get the list of tracking public keys
-    // TODO: this should return a Vec<[u8]> so that key formatting errors do not cause
-    // false rejections. This is not a problem now since we have access to the source
-    // public key, but that may not always be true.
+    /// Get the list of tracking public keys converted to raw, x962 format byte arrays.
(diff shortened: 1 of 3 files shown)


autoendpoint/src/settings.rs
```diff
@@ -175,13 +177,18 @@ impl Settings {
     // TODO: this should return a Vec<[u8]> so that key formatting errors do not cause
     // false rejections. This is not a problem now since we have access to the source
     // public key, but that may not always be true.
-    pub fn tracking_keys(&self) -> Vec<String> {
+    pub fn tracking_keys(&self) -> Result<Vec<Vec<u8>>, ConfigError> {
```
pjenvey, line 180: Just this method's doc comment needs updating per the new signature

pjenvey (dismissed):

autoendpoint/src/settings.rs
jrconlin: Ah, good catch. Gives me a chance to improve that doc string anyway.

pjenvey (approved):

Issue #895, opened 2025-04-08
Create terminal state cleanup system


data-sync-user:
Currently, the push reliability system records current milestone states. Messages progress through each milestone, until reaching one of the final “terminal” states:

* `Delivered` - Message was successfully delivered to the WebApp
* `DecryptionError` - Message was not delivered due to a decryption error in the UA
* `NotDelivered` - Message was unable to be delivered by the UA for an unknown reason
* `Expired` - Message Expired while in transit
* `Errored` - Message errored out during progression for unknown reasons.

These milestones have no “clean-up” and thus could continue to increment. This could be a problem for a number of reasons.

We should implement a way to limit the growth of these to reasonable numbers. I suggest that once a day, the current terminal milestone counts be recorded and stored. Once the maximum information retention period expires (`RELIABLE_LOG_TTL`) the “terminal” milestone numbers are adjusted to remove the counts of the elapsed states  (e.g. on May 30, the terminal counts recorded on April 01 would be removed from the current respective terminal states, ensuring that only states for currently tracked messages would be recorded.) 

While this would obscure some longer term history (e.g. count of messages successfully delivered per year), this would also prevent a substantial amount of noise potentially obscuring smaller values when viewing charts. (e.g. a spike of 50% of 1000 errors would be hard to spot if the max display value is several billion total messages.)

This would increase storage to include 60 new records listing the daily snap-shot values, and would require adding code to both record the snapshot and apply the corrections in an atomic manner.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/PUSH-527)

pjenvey:
test

PR #899, opened 2025-04-25
feat: Clean up terminal states

jrconlin:
Address potential problem where final message states may grow without bounds.
This also updates the `reliability_report.py` to write to a Google Storage Bucket. 

Closes: #PUSH-12 / #895


jrconlin:
pending gcp storage permission to test.

autopush-common/src/reliability.rs
```diff
@@ -60,6 +59,22 @@ pub enum ReliabilityState {
...
 }
 
+impl ReliabilityState {
+    ///  Has the message reached a state where no further transitions should be possible?
+    pub fn is_terminal(self) -> bool {
+        // NOTE: this list should match the daily snapshot captured by `reliability_cron.py`
+        // which will trim these counts after the max message TTL has expired.
+        matches!(
+            self,
+            ReliabilityState::DecryptionError
```
pjenvey, line 69: I had a prototype `BridgeTransmitted` to distinguish between a webpush notification vs bridges, which would have been a terminal state. Would it be worth adding such a distinction set as terminal?

jrconlin: Ah, yes. If a message never exits `BridgeTransmitted` on it's own, then we should add it. 
I may add a comment about how this could be transitory if/whenever we add mobile client reception reporting, but that's not currently available.

autopush-common/src/reliability.rs
```diff
@@ -46,18 +45,36 @@ const CONNECTION_EXPIRATION: TimeDelta = TimeDelta::seconds(10);
...
-    IntTransmitted,  // Message was handed off between autoendpoint and autoconnect
-    IntAccepted,     // Message was accepted by autoconnect from autopendpoint
-    Transmitted,     // Message was handed off for delivery to the UA
-    Accepted,        // Message was accepted for delivery by the UA
-    Delivered,       // Message was provided to the WebApp recipient by the UA
-    DecryptionError, // Message was provided to the UA and it reported a decryption error
-    NotDelivered,    // Message was provided to the UA and it reported a not delivered error
-    Expired,         // Message expired naturally (e.g. TTL=0)
-    Errored,         // Message resulted in an Error state and failed to be delivered.
+    Received,          // Message was received by the Push Server
```
gruberb, line 48: I would love these to be doc-comments (`///`) above each variant. And the variant itself. Once we generate docs, we can read through them :)

gruberb (approved):
Just a little nit about comments vs. doc comments on the enum.

autopush-common/src/reliability.rs
jrconlin: https://mozilla-hub.atlassian.net/browse/PUSH-534

Issue #897, opened 2025-04-11
Convert "magic strings" to enums


data-sync-user:
There are a number of places where Autopush uses “magic strings”. For instance, in `autoconnect-ws-sm::unidentified.rs` we identify various message types by strings like “hello” instead of referring to an Enum value. This can lead to discrepancies due to spelling or formatting. We should instead use declared `enum` with string serialization to provide consistent values. (Crates like [strum](https://crates.io/crates/strum)  can be useful in providing easy enum to string conversions.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/PUSH-141)

PR #903, opened 2025-05-08
feat: Remove magic strings and add enum variants instead

gruberb:
Closes: https://github.com/mozilla-services/autopush-rs/issues/897

- Added a `BroadcastErrorType` enum
- Added a `MessageType`  enum with a conversion from `ClientMessage`/`ServerMessage` to `MessageType`
- Added a `MetricName` enum
- To work easier with Metrics, I added an extension trait in `autopush-common/metrics.rs`, with methods `inc`, `inc_raw`, inc_with_tags`. The `inc` and `inc_with_tags` methods accept a `MetricName`
-  This replaces the `cadence::CountedExt` trait with our own `StatsdClientExt` trait with almost the same methods. 

Since it replaces a lot of text, and it's my first PR, I might have missed something. But the local tests run through.


autoconnect/autoconnect-common/src/test_support.rs
```diff
@@ -13,10 +14,20 @@ pub const DUMMY_CHID: Uuid = Uuid::from_u128(0xdeadbeef_0000_0000_abad_1dea00000
...
+    format!(
+        r#"{{"messageType": "{}", "use_webpush": true}}"#,
+        MessageType::Hello.as_str()
+    )
+}
+
+pub fn hello_again_json() -> String {
+    format!(
+        r#"{{"messageType": "{}", "use_webpush": true,
+                "uaid": "deadbeef-0000-0000-deca-fbad00000000"}}"#,
```
jrconlin, line 27: I'd really appreciate if we used `DUMMY_UAID` here instead of the string. That way we're consistent and if we ever need to change, we don't have a leak.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/unidentified.rs
```diff
@@ -280,12 +294,9 @@ mod tests {
             db: hello_db().into_boxed_arc(),
             ..Default::default()
         });
-        // Ensure that we do not need to pass the "use_webpush" flag.
-        // (yes, this could just be passing the string, but I want to be
-        // very explicit here.)
-        let json = serde_json::json!({"messageType":"hello"});
-        let raw = json.to_string();
-        let msg = ClientMessage::from_str(&raw).unwrap();
+        // Using hello_json helper ensures consistent message type strings
```
jrconlin, line 297: It also breaks the test.
We want to ensure that `use_webpush` is _*NOT*_ required by the `hello` message, even though it is commonly passed by clients. Thus why we were hand building things.

autoendpoint/src/metrics.rs
```diff
@@ -128,28 +128,25 @@ impl Metrics {
     }
 
     // increment a counter with no tags data.
-    pub fn incr(self, label: &str) {
-        self.incr_with_tags(label, None)
+    pub fn incr(self, metric: MetricName, tags: Option<Tags>) {
```
jrconlin, line 131: I'm confused.
`incr()` was pretty much a convenience method to allow the caller not to include the tags (we would just pass a `None` to `incr_with_tags`), yet now they have to include the `None`?
Doesn't including the tags make this less convenient?

jrconlin (changes requested):

autoconnect/autoconnect-common/src/test_support.rs
```diff
@@ -13,10 +14,20 @@ pub const DUMMY_CHID: Uuid = Uuid::from_u128(0xdeadbeef_0000_0000_abad_1dea00000
 
 /// A minimal websocket Push "hello" message, used by an unregistered UA with
 /// no existing channel subscriptions
-pub const HELLO: &str = r#"{"messageType": "hello", "use_webpush": true}"#;
-/// A post initial registration response
-pub const HELLO_AGAIN: &str = r#"{"messageType": "hello", "use_webpush": true,
-                                  "uaid": "deadbeef-0000-0000-deca-fbad00000000"}"#;
+pub fn hello_json() -> String {
```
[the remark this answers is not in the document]
gruberb: Good point 👍

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -21,7 +23,8 @@ impl WebPushClient {
     ) -> Result<Vec<ServerMessage>, SMError> {
         match msg {
             ClientMessage::Hello { .. } => {
-                Err(SMError::invalid_message("Already Hello'd".to_owned()))
+                // Hello messages should be handled by UnidentifiedClient, not here
+                Err(SMError::expected_message_type(MessageType::Hello))
```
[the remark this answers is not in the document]
gruberb: Good point, reverted the change! 👍

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -209,9 +220,10 @@ impl WebPushClient {
                 .position(|n| n.channel_id == notif.channel_id && n.version == notif.version);
             if let Some(pos) = pos {
                 debug!(
-                    "✅ Ack (Stored)";
-                    "channel_id" => notif.channel_id.as_hyphenated().to_string(),
-                    "version" => &notif.version
+                    "✅ Ack (Direct)";
```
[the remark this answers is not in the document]
gruberb: Ah, sorry, I think I changed these lines a few times and never double checked if I have the correct Ack when reverting part of it back!

autoendpoint/src/metrics.rs
gruberb: Thanks for catching this! 👍

Cargo.toml
```diff
@@ -79,9 +79,9 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 slog = { version = "2.7", features = [
-  "dynamic-keys",
-  "max_level_trace",
-  "release_max_level_info",
+    "dynamic-keys",
+    "max_level_trace",
+    "release_max_level_info",
```
gruberb, line 84: No matter the editor, I am getting these diffs....

jrconlin: Huh, it *might* be a `cargo fmt` change? There have been a few similar things in the past. 
FWIW, I tend to run the latest stable, but you'll notice that the version in `#RUST_VER` tends to lag behind a bit.

jrconlin (approved):
Thank you for this!

PR #981, opened 2025-06-04
bug: Use full paths in compiled docker image

jrconlin:
Deployment is reporting that it can't find the executable, so specifying the full paths to all elements.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -590,7 +590,7 @@ workflows:
             tags:
               only: /.*/
 
-      # Comment out the following two sections for local CircleCI testing.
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -590,7 +590,7 @@ workflows:
             tags:
               only: /.*/
 
-      # Comment out the following two sections for local CircleCI testing.
+      # Comment out the following three sections for local CircleCI testing.
```
jrconlin, line 593: Corrected to include the new reliability command

taddes (approved):
👍

PR #987, opened 2025-06-09
bug: Fix the report args.

jrconlin:
Due to a few last minute recommendations, the args fell out of sync.

Taking the opportunity to fix up some of the arguments to make better lexicographic sense.

### DIFF
--- a/scripts/reliability/reliability_report.py
+++ b/scripts/reliability/reliability_report.py
@@ -437,9 +437,9 @@ def config(env_args: os._Environ = os.environ) -> argparse.Namespace:
         "-s",
         help="Database settings",
         default=env_args.get(
-            "AUTOEND__DB_SETTINGS",
+            "AUTOCONNECT__DB_SETTINGS",
             env_args.get(
-                "AUTOCONNECT__DB_SETTINGS",
+                "AUTOEND__DB_SETTINGS",
                 '{"message_family":"message","router_family":"router", "table_name":"projects/test/instances/test/tables/autopush"}',
             ),
         ),


scripts/reliability/reliability_report.py
```diff
@@ -430,16 +432,6 @@ def config(env_args: os._Environ = os.environ) -> argparse.Namespace:
 
     parser.add_argument("-c", "--config", help="configuration_file", action="append")
 
-    parser.add_argument(
```
jrconlin, line 433: This arg was unused. We draw the arguments from the `settings`, and the actual db is determined by the running environment.

scripts/reliability/reliability_report.py
```diff
@@ -525,18 +511,10 @@ def config(env_args: os._Environ = os.environ) -> argparse.Namespace:
...
-        default=env_args.get("AUTOTRACK_REPORT_MAX_RETENTION_DAYS", 30),
-    )
-
     parser.add_argument(
         "--report_bucket_name",
         help="Name of the bucket to store reliability reports",
-        default=env_args.get(
-            "AUTOTRACK_REPORT_BUCKET_NAME", "autopush-dev-reliability"
-        ),
+        default=env_args.get("AUTOTRACK_REPORT_BUCKET_NAME", "autopush-reliability"),
```
jrconlin, line 517: Defaulting for the "production" bucket name.

scripts/reliability/reliability_report.py
```diff
@@ -485,8 +471,8 @@ def config(env_args: os._Environ = os.environ) -> argparse.Namespace:
         "-r",
         help="DSN to connect to the Redis like service.",
         default=env_args.get(
-            "AUTOEND__RELIABILITY_DSN",
-            env_args.get("AUTOCONNECT__RELIABILITY_DSN", "redis://localhost"),
+            "AUTOCONNECT__RELIABILITY_DSN",
+            env_args.get("AUTOEND__RELIABILITY_DSN", "redis://localhost"),
```
pjenvey, line 475: `--db_settings` still defaults to `AUTOEND__` before `AUTOCONNECT__`, it should be changed as well

pjenvey (changes requested):

scripts/reliability/reliability_report.py
jrconlin: Yeah, I was debating that. This one is actively used, so I prioritized, but you're right, I should do all of them for consistency .

PR #991, opened 2025-06-12
feat: Improve endpoint reliability check

jrconlin:
* Add some hinting to `health` to report vapid key signatures (to ensure that key values are propagating)
* Add db check to autoendpoint `health` check
* Try fix for off counts in `internal_record`
   * wrap action in a transaction (with retries), 
   * Add missing old state removal (not sure when/how that got dropped)
   * Add unit test for `internal_record`
* Add `LOCK_` prefix for redis lock record (because otherwise it's confusing)

### DIFF
--- a/autoendpoint/src/routes/health.rs
+++ b/autoendpoint/src/routes/health.rs
@@ -11,9 +11,10 @@ use serde_json::json;
 
 use crate::error::{ApiErrorKind, ApiResult};
 use crate::server::AppState;
-use autopush_common::db::error::DbResult;
 use autopush_common::metric_name::MetricName;
 use autopush_common::metrics::StatsdClientExt;
+use autopush_common::util::b64_encode_url;
+use autopush_common::{db::error::DbResult, errors::ApcError};
 
 /// Handle the `/health` and `/__heartbeat__` routes
 pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
@@ -39,23 +40,36 @@ pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
 
     #[cfg(feature = "reliable_report")]
     {
-        let mut reliability_health = state.reliability.health_check().await.unwrap_or_else(|e| {
-            state
-                .metrics
-                .incr_with_tags(MetricName::ReliabilityErrorRedisUnavailable)
-                .with_tag("application", "autoendpoint")
-                .send();
-            error!("🔍🟥 Reliability reporting down: {:?}", e);
-            "STORE_ERROR"
-        });
-        if state
-            .settings
-            .tracking_keys()
-            .unwrap_or_default()
-            .is_empty()
-        {
-            reliability_health = "NO_TRACKING_KEYS";
-        }
+        let reliability_health: Result<String, ApcError> = state
+            .reliability
+            .health_check()
+            .await
+            .map(|_| {
+                let keys: Vec<String> = state
+                    .settings
+                    .tracking_keys()
+                    .unwrap_or_default()
+                    .iter()
+                    .map(|k|
+                        // Hint the key values
+                        b64_encode_url(k)[..8].to_string())
+                    .collect();
+                if keys.is_empty() {
+                    Ok("NO_TRACKING_KEYS".to_owned())
+                } else {
+                    Ok(format!("OK: {}", keys.join(",")))
+                }
+            })
+            .unwrap_or_else(|e| {
+                // Record that Redis is down.
+                state
+                    .metrics
+                    .incr_with_tags(MetricName::ReliabilityErrorRedisUnavailable)
+                    .with_tag("application", "autoendpoint")
+                    .send();
+                error!("🔍🟥 Reliability reporting down: {:?}", e);
+                Ok("STORE_ERROR".to_owned())
+            });
         health.insert("reliability", json!(reliability_health));
     }
     Json(json!(health))
--- a/autoendpoint/src/routes/health.rs
+++ b/autoendpoint/src/routes/health.rs
@@ -24,19 +24,25 @@ pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
     routers.insert("apns", state.apns_router.active());
     routers.insert("fcm", state.fcm_router.active());
 
-    let mut health: HashMap<&str, serde_json::Value> = HashMap::new();
-    health.insert(
-        "status",
-        json!(if state.db.health_check().await.is_ok() {
+    let mut health = json!({
+        "status": if state
+            .db
+            .health_check()
+            .await
+            .map_err(|e| {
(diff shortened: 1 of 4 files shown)


autoendpoint/src/routes/health.rs
```diff
@@ -23,27 +23,42 @@ pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
...
+    );
+    health.insert("version", json!(env!("CARGO_PKG_VERSION").to_string()));
+    health.insert("router_table", router_health);
+    health.insert("message_table", message_health);
+    health.insert("routers", json!(routers));
 
     #[cfg(feature = "reliable_report")]
     {
-        health["reliability"] = json!(state.reliability.health_check().await.unwrap_or_else(|e| {
+        let mut reliability_health = state.reliability.health_check().await.unwrap_or_else(|e| {
```
jrconlin: With the old structure, I'm not sure that this value was being included in `health`, since it was not appearing in the result. I've switched to the more manual approach.

autoendpoint/src/routes/health.rs
```diff
@@ -23,27 +23,42 @@ pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
...
-        "status": "OK",
-        "version": env!("CARGO_PKG_VERSION"),
-        "router_table": router_health,
-        "message_table": message_health,
-        "routers": routers,
-    });
+    let mut health: HashMap<&str, serde_json::Value> = HashMap::new();
+    health.insert(
+        "status",
+        json!(if state.db.health_check().await.is_ok() {
```
jrconlin: Yeah, we really should have been checking for the database connection here.

pjenvey (commented):
The problem is https://updates-autopush.stage.mozaws.net/health is too old (v1.74.3) for the new health check w/ `reliability`

autopush-common/src/reliability.rs
```diff
@@ -146,46 +146,71 @@ impl PushReliability {
...
+                    let key = format!("{}#{}", id, new);
+                    let _ = pipe.zadd(EXPIRY, &key, expr.unwrap_or_default());
+                    trace!("🔍 internal record result: {:?}", key);
+                }
+                let _ = pipe.exec_async(conn).await.inspect_err(|e| {
+                    warn!("🔍 Redis internal storage error: {:?}", e);
                 });
-            trace!("🔍 internal record result: {:?}", cc);
-        }
+                Ok(None)
```
pjenvey: I copied the original `transaction` API but note it's a bit weird: it's really built to always be called with `query_async` instead of `exec_async` -- `Ok(None)` actually signals it to retry the entire transaction over again (I guess the intention was something from `query` returning `None` at the end is always assumed to be a particular read failure? not sure as I've never used it this way).

Take a look at the other example in this file, you probably want something like:
```rust
    let _: () = crate::redis_util::transaction(
    ...
    Ok(pipe.query_async(conn).await.inspect_err(|e| {
        warn!("🔍 Redis internal storage error: {:?}", e);
    })?)
    <the end of the transaction>

File: autoendpoint/src/routes/health.rs [hunk 4]
```diff
@@ -33,15 +46,37 @@ pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
...
-        health["reliability"] = json!(state.reliability.health_check().await.unwrap_or_else(|e| {
-            state
-                .metrics
-                .incr_with_tags(MetricName::ReliabilityErrorRedisUnavailable)
-                .with_tag("application", "autoendpoint")
-                .send();
-            error!("🔍🟥 Reliability reporting down: {:?}", e);
-            "ERROR"
-        }));
+        let reliability_health: Result<String, ApcError> = state
```
pjenvey, line 49: Do we want this in autoconnect's health check as well?

Seems like we're not using the &str result of this `health_check` method any longer, maybe it should return a simpler `bool`.

This could simplify a bit without map/_err'ing the Result, e.g.:

```rust
    let reliability_health = if let Err(e) = state.reliability.health_check().await {
        ...
        "STORE_ERROR".to_owned()
    } else {
```

autopush-common/src/reliability.rs
```diff
@@ -146,46 +146,71 @@ impl PushReliability {
...
+            || ApcErrorKind::GeneralError("Exceeded reliability record retry attempts".to_owned()),
+            async |conn, pipe| {
+                // First, increment the new state.
+                pipe.hincr(COUNTS, new.to_string(), 1);
+                // If there is an old state, decrement it.
+                if let Some(old_state) = old {
+                    pipe.hincr(COUNTS, old_state.to_string(), -1);
+                    // remove the old state from the expiry set, if it exists.
+                    let key = format!("{}#{}", id, old_state);
+                    pipe.zrem(EXPIRY, &key);
```
pjenvey, line 200: I recall this `zrem` being explicitly killed as part of https://github.com/mozilla-services/autopush-rs/pull/896 (but can't remember the details) -- we definitely need it back?

autopush-common/src/reliability.rs
jrconlin: Ah, you're right. Looked inside of `transaction` and it does, indeed, retry on `None`. 
I can't do what you suggest, but I've got a work-around that should be fine.

autopush-common/src/reliability.rs
jrconlin: I'm not sure that was where it was removed, and I also can't remember the details, but when I tried doing a "clean" run, I saw multiple states for the same key (e.g. 
```
 7) "int_accepted#8d167fe068df4281839e6c2b7e179e5a"
 8) "1752080459"
 9) "retrieved#8d167fe068df4281839e6c2b7e179e5a"
10) "1752080459"
11) "stored#8d167fe068df4281839e6c2b7e179e5a"
12) "1752080459"
13) "transmitted#8d167fe068df4281839e6c2b7e179e5a"
14) "1752080459"
```
)
There really should only be one record for each message in the `expiry` table, indicating it's last state, since we decrement counts based on the expiration value in the `gc()` handler. Having multiple "states" could cause the counts to vary.

autopush-common/src/reliability.rs
```diff
@@ -146,46 +146,84 @@ impl PushReliability {
...
+                // We really don't care about the returned result here, but transaction
+                // retries if we return Ok(None), so we run the exec and return
+                // a nonce `Some` value.
+                // The turbo-fish is a fallback for edition 2024
+                pipe.query_async::<()>(conn)
+                    .await
+                    .inspect_err(|e| {
+                        warn!("🔍 Redis internal storage error: {:?}", e);
+                    })
+                    .map_err(ApcErrorKind::RedisError)?;
```
pjenvey: This `map_err` isn't needed

```suggestion
                    })?;
```

pjenvey (approved):

PR #992, opened 2025-06-17
feat: Add VAPID header to load testing

jrconlin:
Closes: PUSH-544


tests/load/locustfiles/locustfile.py
```diff
@@ -53,6 +55,12 @@ def _(parser: Any):
         help="AutopushUser wait time between tasks",
         default="25, 30",
     )
+    parser.add_argument(
+        "--vapid_key",
+        type=str,
+        env_var="AUTOPUSH_VAPID_KEY",
+        help="Use a VAPID key generated from this private key.",
```
pjenvey: If it's a path this and the `LocustError` message should make that clear e.g.
```suggestion
        help="Optional path to a private key for generating a VAPID key header.",
```

tests/load/locustfiles/locustfile.py
```diff
@@ -79,6 +87,18 @@ def __init__(self, environment) -> None:
         self.uaid: str = ""
         self.ws: WebSocketApp | None = None
         self.ws_greenlet: Greenlet | None = None
+        if environment.parsed_options.vapid_key:
+            try:
+                logging.info("Vapid key requested.")
+                self.vapid = Vapid02.from_pem(environment.parsed_options.vapid_key)
```
pjenvey: You're specifying a path so you should either read the file yourself to force PEM or use `from_file` (which would also allow DER)
```suggestion
                self.vapid = Vapid02.from_from(environment.parsed_options.vapid_key)
```

tests/load/locustfiles/locustfile.py
```diff
@@ -33,6 +34,7 @@
 )
 from pydantic import ValidationError
 from websocket import WebSocket, WebSocketApp, WebSocketConnectionClosedException
+from py_vapid import Vapid02
```
pjenvey, line 38: This needs to be added as a new dep in `tests/pyproject.toml`

tests/load/locustfiles/locustfile.py
```diff
@@ -232,13 +252,27 @@ def post_notification(self, endpoint_url: str) -> None:
         )
 
         record = NotificationRecord(send_time=time.perf_counter(), data=data)
+        headers = self.REST_HEADERS.copy()
+        if self.vapid:
+            logging.info("Using VAPID key for Autopush notification.")
```
pjenvey: let's avoid the copy unless we need it
```suggestion
        headers = self.REST_HEADERS
        if self.vapid:
            headers = headers.copy()
            logging.info("Using VAPID key for Autopush notification.")
```

tests/load/locustfiles/locustfile.py
```diff
@@ -79,6 +87,18 @@ def __init__(self, environment) -> None:
         self.uaid: str = ""
         self.ws: WebSocketApp | None = None
         self.ws_greenlet: Greenlet | None = None
+        if environment.parsed_options.vapid_key:
```
pjenvey: IIRC it's preferred to initialize this in the `@events.test_start.add_listener` function above so it's only done once vs for every single load test user (of which there are many)

tests/load/keys/public_key.pem
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
```
pjenvey, line 1: Do the public keys even need checking in?

tests/load/locustfiles/locustfile.py
```diff
@@ -232,13 +252,27 @@ def post_notification(self, endpoint_url: str) -> None:
...
+            host = f"{parsed.scheme}://{parsed.netloc}"
+            # The key should already be created.
+            vapid = self.vapid.sign(
+                claims={
+                    "sub": "mailto:loadtest@example.com",
+                    "aud": host,
+                    "exp": int(time.time()) + 86400,
+                }
+            )
+            headers["Authorization"] = f"WebPush {vapid['auth']}"
```
pjenvey: I think you want this as it returns the actual auth header (and vapid02 is the "Vapid" scheme not "WebPush" according to autoendpoint code)
```suggestion
            headers.extend(vapid)
```

tests/load/keys/public_key.pem
jrconlin: Yes, because the x962 key needs to be copied and stored in the webservices repo so that they match up.

tests/load/locustfiles/locustfile.py
```diff
@@ -232,13 +259,27 @@ def post_notification(self, endpoint_url: str) -> None:
...
+            # The key should already be created.
+            vapid = self.vapid.sign(
+                claims={
+                    "sub": "mailto:loadtest@example.com",
+                    "aud": host,
+                    "exp": int(time.time()) + 86400,
+                }
+            )
+            headers.update(vapid)
         self.notification_records[sha1(data.encode()).digest()] = record  # nosec
```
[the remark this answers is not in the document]
jrconlin: This change is invalidated by the above change. The `headers` are independent of both the `record` and the list of notifications sent contained in `self.notification_records`, which is only used as a check value, and thus do not need to have `header` values recorded.

tests/load/locustfiles/locustfile.py
```diff
@@ -232,13 +259,27 @@ def post_notification(self, endpoint_url: str) -> None:
...
```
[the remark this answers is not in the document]
jrconlin: Your comment is contradictory. 
Recording the `notification_record` happens after the headers are set and before the request is made, as your comment has specified. Your suggested line removal will break the test. I am disregarding this suggestion.

tests/load/keys/README.md
```diff
@@ -0,0 +1,27 @@
+# VAPID key storage directory
+
+This directory should contain a `private_key.pem` file that matches at
+least one of the `AUTOEND_TRACKING_KEYS`. Note that
+`AUTOEND_TRACKING_KEYS` uses a x962 formatted public key string.
```
pjenvey:
```suggestion
least one of the `AUTOEND__TRACKING_KEYS`. Note that
`AUTOEND__TRACKING_KEYS` uses a x962 formatted public key string.
```

tests/load/locustfiles/locustfile.py
```diff
@@ -53,11 +56,30 @@ def _(parser: Any):
         help="AutopushUser wait time between tasks",
         default="25, 30",
     )
+    parser.add_argument(
+        "--vapid_key",
+        type=str,
+        env_var="AUTOPUSH_VAPID_TEST",
```
pjenvey: load/README.md still refers to `AUTOPUSH_VAPID_KEY` (needs a rename there).

Also nit the command line argument is still `--vapid-key`, maybe you want that to change to `--vapid-test` also?

jrconlin: Switching this back to `AUTOPUSH_VAPID_KEY` since I was wrong about some git rules.

tests/load/locustfiles/locustfile.py
pjenvey: Forgot [this one](https://github.com/mozilla-services/autopush-rs/pull/992#discussion_r2153313998) -- I still think it's probably worth changing as we've done some pretty large load tests and we *probably* won't use the vapid option most of the time?

pjenvey (changes requested):

pjenvey (approved):

PR #997, opened 2025-06-23
bug: Address potential double counting reliability

jrconlin:
Closes: [PUSH-545](https://mozilla-hub.atlassian.net/jira/software/c/projects/PUSH/issues/PUSH-545)

[PUSH-545]: https://mozilla-hub.atlassian.net/browse/PUSH-545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoconnect/autoconnect-web/src/routes.rs
+++ b/autoconnect/autoconnect-web/src/routes.rs
@@ -43,9 +43,11 @@ pub async fn push_route(
             .await;
     }
     // Attempt to send the notification to the UA using WebSocket protocol, or store on failure.
+    // NOTE: Since this clones the notification, there is a potential to
+    // double count the reliability state.
     let result = app_state
         .clients
-        .notify(uaid.into_inner(), notif.clone())
+        .notify(uaid.into_inner(), notif.clone_without_reliability_state())
         .await;
     if result.is_ok() {
         #[cfg(feature = "reliable_report")]
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
@@ -225,9 +225,9 @@ impl WebPushClient {
                        "message_type" => MessageType::Ack.as_ref()
                 );
                 // Get the stored notification record.
-                let n = &mut self.ack_state.unacked_stored_notifs[pos];
-                let is_topic = n.topic.is_some();
-                debug!("✅ Ack notif: {:?}", &n);
+                let acked_notification = &mut self.ack_state.unacked_stored_notifs[pos];
+                let is_topic = acked_notification.topic.is_some();
+                debug!("✅ Ack notif: {:?}", &acked_notification);
                 // Only force delete Topic messages, since they don't have a timestamp.
                 // Other messages persist in the database, to be, eventually, cleaned up by their
                 // TTL. We will need to update the `CurrentTimestamp` field for the channel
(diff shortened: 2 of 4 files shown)


autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -225,9 +225,9 @@ impl WebPushClient {
                        "message_type" => MessageType::Ack.as_ref()
                 );
                 // Get the stored notification record.
-                let n = &mut self.ack_state.unacked_stored_notifs[pos];
-                let is_topic = n.topic.is_some();
-                debug!("✅ Ack notif: {:?}", &n);
+                let acked_notification = &mut self.ack_state.unacked_stored_notifs[pos];
```
jrconlin, line 228: Be more verbose about what this is.

autoconnect/autoconnect-web/src/routes.rs
```diff
@@ -43,9 +43,11 @@ pub async fn push_route(
             .await;
     }
     // Attempt to send the notification to the UA using WebSocket protocol, or store on failure.
+    // NOTE: Since this clones the notification, there is a potential to
+    // double count the reliability state.
     let result = app_state
         .clients
-        .notify(uaid.into_inner(), notif.clone())
+        .notify(uaid.into_inner(), notif.clone_without_reliability_state())
```
jrconlin, line 50: This could be inlined, but I'll let the compiler do that. Having the function makes it descriptive.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_server_notif.rs
```diff
@@ -43,11 +43,16 @@ impl WebPushClient {
     /// Send a Direct Push Notification to this user
     fn notif(&mut self, notif: Notification) -> Result<ServerMessage, SMError> {
         trace!("WebPushClient::notif Sending a direct notif");
+        // The notification we return here is sent directly to the client.
+        // No reliability state is recorded.
+        let response = notif.clone();
```
jrconlin, line 48: Again, being a bit more explicit about the function here.
Technically, we want the `notif` to be consumed by the `unacked_direct_notifs` list since that's where the state goes. The return value is sent to the client, so it should be a copy of the original `notif` since no state data is recorded.

autoconnect/autoconnect-web/src/routes.rs
pjenvey, line 50: what state are you seeing double counted? nulling out the state here seems like it would guard against the `old` state being double decremented

jrconlin: I'm seeing some older states like `Received` which are being double counted. My theory is that this happens because of a hand-off of a cloned item. (The original and cloned item both would have the old state `Received`, which would be decremented by the new states.)

pjenvey: I don't see how this helps the `Received` count since as you say it's an older state but I could see it being helpful with the cloning for other states

pjenvey (approved):

autoconnect/autoconnect-web/src/routes.rs
jrconlin: I _think_ there was an issue with double counting `Received` because of a hand-off through internal routing, but I agree, that's only a theory. (This is also one of the reasons that I avoid state machines whenever possible. I've had a lot of weird, subtle bugs that crop up because of some strange relationship or impact of using them. Yes, it's a dumb fear, but it's my dumb fear.)

PR #1002, opened 2025-06-30
docs: mdbook-mermaid now needed

pjenvey:
and upgrade mdbook tools

Issue PUSH-536

This failed to [build on master](https://github.com/mozilla-services/autopush-rs/actions/runs/15982630130/job/45080290720) -- I'm not sure why, going to try upgrading the tools.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -14,8 +14,8 @@ jobs:
     build:
         runs-on: ubuntu-latest
         env:
-            MDBOOK_ENV: 0.4.44
-            MERMAID_ENV: 0.14.1
+            MDBOOK_ENV: 0.4.51
+            MERMAID_ENV: 0.15.0
             DEST_DIR: /home/runner/.cargo/bin
         steps:
             - uses: actions/checkout@v4
--- a/docs/notes.md
+++ b/docs/notes.md
@@ -5,7 +5,7 @@
 To build the documentation, install mdbook
 
 ```bash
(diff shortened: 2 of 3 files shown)


.github/workflows/publish_docs.yml
```diff
@@ -14,8 +14,8 @@ jobs:
     build:
         runs-on: ubuntu-latest
         env:
-            MDBOOK_ENV: 0.4.44
```
jrconlin, line 17: The [error](https://github.com/mozilla-services/autopush-rs/actions/runs/15982630130/job/45080290720#step:5:16) cites that the preprocessor is running 0.4.36. I'm not sure where that preprocessor is or what container it might be in. Is it possible to isolate that and figure out the dependency or update the dependency in the pre-processor container?

jrconlin: Ah, 
`https://github.com/mozilla-services/autopush-rs/actions/runs/15982630130/job/45080290720#step:5:14`
`home/runner/work/autopush-rs/autopush-rs/docs/output/mermaid.min.js` is not found in the source. 

Maybe a missed check-in?

pjenvey: I'm assuming `mdbook-mermaid` should be handling that (it does locally) -- the github action already installs it, but it's never been utilized until now

pjenvey: So I'm guessing/hoping this is just due to old and or mismatched versions

PR #1004, opened 2025-07-01
chore: Clippy updates for 1.88.0

jrconlin:
clippy now complains about non-inline variables in `format!()` macros


pjenvey (approved):
Hopefully `cargo clippy --fix` handled it all

jrconlin:
It did not. Fortunately, they were pretty straight forward to fix.

PR #1005, opened 2025-07-03
bug: Store message state in Redis to solve for some reliability state management issues

jrconlin:
Move the individual message state to Redis. This is to attempt to stop some reliability counts from trending negative. This is after the code paths were scrutinized to ensure that the struct cloning was minimized, and some states (notably, ones where there was a hand-off between active services) continued to trend negative. 

Unfortunately, the currently available version of Redis is 7.2, so many of the auto-expiring records (like [HSETEX](https://redis.io/docs/latest/commands/hsetex/) are not present, meaning that we'll have to use more keys, and use traditional `SET` with an `EX`piration set. Fortunately, that should not be a problem, since there appears to be plenty of available memory and CPU on the active Redis servers. 

The new approach _only_ updates counts if the prior state matches what has been last recorded. 

Closes: PUSH-546


autopush-common/src/reliability.rs
```diff
@@ -617,6 +804,125 @@ mod tests {
         Ok(())
     }
     // */
+    //*
```
gruberb: Left over comments?

autopush-common/src/reliability.rs
```diff
@@ -239,26 +292,66 @@ impl PushReliability {
...
+                    .last()
+                    .map(|v| {
+                        v.as_sequence()
+                            .unwrap_or_default()
+                            .last()
+                            .unwrap_or(&redis::Value::Nil)
+                    })
+                    .unwrap_or(&redis::Value::Nil);
+                trace!("🔍 state_key set returned: {:?}", &rcheck);
+                //*
```
gruberb: Left over comment

autopush-common/src/reliability.rs
```diff
@@ -239,26 +292,66 @@ impl PushReliability {
...
+                if new != ReliabilityState::Received {
+                    if *rcheck != redis::Value::Nil {
+                        Ok(Some(redis::Value::Okay))
+                    } else {
+                        warn!("🔍⚠️ state_key set returned nil value, hope that was an error");
+                        // Something odd went on, retry.
+                        Ok(None)
```
gruberb: How can we tell that's it's going to retry here?

autopush-common/src/reliability.rs
```diff
@@ -239,26 +292,66 @@ impl PushReliability {
...
+                    }
+                } else {
+                    // Received will never have a prior value, so will return nil. Sadly,
+                    // if this fails, it will also return nil, so presume all is well.
+                    Ok(Some(redis::Value::Okay))
+                }
+                // */
+                // Ok(Some(redis::Value::Okay))
```
gruberb: Remove comment

autopush-common/src/reliability.rs
```diff
@@ -211,25 +225,64 @@ impl PushReliability {
...
+
+        let state_key = format!("state.{id}");
+        trace!("🔍 state key: {}", &state_key);
+        // The first state is special, since we need to create the `state_key`.
+        if new == ReliabilityState::Received {
+            trace!("🔍 Creating new record");
+            // we can't perform this in a transaction because we can only increment if the set succeeds.
+            // Create the new `state.{id}` key if it does not exist, and set the expiration.
+            let options = redis::SetOptions::default()
+                .with_expiration(redis::SetExpiry::EX(expr.unwrap_or(0)))
```
gruberb: Expiry to `0`, does it mean it expires right away? Should this be a `const` or `enum` instead of this magic number?

autopush-common/src/reliability.rs
```diff
@@ -239,26 +292,66 @@ impl PushReliability {
...
+                // The last element returned from the command is the result of the pipeline.
+                // The last of which should be the result of the `SET` command, which has `GET`
+                // set. This should either return the prior value or `Ok` if things worked, else
+                // it should return `nil`.
+                let rcheck = result
+                    .as_sequence()
+                    .unwrap_or_default()
+                    .last()
+                    .map(|v| {
+                        v.as_sequence()
```
gruberb: There is a lot of `default` and `unwrap_or()`. If we need to debug, it's probably hard to figure out from this code. Can we add comments or add explicit defaults?

autopush-common/src/reliability.rs
```diff
@@ -495,40 +592,111 @@ mod tests {
...
-            Box::new(Arc::new(db)),
+            db_box.clone(),
             &metrics,
             crate::redis_util::MAX_TRANSACTION_LOOP,
         )
         .unwrap();
-        pr.record(&Some(test_id.clone()), new, &None, Some(expr))
-            .await;
+        // `.record()` uses a pool, so we can't pass the moc connection directly.
+        // Instead, we're going to essentially recreate what `.record()` does calling
```
gruberb, line 715: Comment

gruberb (commented):
LGTM! Just some questions and nits.

autoendpoint/src/extractors/notification.rs
```diff
@@ -234,6 +234,10 @@ impl Notification {
                 &self.reliable_state,
                 Some(self.timestamp + self.headers.ttl as u64),
             )
-            .await;
+            .await
+            .inspect_err(|e| {
+                warn!("🔍⚠️ Unable to record reliability state log: {:?}", e);
```
taddes, line 239: A thought I have here is wondering about how much / how frequent the log emissions might be here and in related `inspect_err` method calls. Might that get spammy and hard to debug in cases of lots of problems with recording state?

autopush-common/src/reliability.rs
jrconlin: Yes, `0` means no Expiry. I'll label the number. Thanks!

autopush-common/src/reliability.rs
jrconlin: Heh, this is a bit of internal construct work for `Redis::transaction`. Returning a `None` here forces the retry. I'll add a comment.

autopush-common/src/reliability.rs
jrconlin: Not quite.

I use a shorthand way to enable or disable code by using

```
//*
...
// */
```
What's useful in this approach is that I can remove the leading `/` and instantly comment that block. 

That said, they're no longer useful, so I'm pulling them from this PR.

autopush-common/src/reliability.rs
jrconlin: Heh, yeah, I need to finish what I was talking about here. Thanks!

autopush-common/src/reliability.rs
jrconlin: I'm not sure. The code is basically ...`.as_sequence::<Option<Vec<_>>>().unwrap_or_default()`... which would return an empty `Vec`. Ultimately, the result is either a value or `redis::Value::Nil`. 

I can add comments for folk that may not be familiar with rust constructs, but this is a fairly common pattern.

autopush-common/src/reliability.rs
taddes: I had the same thought if it might be nice to represent that as an `enum` as well. Thoughts?

autopush-common/src/reliability.rs
```diff
@@ -178,12 +179,14 @@ impl PushReliability {
             );
             match pool.get().await {
                 Ok(mut conn) => {
-                    let _ = self
-                        .internal_record(&mut conn, old, new, expr, id)
+                    self.internal_record(&mut conn, old, new, expr, id)
                         .await
-                        .inspect_err(|e| {
+                        .map_err(|e| {
```
taddes, line 185: I like it using `map_err` here, it's a good improvement.

autopush-common/src/reliability.rs
```diff
@@ -617,6 +804,125 @@ mod tests {
...
+            crate::redis_util::MAX_TRANSACTION_LOOP,
+        )
+        .unwrap();
+        let _ = pr
+            .internal_record(&mut conn, &Some(stored), new, Some(expr), &test_id)
+            .await;
+
+        Ok(())
+    }
+    // */
```
taddes: lil comment dude here all alone, but yah, saw for easy commenting the block

autopush-common/src/reliability.rs
```diff
@@ -566,12 +734,31 @@ mod tests {
             .ignore()
             .cmd("ZADD")
             .arg(EXPIRY)
+            .arg(1)
```
taddes: I would say same clarifying comment regarding status, since I guess this means expiry?

taddes (commented):
Just some nits as well, thanks for discussing the approach a bit in our 1 on 1, really great work on this

autoendpoint/src/extractors/notification.rs
jrconlin: We don't want errors going to Sentry, or causing Push to fail. This captures the sub-errors that could have occurred and writes them to the Standard Error log. There's the potential that we might want to log a metric for these, but they should be fairly rare, but easily discoverable using BigQuery.

autopush-common/src/reliability.rs
jrconlin: Heh, well, essentially, it's moving the error processing up the chain a bit. We still don't want an error in this process causing Push to fail to deliver the message. (If it's a larger system error, there are other traps for it.)

tests/load/keys/README.md
```diff
@@ -0,0 +1,27 @@
+# VAPID key storage directory
```
pjenvey: stray file from the other PR

jrconlin: Not really, these PRs need to build off of each other. 

Including it here is harmless.

autopush-common/src/reliability.rs
pjenvey: I'm not sure about the handling of the result and the retry logic here -- a couple things:

1) FYI you can call `.ignore()` on the tale end of all calls you don't care about to make the return type simpler. pipeline's query result will still essentially be a Vec of the results, but if you're only expecting one you can cast it to a single element tuple e.g. (`let check: Option<(Option<redis::Value>,)> = ..`)

2) You really only want to trigger the retry logic when the `EXEC` returns nil -- if I'm reading the [pipeline code correctly (`query_async` calls this in `atomic` mode)](https://github.com/redis-rs/redis-rs/blob/ecb96fe84e8cd09c97cf1975ff44d539c9c662c6/redis/src/pipeline.rs#L178), it seems like only the outter `Value` from `query_async` would be that case, whereas the inner `Value` would always be the actual result of the `get` -- what you have here seems to mix both of them into the same `rcheck` which seems like a potential bug

3) If we can detect the need to retry the transaction in the outter result, do we even need any check on the inner `get` (`*rcheck == redis::Value::Nil`) at all? It seems like it's only support code for detecting the need to retry the transaction. If that's the case doesn't 316-350 just simplify down to `Ok(pipe.query_async(conn).await?)` (ignoring the result of the `get`)?

jrconlin: This is a case of being overly paranoid.
Redis will return `nil` if a lock fails, or if a `SET XX GET` operation fails because the value did not previously exist.
Technically, the latter should never happen because we're setting a transaction lock on both the `expiry` table and the `state.{ID}` entry, the value that we're doing the SET/GET.
I had collapsed this into a single operation partly because I didn't want to draw out the conditionals, but I can understand that it impacted readability.
I've reworked things to be clearer and more verbose, and also just recorded an `error!()` instead of rerunning the transaction, which should make things more readable.

gruberb (dismissed):
Hard to do a proper review for me, missing a lot of the business context (and technical to some degree). What I see makes sense. Cleaning up the comments as pointed out, otherwise, all okay from my side.

taddes (dismissed):
LGTM. Helped with the context you provided during our discussion. You're definately the expert on the Redis impl, so I defer to your judgement there. The test cases help understand the flow, so thanks for those.

autopush-common/src/reliability.rs
pjenvey: If the final check is solely to emit an `error!` it could alternatively go outside the transaction entirely

autopush-common/src/reliability.rs
```diff
@@ -211,25 +226,100 @@ impl PushReliability {
...
+                    warn!("🔍⚠️ Could not create state key: {:?}", e);
+                    ApcErrorKind::GeneralError("Could not create the state key".to_owned())
+                })?;
+            // We were able to create the key, so we can increment the count.
+            conn.hincr::<_, _, _, ()>(COUNTS, new.to_string(), 1)
+                .await
+                .map_err(|e| {
+                    warn!("🔍⚠️ Could not increment received count: {:?}", e);
+                    ApcErrorKind::GeneralError("Could not increment the received count".to_owned())
+                })?;
```
pjenvey: Aren't we getting 2 increments for this case, one here and one in the transaction?

autopush-common/src/reliability.rs
```diff
@@ -239,26 +332,67 @@ impl PushReliability {
...
                     .to_string();
-                    let _ = pipe.zadd(EXPIRY, &key, expr.unwrap_or_default());
+                    pipe.zadd(EXPIRY, &key, expr.unwrap_or_default());
                     trace!("🔍 internal record result: {:?}", key);
                 }
+                trace!("🔍 upping {:?}", &new);
+                // Bump up the new state count, and set the state key's state if it still exists.
+                pipe.hincr(COUNTS, new.to_string(), 1);
+                let options = redis::SetOptions::default()
+                    .conditional_set(redis::ExistenceCheck::XX)
```
pjenvey, line 348: otherwise this set will reset the TTL:
```suggestion
                    .with_expiration(redis::SetExpiry::KEEPTTL)
                    .conditional_set(redis::ExistenceCheck::XX)
```

pjenvey: Forgot this

pjenvey (approved):
I think there's some potential cleanup we could do to the transaction (like: maybe doing the initial get up front, checking it, then folding the `Received` set into the other set. also https://github.com/mozilla-services/autopush-rs/pull/1005#discussion_r2223828278) but I'm fine getting this out there and see how it fares

PR #1016, opened 2025-07-25
bug: restore state_key generation outside of pipe

jrconlin:
Also added some additional trace statements to dump out the Redis
commands to aid in debugging

Closes [PUSH-564](https://mozilla-hub.atlassian.net/browse/PUSH-564)

[PUSH-564]: https://mozilla-hub.atlassian.net/browse/PUSH-564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autopush-common/src/reliability.rs
+++ b/autopush-common/src/reliability.rs
@@ -267,7 +267,6 @@ impl PushReliability {
         // Create the initial state key.
         // Do not place this inside of the transaction. We monitor the state key and the transaction will
         // fail because the value will change before the transaction completes. Yes, really.
-        // The alternative is that we do not monitor ``
         if new == ReliabilityState::Received {
             trace!(
                 "🔍 Creating new record {state_key} ex {:?}",
--- a/autopush-common/src/reliability.rs
+++ b/autopush-common/src/reliability.rs
@@ -377,9 +377,10 @@ impl PushReliability {
                 // The last element returned from the command is the result of the pipeline.
                 // If Redis encounters an error, it will return a `nil` as well. We handle both
                 // the same (retry), so we can normalize errors as `nil`.
-                // The last of which should be the result of the `SET` command, which has `GET`
-                // set. This should either return the prior value or `Ok` if things worked, else
-                // it should return `nil`, in which case we record a soft error.
+                // The last of which should be the result of the piped command set.
+                // This should return `nil` if there is any error, in which case we record
+                // a soft error. (On success, it will return the result of the last command
+                // in the pipe, which may vary due to the current state).
                 // This could also be strung together as a cascade of functions, but it's broken
                 // out to discrete steps for readability.
                 if result == redis::Value::Nil {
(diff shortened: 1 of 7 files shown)


autopush-common/src/reliability.rs
```diff
@@ -363,16 +382,18 @@ impl PushReliability {
                 // it should return `nil`, in which case we record a soft error.
                 // This could also be strung together as a cascade of functions, but it's broken
                 // out to discrete steps for readability.
+                if result == redis::Value::Nil {
+                    warn!("🔍⚠🪈 {id} - Pipe failed, retry.");
+                    return Ok(None);
+                }
                 if let Some(operations) = result.as_sequence() {
```
pjenvey, line 390: Now that this set is conditional on `!= Received` shouldn't the handling of the result be as well? (with the last operation being hincr instead of set)

jrconlin: No. In this case, a `Nil` response would indicate that the pipe failed. Otherwise, it would return a structured response that contains the result of the queued operation.

Redis has horrible error handling.

pjenvey: I mean post this first `result == redis::Value::Nil {` check -- we're looking at the `last()` of the sequence, which would in the case of Received be the `hincr`

jrconlin: Well, while the operation has changed, the result is still functionally the same. If there is a problem with the Pipe, we would get back a `Nil`. If the pipe succeeded, we would get back the adjusted value. if the pipe failed, we would still want to note it as an error. 

So the first check is to see if the lock failed and returned Nil. The second check is to see if the internal pipe failed and returned Nil.

pjenvey: Sorry, I've made this entirely too confusing (at least to myself) by mistakenly commenting on the wrong line.. Let me try again on line 396 just so we're on the same page..

autopush-common/src/reliability.rs
```diff
@@ -363,16 +382,18 @@ impl PushReliability {
...
                     // We have responses, the first items report the state of the commands,
                     // the final line is a list of command results.
                     if let Some(result_values) = operations.last() {
                         if let Some(results) = result_values.as_sequence() {
                             // The last command should contain the prior state. If it returned `Nil`
                             // for some, unexpected reason, note the error.
-                            if new != ReliabilityState::Received
-                                && Some(&redis::Value::Nil) == results.last()
-                            {
+                            if Some(&redis::Value::Nil) == results.last() {
```
pjenvey, line 397: I meant here: the comment suggests we're checking the prior state from the `SET/GET`, but we're not when `== Received` -- we're instead checking the final call of an `HINCR`. Per the docs I guess `HINCR` never fails -- always returning the new value -- if so it wouldn't trigger this, but we could clarify this check either by adjusting the comment or keeping the previous code that wraps the check in a `new != ReliabilityState::Received` also clarifies it and the comment as is

jrconlin: Ah! I see what you mean. I'll rework the comment to be more correct and general.

PR #1029, opened 2025-09-25
bug: Remove 'hostname' tag from periodic reporter

jrconlin:
The `hostname` tag was producing too much cardinality with the metrics,
leading to the telegraf server to consume too much memory and eventually
start dropping metrics.

Closes [PUSH-578](https://mozilla-hub.atlassian.net/browse/PUSH-578)

### DIFF
--- a/autopush-common/src/db/reporter.rs
+++ b/autopush-common/src/db/reporter.rs
@@ -21,7 +21,7 @@ pub fn spawn_pool_periodic_reporter(
     });
 }
 
-fn pool_periodic_reporter(db: &dyn DbClient, metrics: &StatsdClient, hostname: &str) {
+fn pool_periodic_reporter(db: &dyn DbClient, metrics: &StatsdClient, _hostname: &str) {
     let Some(status) = db.pool_status() else {
         return;
     };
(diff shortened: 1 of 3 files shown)


autopush-common/src/db/reporter.rs
```diff
@@ -30,14 +30,12 @@ fn pool_periodic_reporter(db: &dyn DbClient, metrics: &StatsdClient, hostname: &
             "database.pool.active",
             (status.size - status.available) as u64,
         )
-        .with_tag("hostname", hostname)
+        //.with_tag("hostname", hostname)  // Do not include hostname due to cardinality
```
gruberb, line 33: Can we remove this instead of commenting out?

autopush-common/src/db/reporter.rs
```diff
@@ -21,7 +21,7 @@ pub fn spawn_pool_periodic_reporter(
     });
 }
 
-fn pool_periodic_reporter(db: &dyn DbClient, metrics: &StatsdClient, hostname: &str) {
+fn pool_periodic_reporter(db: &dyn DbClient, metrics: &StatsdClient, _hostname: &str) {
```
gruberb, line 24: Can we remove the parameter all together?

autopush-common/src/db/reporter.rs
jrconlin: Nope. This is a precautionary scar. We definitely do NOT want to replicate this problem in the future.

autopush-common/src/db/reporter.rs
jrconlin: I was thinking that I'd retain it because removing it would cascade to the calling function and we _may_ want to add a logging comment that does include the hostname. (I'll defer to pjenvey about the utility of that) It would be easier in the future to have this value available in that case.

PR #1039, opened 2025-12-10
feat: Add redis backend for Autopush

jrconlin:
This PR is taken from https://github.com/mozilla-services/autopush-rs/pull/813, which adds a Redis backend data store. The
original patch was written by [@p1gp1g](https://github.com/p1gp1g).

Note that this patch was originally created for small, stand-alone
server uses, and has not been fully load tested.

Closes [PUSH-587](https://mozilla-hub.atlassian.net/browse/PUSH-587)

[PUSH-587]: https://mozilla-hub.atlassian.net/browse/PUSH-587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autopush-common/src/db/redis/redis_client/mod.rs
+++ b/autopush-common/src/db/redis/redis_client/mod.rs
@@ -542,7 +542,6 @@ impl DbClient for RedisClientImpl {
             .into_iter()
             .filter_map(|opt: Option<String>| {
                 if let Some(m) = opt {
-                    dbg!(&m);
                     serde_json::from_str(&m)
                         .inspect_err(|e| {
                             // Since we can't raise the error here, at least record it


autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+        }
+        let messages: Vec<Notification> = con
+            .mget::<&Vec<String>, Vec<Option<String>>>(&messages_id)
+            .await?
+            .into_iter()
+            .filter_map(|opt: Option<String>| {
+                if opt.is_none() {
+                    // We return dummy expired event if we can't fetch the said event,
+                    // it means the event has expired
+                    Some(Notification {
```
gruberb: Might be very hard to track why we do this? Can we return a proper error and log/count it somewhere?

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+
+    async fn get_channels(&self, uaid: &Uuid) -> DbResult<HashSet<Uuid>> {
+        let uaid = Uaid(uaid);
+        let config = if self.settings.timeout.is_some_and(|t| !t.is_zero()) {
+            redis::AsyncConnectionConfig::new()
+                .set_connection_timeout(self.settings.timeout.unwrap())
+        } else {
+            redis::AsyncConnectionConfig::new()
+        };
+        let mut con = self
```
gruberb: Why don't we get get the already established connection like the other methods?
```
self.connection().await?
```

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+    ///
+    /// If the message contains a topic, we remove the old message
+    async fn save_message(&self, uaid: &Uuid, message: Notification) -> DbResult<()> {
+        let uaid = Uaid(uaid);
+        let mut con = self.connection().await?;
+        let msg_list_key = self.message_list_key(&uaid);
+        let exp_list_key = self.message_exp_list_key(&uaid);
+        let msg_id = &message.chidmessageid();
+        let msg_key = self.message_key(&uaid, msg_id);
+        // message.ttl is already min(headers.ttl, MAX_NOTIFICATION_TTL)
```
gruberb: why is this commented out?

autopush-common/src/db/redis/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct RedisDbSettings {
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
+    pub timeout: Option<Duration>,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
+    pub router_ttl: Option<Duration>,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
+    pub notification_ttl: Option<Duration>,
```
gruberb, line 39: This doesn't seem to be used at all?

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+impl RedisClientImpl {
+    pub fn new(metrics: Arc<StatsdClient>, settings: &DbSettings) -> DbResult<Self> {
+        debug!("🐰 New redis client");
+        let dsn = settings
+            .dsn
+            .clone()
+            .ok_or(DbError::General("Could not find DSN".to_owned()))?;
+        let client = redis::Client::open(dsn)?;
+        let db_settings = RedisDbSettings::try_from(settings.db_settings.as_ref())?;
+        info!("🐰 {:#?}", db_settings);
```
gruberb, line 84: Can we... remove the rabbit? :) Or explain why or what we want to express with it?

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+            metrics,
+            redis_opts: SetOptions::default().with_expiration(SetExpiry::EX(router_ttl_secs)),
+        })
+    }
+
+    /// Return a [ConnectionLike], which implement redis [Commands] and can be
+    /// used in pipes.
+    ///
+    /// Pools also return a ConnectionLike, so we can add support for pools later.
+    async fn connection(&self) -> DbResult<redis::aio::MultiplexedConnection> {
```
gruberb, line 103: I think we can have a race condition here. What if multiple calls at roughly the same time? They acquire the lock, no connection is there, release the lock, start creating a connection. Now call B is doing the same, but the connection has not been stored yet. 

I think we could solve this with `OneCell`?

```rust
 use tokio::sync::OnceCell;

  pub struct RedisClientImpl {
      conn: OnceCell<MultiplexedConnection>, 
      // ...
  }

  async fn connection(&self) -> DbResult<MultiplexedConnection> {
      self.conn.get_or_try_init(|| async {
          let config = /* ... */;
          self.client
              .get_multiplexed_async_connection_with_config(&config)
              .await
              .map_err(|e| DbError::ConnectionError(format!("Cannot connect to redis: {}", e)))
      }).await.map(|conn| conn.clone())
  }
```

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+                    // it means the event has expired
+                    Some(Notification {
+                        timestamp: 1,
+                        ..Default::default()
+                    })
+                } else {
+                    opt.and_then(|m| {
+                        serde_json::from_str(&m)
+                            .inspect_err(|e| {
+                                // Since we can't raise the error here, at least record it
```
gruberb: Why can't we raise the error here? "Unit tests should fail" sounds a bit dangerous to me.

Can't we do roughly something like that?

```rust
  let messages: Vec<Notification> = con
      .mget::<&Vec<String>, Vec<Option<String>>>(&messages_id)
      .await?
      .into_iter()
      .enumerate()
      .map(|(i, opt)| -> DbResult<Notification> {
          let key = &messages_id[i];
          match opt {
              None => {
                  // Missing message - could be expired or bug
                  warn!("Message key {} missing from Redis, likely expired", key);
                  // Return a proper "expired" indicator or skip it
                  Err(DbError::Integrity(
                      "Message data missing".to_string(),
                      Some(key.clone())
                  ))
              }
              Some(json) => serde_json::from_str(&json).map_err(|e| {
                  error!("Failed to deserialize message {}: {}", key, e);
                  DbError::Serialization(format!("Invalid JSON: {}", e))
              })
          }
      })
      .collect::<DbResult<Vec<_>>>()?;
```

Then we also don't have to return a dummy notification?

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+    redis_opts: SetOptions,
+}
+
+impl RedisClientImpl {
+    pub fn new(metrics: Arc<StatsdClient>, settings: &DbSettings) -> DbResult<Self> {
+        debug!("🐰 New redis client");
+        let dsn = settings
+            .dsn
+            .clone()
+            .ok_or(DbError::General("Could not find DSN".to_owned()))?;
```
gruberb: Can we be a bit more specific? Maybe:
```rust
.ok_or(DbError::General("Redis DSN not configured. Set db_dsn in settings.".to_owned()))?
```

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,923 @@
...
+        let client = redis::Client::open(dsn)?;
+        let db_settings = RedisDbSettings::try_from(settings.db_settings.as_ref())?;
+        info!("🐰 {:#?}", db_settings);
+        let router_ttl_secs = db_settings.router_ttl.unwrap_or_default().as_secs();
+        Ok(Self {
+            client,
+            conn: Arc::new(Mutex::new(None)),
+            settings: db_settings,
+            metrics,
+            redis_opts: SetOptions::default().with_expiration(SetExpiry::EX(router_ttl_secs)),
```
gruberb: The constructor sets a default redis_opts with router TTL, but `save_message` creates its own `SetOptions` with message-specific TTL. This is fine, but there's no clear documentation about when each is used.

gruberb (changes requested):
Left a view comments. At least the creation of the connections can cause a race condition, so multiple connections could be created at once I believe.

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Yep. Autopush uses emoji identifiers for logging messages. This allows us to quickly (and visually) scan for categories of errors. This can be important when looking for systemic errors. The original patch chose a rabbit (possibly because it's alliterative with "redis") I'm fine with whatever so long as it's unique.

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Huh, thanks! I don't have a lot of experience with `OnceCell`, so I didn't think of it.

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Good catch! Thanks!

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Because it's a comment?

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Yeah, looking closer at this, we really shouldn't return dummy values. There's no reason to not just return an empty set if there are no unexpired items.

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Hmm. Good catch, and interesting. Seems like we should be consistent and move the `notification_opts` into `new()`, and maybe not pass around `db_settings`.

autopush-common/src/db/redis/redis_client/mod.rs
gruberb: Ha, sorry, I mis-read this :)

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,910 @@
...
+
+        debug!("🐰 Saving message {} :: {:?}", &msg_key, &message);
+        trace!(
+            "🐰 timestamp: {:?}",
+            &message.timestamp.to_be_bytes().to_vec()
+        );
+
+        // Remember, `timestamp` is effectively the time to kill the message, not the
+        // current time.
+        let expiry = (SystemTime::now() + Duration::from_secs(message.ttl))
```
gruberb: The code calculates an absolute Unix timestamp.

  If right now is 1736300000 (Unix timestamp) and message.ttl = 3600 (1 hour), then:
  - expiry = 1736300000 + 3600 = 1736303600 (absolute timestamp when message should die)

  Then it's used with:
  SetExpiry::EX(expiry)  // line 350

  The problem: According to the https://docs.rs/redis/latest/redis/enum.SetExpiry.html:

  | Option  | Type     | Meaning                      |
  |---------|----------|------------------------------|
  | EX(n)   | Relative | Expire in n seconds from now |
  | EXAT(n) | Absolute | Expire at Unix timestamp n   |

So SetExpiry::EX(1736303600) tells Redis: "expire this key in 1,736,303,600 seconds from now" — that's ~55 years.

So I think we either use `EXAT` later on, or pass `message.ttl` to `SetExpiry:EX`.

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,910 @@
...
+    /// Delete expired messages
+    async fn increment_storage(&self, uaid: &Uuid, timestamp: u64) -> DbResult<()> {
+        let uaid = Uaid(uaid);
+        debug!("🐰🔥 Incrementing storage to {}", timestamp);
+        let msg_list_key = self.message_list_key(&uaid);
+        let exp_list_key = self.message_exp_list_key(&uaid);
+        let storage_timestamp_key = self.storage_timestamp_key(&uaid);
+        let mut con = self.connection().await?;
+        trace!("🐇 SEARCH: increment: {:?} - {}", &exp_list_key, timestamp);
+        let exp_id_list: Vec<String> = con.zrangebyscore(&exp_list_key, 0, timestamp).await?;
```
gruberb, line 410: This doesn't return the right message_id and will silently fail! This will return messages added with `ZADD`, not Redis keys I believe. Later on, `del(&exp_id_list)` is trying to delete non-existent keys.

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,910 @@
...
+        if !exp_id_list.is_empty() {
```
gruberb, line 411: I don't exactly know how to get the Redis message_ids here, but maybe we can do:
```rust
  let msg_keys: Vec<String> = exp_id_list
              .iter()
              .map(|id| self.message_key(&uaid, id))
              .collect();
```
And then on line 396, do ` .del(&msg_keys)` instead of `.del(&exp_id_list)`

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,910 @@
...
+    /// NOTE: This function is called by mobile during the daily
+    /// [autoendpoint::routes::update_token_route] handling, and by desktop
+    /// [autoconnect-ws-sm::get_or_create_user]` which is called
+    /// during the `HELLO` handler. This should be enough to ensure that the ROUTER records
+    /// are properly refreshed for "lively" clients.
+    ///
+    /// NOTE: There is some, very small, potential risk that a desktop client that can
+    /// somehow remain connected the duration of MAX_ROUTER_TTL, may be dropped as not being
+    /// "lively".
+    async fn update_user(&self, user: &mut User) -> DbResult<bool> {
```
gruberb, line 200: Might be a millisecnd problem, but technically, if a User has a mobile and desktop app open, network issues etc, we can get duplicated HELLO messages at roughly the same time. So we could have two writes, and then we could send notification out to a stale connection.

Redis I believe doesn't have a form of lock, just a `WATCH` and `UNWATCH`. I don't know if we want to introduce this here (we need a loop, it's not a whole lot more code, maybe 5-10 lines). But we should leave a comment that this can happen? Might be one of these nasty bugs which is really hard to find later.

autopush-common/src/db/redis/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+            Err(e) => Err(DbError::General(format!(
+                "Could not parse RedisDbSettings: {:?}",
+                e
+            )))?,
+        };
+        Ok(me)
+    }
+}
+
+mod tests {
```
gruberb, line 134: Needs a `#[cfg(test)]` on top.

autopush-common/src/notification.rs
```diff
@@ -9,16 +9,19 @@ use crate::util::ms_since_epoch;
 #[derive(Serialize, Default, Deserialize, Clone, Debug)]
 /// A Publishable Notification record. This is a notification that is either
 /// received from a third party or is outbound to a UserAgent.
+///
 pub struct Notification {
+    // Required values
     #[serde(rename = "channelID")]
     pub channel_id: Uuid,
     pub version: String,
+    pub timestamp: u64,
```
gruberb, line 19: This was skipped before. Does this cause issues?

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: AH! Thanks! Got those confused.

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Hmm. 
Adding some notes for myself:
the `MSG_ID` is message unique identifier constructed in `Notification::chidmessageid()`. 
[ZRangeByScore](https://redis.io/docs/latest/commands/zrangebyscore/) returns all the elements associated with a key (`autopush/msgs_exp/{UAID}`) from the start "score" (`0`) to the end "score" of the current timestamp. This should return a list of message_id "members". In `save_message()`, we save the message under "autopush/msg/{UAID}/{MSG_ID}", then `.zadd("autopush/msgs_exp/{UAID}", {MSG_ID}, {EXPIRY})` (remembering that the [rust function](https://docs.rs/redis/latest/redis/struct.Pipeline.html#method.zadd) is `zadd(key, member, score)` vs the redis [command](https://redis.io/docs/latest/commands/zadd/) of `ZADD key score member`, which is confusing), so I believe we're pulling the correct values. 

Later, we're calling `del("autopush/msg/{UAID}/{MSG_ID}")` as well as calling `zrem("autopush/msgs_exp/{UAID}", {MSG_ID})` which removes the member from the messages to expire. We also delete the `{MSG_ID}` from the list of pending message ids stored in `"autopush/msgs/{MSG_ID}"`. 

I believe we also have unit tests for this (`wipe_expired`) and integration tests (`test_ttl_expired`, `test_ttl_batch_partly_expired_and_good_one`, etc.)

Some bits of potential confusion:
There are essentially three different key "groups", each specified with a different prefix:
* `autopush/msg/{UAID}/{MSG_ID}` - holds the actual message and is referred to directly by key
* `autopush/msgs/{UAID}` - holds the list of pending message IDS for a given UAID
* `autopush/msgs_exp/{UAID}` - holds the list of message IDs as members ranked by timestamp score. 

I can see how this could be optimized a bit to have the list of pending messages and expirations combined, but I don't think that's a major concern.  

Am I misunderstanding the problem or are you seeing something different?

jrconlin: AH! I think I see the problem. Please confirm if you agree.
The problem is that the `zrangebyscore()` is returning a list of raw `{MSG_ID}` fields. This is a problem because, as noted above, we store the messages with a key prefix. This means that the actual message is never deleted. It also means that the tests are passing, because we remove the references to the message from the pending queue, but not the source message. 

Good catch! Sorry for my confusion.

autopush-common/src/notification.rs
jrconlin: It shouldn't, but there's more to it.
Since the prior storage systems built the notifications artisanally, there was nothing that just dumped the notification as straight JSON. The original author decided that they would just re-use the `Notification` struct and make `timestamp` a real field, since it would be ignored by the UserAgent. 
That's fine for small systems, but isn't a great idea, since there may be additional information that could differ from what we send out. 
I'm going to create a generic `StorableNotification` that should be used in this case.

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: I'm not sure that this is a problem.
The desktop app and the mobile app should have two different UAIDs. What _can_ happen is that a UAID could do a HELLO, then the connection is reset (and we don't get the websocket report), and the same UAID tries again. This is a VERY rare event (so much so that we stopped tracking it because after 5 years we never once got a hit).

autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -0,0 +1,982 @@
...
+    /// Pools also return a ConnectionLike, so we can add support for pools later.
+    async fn connection(&self) -> DbResult<redis::aio::MultiplexedConnection> {
+        let config = if self.timeout.is_some_and(|t| !t.is_zero()) {
+            redis::AsyncConnectionConfig::new().set_connection_timeout(self.timeout.unwrap())
+        } else {
+            redis::AsyncConnectionConfig::new()
+        };
+        Ok(self
+            .conn
+            .get_or_try_init(|| async {
```
gruberb, line 111: This makes me think: What if the connection breaks after a while, have we a way to recover?

gruberb (approved):
👍 LGTM! One thing as a comment, not exactly sure how severe it is! We can also land a second PR with a patch after this.

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Yeah, the advise is to use a connection pool for this: https://docs.rs/redis/latest/redis/#connection-pooling.

I'll add that later.

PR #1040, opened 2026-01-03
feat: Add postgres backend storage

jrconlin:
This is an update of a very old PR where I added postgres as a storage
system. It's not as fast as bigtable, but will use a lot less resident
memory than Redis

Issue: [Push-593](https://mozilla-hub.atlassian.net/browse/PUSH-593)

### DIFF
--- a/autopush-common/src/notification.rs
+++ b/autopush-common/src/notification.rs
@@ -6,6 +6,7 @@ use std::str::FromStr;
 use serde_derive::{Deserialize, Serialize};
 use uuid::Uuid;
 
+#[cfg(feature = "postgres")]
 use crate::db::error::DbError;
 use crate::util::ms_since_epoch;
 
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,5 +1,5 @@
 [advisories]
 ignore = [
     "RUSTSEC-2024-0437", # Bound by grpcio 0.13
-    "RUSTSEC-2023-0071" # Bound by jsonwebtoken 10.3.0
+    "RUSTSEC-2023-0071", # Bound by jsonwebtoken 10.3.0
 ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+<a name="1.78.0"></a>
+## 1.78.0 (2026-01-29)
+
+
+#### Bug Fixes
+
+*   updated mozcloud-publish workflow to new version that does not require artifact write permissions (#1038) ([fe5cfbe5](https://github.com/mozilla-services/autopush-rs/commit/fe5cfbe520b3e46b4d860731da8fde98d727f9cf))
+
+#### Chore
+
+*   tag 1.77.4 (#1032) ([efc0ea55](https://github.com/mozilla-services/autopush-rs/commit/efc0ea55cf23f8fa97a38fd0a01c79a1d6c8bc48))
+
+#### Features
+
+*   PUSH-600 Redis connection pooling (#1041) ([20d3b47a](https://github.com/mozilla-services/autopush-rs/commit/20d3b47acc2c9fff89466413c500211f9874dc5f))
+*   Add redis backend for Autopush (#1039) ([33398b4b](https://github.com/mozilla-services/autopush-rs/commit/33398b4b0cf2ca234acb4a152c8b8186769afd29))
+*   Convert Hardcoded TTLs to configs (#1037) ([a3d7b257](https://github.com/mozilla-services/autopush-rs/commit/a3d7b257d0574fd9d40198bb03d498989db5158b))
+*   Document local development and debug for autopush (#1034) ([68670665](https://github.com/mozilla-services/autopush-rs/commit/686706654aca30cb7e60168a9d3edf2c1ad35c67))
+*   Add smoke tests to verify autoendpoint and autoconnect images b… (#1036) ([ab4c7f87](https://github.com/mozilla-services/autopush-rs/commit/ab4c7f877bf5e8fd01776dbd792dfc8cb149abf9))
+
+
+
 <a name="1.77.4"></a>
 ## 1.77.4 (2025-10-22)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -424,12 +424,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.100"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
-
-[[package]]
 name = "arc-swap"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -512,7 +506,7 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "autoconnect"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "actix-http",
  "actix-server",
@@ -532,7 +526,7 @@ dependencies = [
 
 [[package]]
 name = "autoconnect_common"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "actix-web",
  "autopush_common",
@@ -541,7 +535,7 @@ dependencies = [
  "futures 0.3.31",
  "futures-locks",
  "hyper 1.8.1",
- "reqwest 0.12.28",
+ "reqwest 0.13.2",
  "sentry",
  "serde",
  "serde_derive",
@@ -555,7 +549,7 @@ dependencies = [
 
 [[package]]
 name = "autoconnect_settings"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "autoconnect_common",
  "autopush_common",
@@ -564,7 +558,7 @@ dependencies = [
  "fernet",
  "lazy_static",
  "mozsvc-common",
- "reqwest 0.12.28",
+ "reqwest 0.13.2",
  "serde",
  "serde_derive",
  "serde_json",
@@ -575,7 +569,7 @@ dependencies = [
 
 [[package]]
 name = "autoconnect_web"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "actix-codec",
  "actix-http",
@@ -597,7 +591,7 @@ dependencies = [
 
 [[package]]
 name = "autoconnect_ws"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "actix-http",
  "actix-rt",
@@ -623,7 +617,7 @@ dependencies = [
 
 [[package]]
 name = "autoconnect_ws_sm"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "actix-rt",
  "actix-web",
@@ -636,7 +630,7 @@ dependencies = [
  "ctor",
  "futures 0.3.31",
  "mockall",
- "reqwest 0.12.28",
+ "reqwest 0.13.2",
  "sentry",
  "serde_json",
  "slog-scope",
@@ -647,7 +641,7 @@ dependencies = [
 
 [[package]]
 name = "autoendpoint"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "a2",
  "actix-cors",
@@ -672,7 +666,7 @@ dependencies = [
  "mockito 1.7.2",
  "openssl",
  "regex",
- "reqwest 0.12.28",
+ "reqwest 0.13.2",
  "sentry",
  "serde",
  "serde_json",
@@ -690,7 +684,7 @@ dependencies = [
 
 [[package]]
 name = "autopush_common"
-version = "1.77.4"
+version = "1.78.0"
 dependencies = [
  "actix-rt",
  "actix-web",
@@ -783,6 +777,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.15.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b7b6141e96a8c160799cc2d5adecd5cbbe5054cb8c7c4af53da0f83bb7ad256"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.37.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c34dda4df7017c8db52132f0f8a2e0f8161649d15723ed63fc00c82d0f2081a"
+dependencies = [
+ "cc",
+ "cmake",
+ "dunce",
+ "fs_extra",
+]
+
+[[package]]
 name = "backtrace"
 version = "0.3.76"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -839,7 +855,7 @@ dependencies = [
  "proc-macro2",
  "quote 1.0.44",
  "regex",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "shlex",
  "which",
 ]
@@ -861,7 +877,7 @@ dependencies = [
  "proc-macro2",
  "quote 1.0.44",
  "regex",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "shlex",
  "syn 2.0.114",
  "which",
@@ -959,9 +975,9 @@ dependencies = [
 
 [[package]]
 name = "cadence"
-version = "1.6.0"
+version = "1.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3075f133bee430b7644c54fb629b9b4420346ffa275a45c81a6babe8b09b4f51"
+checksum = "d7aff0c323415907f37007d645d7499c378df47efb3e33ffc1f397fa4e549b2e"
 dependencies = [
  "crossbeam-channel",
 ]
@@ -977,6 +993,12 @@ dependencies = [
  "libc",
  "shlex",
 ]
+
+[[package]]
+name = "cesu8"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
 name = "cexpr"
@@ -1173,6 +1195,16 @@ name = "core-foundation"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
+name = "core-foundation"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
 dependencies = [
  "core-foundation-sys",
  "libc",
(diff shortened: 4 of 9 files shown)


autopush-common/src/db/postgres/mod.rs
```diff
@@ -0,0 +1,1167 @@
...
+                _metrics: metrics,
+                db_settings,
+                pool,
+            });
+        };
+        Err(DbError::ConnectionError("No DSN specified".to_owned()))
+    }
+
+    /// Does the given table exist
+    async fn table_exists(&self, table_name: String) -> DbResult<bool> {
```
gruberb, line 123: How about 

```rust
async fn table_exists(&self, table_name: &str) -> DbResult<bool> {
    let row = self
         .pool
         .get()
         .await
         .map_err(DbError::PgPoolError)?
         .query_one(
             "SELECT EXISTS (SELECT FROM pg_tables WHERE schemaname = 'public' AND tablename = $1)", &[&table_name],
           )
           .await
          .map_err(DbError::PgError)?;
     Ok(row.get(0))
}
```

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,64 @@
...
+-- This presumes that you've already created the account user (other than `postgres`)
+-- and have created a database for push to use (e.g.
+--      CREATE DATABASE autopush WITH OWNER = $OWNER;
+--      CREATE SCHEMA autopush;
+--      GRANT ALL ON SCHEMA autopush TO $OWNER;
+--      GRANT ALL ON TABLES IN SCHEMA autopush TO $OWNER;
+--      SET search_path TO autopush, public;
+-- );
+
+DROP TABLE message;
```
gruberb, line 14: Should we move this into a `drop_schema.psql`? 

And we probably want
```sql
DROP TABLE IF EXISTS message;
DROP INDEX IF EXISTS chidmessageidx;
DROP TABLE IF EXISTS meta;
DROP TABLE IF EXISTS reliability;
DROP TABLE IF EXISTS router;
```

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,64 @@
...
+CREATE TABLE IF NOT EXISTS meta (
+    channel_id varchar(81) NOT NULL,
+    uaid varchar(40) NOT NULL REFERENCES router ON DELETE cascade,
+    FOREIGN KEY(uaid) REFERENCES router (uaid),
+    PRIMARY KEY (uaid, channel_id)
+
+);
+
+-- Contains the encrypted messages pending delivery to the User Agents.
+CREATE TABLE IF NOT EXISTS message (
```
gruberb, line 57:
```sql
  CREATE TABLE IF NOT EXISTS message (
      uaid varchar(40) NOT NULL REFERENCES router(uaid) ON DELETE CASCADE,
      channel_id varchar(81),
      chid_message_id varchar(100) NOT NULL PRIMARY KEY,
      version varchar(40),
      ttl bigint,
      topic varchar,
      timestamp bigint,
      data text,
      sortkey_timestamp bigint,
      headers text
  );

CREATE INDEX IF NOT EXISTS message_uaid_idx ON message (uaid);
```

- I think `chid_message_id` should be the primary key, since each message needs a unique identifier?
- `uaid` can be foreign key, so users can have multiple messages?
- Add index for `uaid` for efficient look ups

autopush-common/src/db/postgres/mod.rs
```diff
@@ -0,0 +1,1167 @@
...
+        self.pool
+            .get()
+            .await?
+            .execute(
+                &format!(
+                    "INSERT INTO {tablename} (id, states) VALUES ($1, json_build_object($2, $3) )
+            ON CONFLICT(id)
+            UPDATE {tablename} SET states = jsonb_set(states, array[$2], to_jsonb($3))"
+                ),
+                &[&state, &timestamp],
```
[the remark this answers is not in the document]
gruberb: Yepp, I think that's correct :)

autopush-common/src/notification.rs
```diff
@@ -113,3 +116,50 @@ impl Notification {
 pub(crate) fn default_ttl() -> u64 {
     0
 }
+
+#[cfg(feature = "postgres")]
+/// Postgres Message Row to Notification (may be used as a model for other DBs later)
+impl From<&tokio_postgres::Row> for Notification {
```
gruberb: We should use `TryFrom` so we don't have to use `unwrap` in the codebase? 

```rust
impl TryFrom<&tokio_postgres::Row> for Notification {
      type Error = DbError;

      fn try_from(row: &tokio_postgres::Row) -> Result<Self, Self::Error> {
          let channel_id_str: &str = row.try_get("channel_id").map_err(DbError::PgError)?;
          let channel_id = Uuid::from_str(channel_id_str)
              .map_err(|e| DbError::General(format!("Invalid channel_id UUID: {}", e)))?;
      ...
```

autopush-common/src/db/postgres/mod.rs
```diff
@@ -0,0 +1,1167 @@
...
+                .map_err(DbError::PgError)?
+                .map(|v| v as u64),
+            current_timestamp: row
+                .try_get::<&str, Option<i64>>("last_update")
+                .map_err(DbError::PgError)?
+                .map(|v| v as u64),
+            version: row
+                .try_get::<&str, Option<String>>("version")
+                .map_err(DbError::PgError)?
+                .map(|v| Uuid::from_str(&v).unwrap()),
```
gruberb:
```rust
 .map(|v| Uuid::from_str(&v).map_err(|e| DbError::General(e.to_string())))
  .transpose()?
```

autopush-common/src/db/postgres/mod.rs
```diff
@@ -0,0 +1,1167 @@
...
+                ],
+            )
+            .await
+            .map_err(DbError::PgError)?;
+        Ok(result > 0)
+    }
+
+    /// fetch user information from router_table for uaid.
+    async fn get_user(&self, uaid: &Uuid) -> DbResult<Option<User>> {
+        let rr = self.pool.get().await.map_err(DbError::PgPoolError)?
```
gruberb: Can we do:

```rust
  let row = self.pool.get().await.map_err(DbError::PgPoolError)?
      .query_opt(
          &format!(
              "SELECT connected_at, router_type, router_data, node_id, record_version, last_update, version,
  priv_channels FROM {tablename} WHERE uaid = $1",
              tablename=self.router_table()
          ),
          &[&uaid.simple().to_string()]
      )
      .await
      .map_err(DbError::PgError)?;

  let Some(row) = row else {
      return Ok(None);
  };
```

`query_opt` returns `Option<Row>` (`None` for no rows, `Some(Row)` for exactly one row). Cleaner than the String matching?

autopush-common/src/db/routing.rs
```diff
@@ -2,27 +2,38 @@
...
+    #[cfg(feature = "bigtable")]
     BigTable,
+    #[cfg(feature = "postgres")]
+    Postgres,
+    #[cfg(feature = "redis")]
     Redis,
     None,
 }
 
 impl Default for StorageType {
```
gruberb, line 14: Should we throw a warning if multiple are configured?

```rust
     // When multiple storage backends are enabled, require explicit configuration
      #[cfg(any(
          all(feature = "bigtable", feature = "redis"),
          all(feature = "bigtable", feature = "postgres"),
          all(feature = "redis", feature = "postgres"),
      ))]
      {
          warn!("Multiple storage backends enabled. Defaulting to bigtable. Set db_dsn explicitly.");
          #[cfg(feature = "bigtable")]
          return StorageType::BigTable;
      }
```

autopush-common/src/db/postgres/mod.rs
```diff
@@ -0,0 +1,1167 @@
...
+        limit: usize,
+    ) -> DbResult<FetchMessageResponse> {
+        let messages: Vec<Notification> = self
+            .pool
+            .get()
+            .await
+            .map_err(DbError::PgPoolError)?
+            .query(
+                &format!(
+                "SELECT channel_id, version, ttl, topic, timestamp, data, sortkey_timestamp, headers FROM {tablename} WHERE uaid=$1 LIMIT $2 ORDER BY timestamp DESC", // TODO: Check the timestamp DESC here!
```
gruberb: ORDER has to come before LIMIT.

```sql
"SELECT channel_id, version, ttl, topic, timestamp, data, sortkey_timestamp, headers FROM {tablename} WHERE uaid=$1
   AND topic IS NOT NULL ORDER BY timestamp DESC LIMIT $2"
```

gruberb (commented):
A few things to fix up and tighten! But looks very solid, big PR :)

autopush-common/src/db/postgres/mod.rs
```diff
@@ -0,0 +1,1167 @@
...
+    async fn table_exists(&self, table_name: String) -> DbResult<bool> {
+        let rows = self
+            .pool.get().await.map_err(DbError::PgPoolError)?
+            .query(
+                &format!("SELECT EXISTS (SELECT FROM pg_tables where schemaname='public' AND tablename={tablename});", tablename=table_name),
+                &[],
+            )
+            .await.map_err(DbError::PgError)?;
+        let val: &str = rows[0].get(0);
+        Ok(val.to_lowercase().starts_with('t'))
```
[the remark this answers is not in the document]
jrconlin: Fair point, there's a more compact way to do this.

autoconnect/autoconnect-settings/src/lib.rs
```diff
@@ -231,34 +234,61 @@ impl Settings {
...
+            let host = env::var("POSTGRES_HOST").unwrap_or("localhost:5432".to_owned());
+            let db_dsn = Some(format!("postgres://{}", host));
+            let db_settings = "".to_string();
+            Self {
+                db_dsn,
+                db_settings,
+                ..Default::default()
+            }
         }
     }
```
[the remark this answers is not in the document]
jrconlin: A compile time error is preferred. Only one of these should be enabled. All three being enabled is a work-around.

autopush-common/src/notification.rs
```diff
@@ -113,3 +116,50 @@ impl Notification {
...
+                .unwrap_or_default(),
+            timestamp: row
+                .try_get::<&str, i64>("timestamp")
+                .map(|v| v as u64)
+                .unwrap(),
+            data: row.try_get::<&str, String>("data").map(Some).unwrap(),
+            sortkey_timestamp: row
+                .try_get::<&str, i64>("sortkey_timestamp")
+                .map(|v| Some(v as u64))
+                .unwrap_or_default(),
```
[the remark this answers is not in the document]
jrconlin: For consistency, we should instead ensure that we always save something other than NULL here.

autopush-common/src/db/postgres/schema.psql
jrconlin: I dunno. In reality, we should probably have some migration structure set up to handle this. 
I'm expecting that this will probably only be done once, on initialization, and probably, even then, be hand crafted. 
I may revisit this once I write up the documentation, though, but for now, i'm fine with it being more draconian.

autopush-common/src/db/postgres/mod.rs
jrconlin: I am tempted to say that if we have an invalid UUID we're reconstructing from stored data, it's an integrity error and we should absolutely panic.

pjenvey: There's already a `DbError::Integrity` variant specifically for this case

autopush-common/src/db/postgres/mod.rs
jrconlin: We don't want to default to `public` here. Really what I should do is check if the schema is present in the tablename and use that, else fall back to `public`. 
Good catch.

autopush-common/src/db/routing.rs
jrconlin: Moved some logic to `build.rs` to trap for this sort of thing.

autopush-common/src/db/postgres/mod.rs
jrconlin: Ah! Right! forgot about that one. Thanks!

autopush-common/src/db/postgres/schema.psql
jrconlin: It's true that there is a need for an index for chid_message_id, but UAID is the primary way that messages are searched.

pjenvey: I'm a bit rusty on the details here but the access patterns I believe are messages searched in a couple ways:

- reading topic messages (with a limit)
- reading timestamp messages (with a limit), with an offset of the last timestamp read
- deleting a message (just for topic messages, at least in some backends)

`UAID` should be the sole primary key for users, but messages should have additional data to identify the specific message. I would think you'd want something more a long the lines of a composite key of `UAID`/`chid_message_id`? You want `channel_id` in it, but `channel_id` is already included in the current version of `chid_message_id` -- but it really depends on the other details of the postgres backend which I haven't dug into. It's probably more for deletes than the larger reads (but that also depends on how that's handled on postgres).

Here's a couple higher level views of the other schemas would could be helpful (I was intending to include the DynamoDB schema in our docs, despite deprecating that backend, but never got the chance):

Bigtable: https://docs.google.com/spreadsheets/d/1wK1lhrsLVvP_doVnXeQ4sSYSCfKU2jrHDLyFNoH9Jp8/edit?gid=0#gid=0

DynamoDB:
- Router
router

- Message
message

in both cases `uaid`/`chid_message_id` are some kind of composite key, either as Bigtable's "key" (the impl usually calls it `row_key`) or DynanoDB's primary key (a "composite key" composed of a hash + range key, the only kind of composite Dynamo originally offered, apparently [it's now more flexible](https://aws.amazon.com/about-aws/whats-new/2025/11/amazon-dynamodb-multi-attribute-composite-keys-global-secondary-indexes/))

autopush-common/src/db/postgres/schema.psql
pjenvey: There's https://github.com/rust-db/refinery and tern, it might make more sense to just switch to a higher level db lib that offers migrations.

With you already using raw queries, sqlx would probably be a good choice since it's made for that (and hopefully wouldn't be a huge effort to integrate?). The enterprise team's using it for their axum based admin console.

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,75 @@
...
+    router_type varchar(100),
+    router_data text,
+    last_connect bigint,
+    node_id varchar(100),
+    record_version bigint,
+    version varchar(40),    -- Optional and probably not needed
+    last_update bigint,     -- Note "current_timestamp" seems to be reserved.
+    expiry bigint,
+    priv_channels varchar(40)[]
+);
```
chenba, line 35: Why don't we use more appropriate(?) Postgres types, e.g. uuid and date/time types instead of varchar and ints?

autopush-common/src/db/postgres/mod.rs
```diff
@@ -0,0 +1,1251 @@
...
+        let response: Vec<Row> = if let Some(ts) = timestamp {
+            trace!("📮 Fetching messages for user {} since {}", &uaid, ts);
+            self.pool
+                .get()
+                .await
+                .map_err(DbError::PgPoolError)?
+                .query(
+                    &format!(
+                        "SELECT * FROM {} 
+                         WHERE uaid = $1 AND timestamp > $2
```
chenba: I was going to suggest maybe adding a composite index for this.  But then I remembered that `uaid` is the primary key, so the `timestamp` condition here doesn't make sense?

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,75 @@
...
+
+-- This contains "meta" or general information for the routing table.
+-- It may also contain general setup and configuration data for autopush.
+-- Right now, it contains the channel_ids associated with the UAID.
+CREATE TABLE IF NOT EXISTS meta (
+    channel_id varchar(81) NOT NULL,
```
chenba, line 41: `channel_id` appears to be a UUID.  I don't understand why this is a `varchar(81)`.

autopush-common/src/db/postgres/schema.psql
jrconlin: Fair questions. 
The UUID one is a bit tricky because we've been burned by that in the past. Although the ChannelID is a UUID, we need to preserve whatever the string representation is that the client hands us, because the client doesn't have a way to convert strings to UUIDs cleanly. (We could definitely use UUID for UAIDs, since the server technically "owns" them.

Right now, I'm basically treating Postgres more as a template for whoever wants to build a SQL database storage engine, so I'm going more for readability rather than optimization. I could probably solve for that by adding lots of documentation into things, but I'm also trying to keep this fairly low-effort until we have demand for it.

autopush-common/src/db/postgres/mod.rs
jrconlin: Well, we still only want to grab messages that are later than the provided timestamp. (Timestamp, for messages, is the time that the message was provided. The server sends the messages in chronological order, in batches, so that we want to know where to "pick up" from later.)

You do raise a good point that there should be a check on `expiry` here to make sure that we're not sending expired messages. Thanks!

autopush-common/src/db/postgres/schema.psql
jrconlin: This gets into the "We need to preserve the string that we get back from the UA" problem. `81` is VERY generous, though. I should be consistent and limit this to 40 like I do in other places.

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,75 @@
...
+--      CREATE SCHEMA autopush;
+--      GRANT ALL ON SCHEMA autopush TO $OWNER;
+--      GRANT ALL ON TABLES IN SCHEMA autopush TO $OWNER;
+--      SET search_path TO autopush, public;
+-- );
+
+SET search_path TO autopush, public;
+
+DROP TABLE message;
+DROP INDEX chidmessageidx;
```
ncloudioj, line 15: Nit:
```suggestion
DROP INDEX chid_message_idx;
```

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,75 @@
...
+    priv_channels varchar(40)[]
+);
+
+-- This contains "meta" or general information for the routing table.
+-- It may also contain general setup and configuration data for autopush.
+-- Right now, it contains the channel_ids associated with the UAID.
+CREATE TABLE IF NOT EXISTS meta (
+    channel_id varchar(81) NOT NULL,
+    uaid varchar(40) NOT NULL REFERENCES router ON DELETE cascade,
+    FOREIGN KEY(uaid) REFERENCES router (uaid),
```
ncloudioj, line 43:
```suggestion
    uaid varchar(40) REFERENCES router(uaid) ON DELETE cascade,
```

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,75 @@
...
+CREATE TABLE IF NOT EXISTS reliability (
+    id varchar(40) NOT NULL PRIMARY KEY,
+    states jsonb,
+    last_update_timestamp bigint
+);
+
+
+-- Contains the encrypted messages pending delivery to the User Agents.
+CREATE TABLE IF NOT EXISTS message (
+    uaid varchar(40) NOT NULL PRIMARY KEY REFERENCES router ON DELETE cascade,
```
ncloudioj, line 58: `uaid` serves both as a primary key and a foreign key, which feels a little odd. Usually we'd keep it as a foreign key and create an index for it for fast lookup and delete.

Based on Phil's comment above, it sounds like we can have a composite primary key on `uaid` and `chid_message_id` and then leave `uaid` as a foreign key for cascad delete. Presumably, the `message` table could store lots of records, this composite key can be used for message lookup and cascade delete.

autopush-common/src/db/postgres/schema.psql
ncloudioj: You might consider creating an index on `uaid`, otherwise, the cascade delete may take excessively long when this table grows big.

autopush-common/src/db/postgres/schema.psql
```diff
@@ -0,0 +1,75 @@
...
+DROP INDEX chidmessageidx;
+DROP INDEX message_reliability_id_idx;
+DROP INDEX update_idx;
+DROP TABLE meta;
+DROP TABLE router;
+DROP TABLE reliability;
+
+-- This table contains the routing information for push message recipients.
+CREATE TABLE IF NOT EXISTS router (
+    uaid varchar(40) NOT NULL PRIMARY KEY UNIQUE,
```
ncloudioj, line 24: Notice that we have update methods defined on this table, would we ever update `uaid` by any chance? If so, we'd also add "ON UPDATE CASCADE" to the tables referencing this key.

ncloudioj (approved):
Unsolicited visitor :)

It looks good to me, a few comments for discussion.

Since this is an experimental feature, I'd vote for merging it as is now (we can call this out in the doc and flag that the db schema is subject to changes&breakages) and polish it as we go.

autopush-common/src/db/postgres/schema.psql
jrconlin: Ah! It was my understanding that the `FOREIGN KEY` was also a table index. I'll add that as a follow up PR. Thanks!

autopush-common/src/db/postgres/schema.psql
jrconlin: We shouldn't update UAID. (Well, we do, but we generate a new UAID in that case.)

PR #1041, opened 2026-01-13
feat: PUSH-600 Redis connection pooling

jrconlin:
Redis offers a multiplexed connection which will allow for traffic over a single connection. There are some questions about how this handles a connection drop. The reliability reporting system uses a connection pool. For consistency, switch to using a pooled connection for the data store as well.

* lock to redis 0.32 until `deadpool-redis` updates. ( _Note: we may want to switch to the multiplexed_async_connector if deadpool decides to stop updating_ )
* in addition, update the docker container with some quality of life improvements: 
  * Bring in .ssh directory, 
  * adds `direnv` for easy environment setting

Closes: PUSH-600

### DIFF
--- a/autopush-common/src/db/redis/mod.rs
+++ b/autopush-common/src/db/redis/mod.rs
@@ -30,7 +30,7 @@ use uuid::Uuid;
 pub struct RedisDbSettings {
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
-    pub create_timeout: Option<Duration>, 
+    pub create_timeout: Option<Duration>,
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
     pub router_ttl: Option<Duration>,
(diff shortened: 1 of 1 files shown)


autopush-common/src/db/redis/redis_client/mod.rs
```diff
@@ -79,16 +76,26 @@ impl RedisClientImpl {
...
         let db_settings = RedisDbSettings::try_from(settings.db_settings.as_ref())?;
         info!("🐰 {:#?}", db_settings);
         let router_ttl_secs = db_settings.router_ttl.unwrap_or_default().as_secs();
         let notification_ttl_secs = db_settings.notification_ttl.unwrap_or_default().as_secs();
+
+        let config = Config::from_url(dsn);
+        let pool = config
+            .builder()
+            .map_err(|e| DbError::General(format!("Could not create Redis pool: {:?}", e)))?
+            .create_timeout(db_settings.timeout)
```
gruberb: Should we document the `timeout` in `db_settings` that this is for CREATE and not for connection timeout now?

gruberb (approved):
LGTM 👍

autopush-common/src/db/redis/redis_client/mod.rs
jrconlin: Yep, let me apply a quick update to rename this to something more specific

PR #1060, opened 2026-01-28
docs: add documentation for new data stores

jrconlin:
Closes PUSH-605

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "f279fc8b1f1a64138f0f4b9cda9be488ae35bc2f8556c7ffe60730f1c07d005a"
 dependencies = [
  "base64 0.21.7",
  "erased-serde 0.3.31",
- "http 1.3.1",
+ "http 1.4.0",
  "http-body-util",
  "hyper 1.8.1",
  "hyper-rustls 0.26.0",
@@ -127,8 +127,8 @@ version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
- "quote 1.0.42",
- "syn 2.0.110",
+ "quote 1.0.44",
+ "syn 2.0.114",
 ]
 
 [[package]]
@@ -218,7 +218,7 @@ dependencies = [
  "actix-utils",
  "futures-core",
  "http 0.2.12",
- "http 1.3.1",
+ "http 1.4.0",
  "impl-more",
  "pin-project-lite",
  "tokio",
@@ -238,9 +238,9 @@ dependencies = [
 
 [[package]]
 name = "actix-web"
-version = "4.12.0"
+version = "4.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2233f53f6cb18ae038ce1f0713ca0c72ca0c4b71fe9aaeb59924ce2c89c6dd85"
+checksum = "1654a77ba142e37f049637a3e5685f864514af11fcbc51cb51eb6596afe5b8d6"
 dependencies = [
  "actix-codec",
  "actix-http",
@@ -273,7 +273,7 @@ dependencies = [
  "serde_json",
  "serde_urlencoded",
  "smallvec",
- "socket2 0.6.1",
+ "socket2 0.6.2",
  "time",
  "tracing",
  "url",
@@ -287,15 +287,15 @@ checksum = "f591380e2e68490b5dfaf1dd1aa0ebe78d84ba7067078512b4ea6e4492d622b8"
 dependencies = [
  "actix-router",
  "proc-macro2",
- "quote 1.0.42",
- "syn 2.0.110",
+ "quote 1.0.44",
+ "syn 2.0.114",
 ]
 
 [[package]]
 name = "actix-ws"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a3a1fb4f9f2794b0aadaf2ba5f14a6f034c7e86957b458c506a8cb75953f2d99"
+checksum = "12d4f2fbee3ef7a22fa6cb0e416b962237a167ed0419f22d4e451da2d7f082f8"
 dependencies = [
  "actix-codec",
  "actix-http",
@@ -431,9 +431,12 @@ checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
 name = "arc-swap"
-version = "1.7.1"
+version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
+checksum = "9ded5f9a03ac8f24d1b8a25101ee812cd32cdc8c50a4c50237de2c4915850e73"
+dependencies = [
+ "rustversion",
+]
 
 [[package]]
 name = "arraydeque"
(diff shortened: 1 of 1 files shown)


docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
+# Data Storage Systems
+
+## Introduction
+
+*PLEASE NOTE* Autopush only used Bigtable in production. Any other datastore is considered experimental and has only been used in small, test environments. Please use caution if you decide to use an alternative data storage system. As always, you're improvements and suggestions are welcome.
```
ncloudioj: s/you're/your

docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Storage Options
+
+Please note: configuration options will presume that you are using configuration files. Remember, all configuration options can be specified using environment variables. See the sample configuration options for details.
+
+From what we've observed, the general usage pattern for our system tends to be that a large number of users create a subscription, then those subscriptions go "idle". The bulk of our messages go into storage and are never retrieved. Internally, we consider this the "long storage" problem.
+
+As an additional "bonus" Publishers frequently ignore our 404/410 responses and continue to publish to subscription URLs that are no longer valid. These invalid subscriptions still require a DB lookup, so you may wish to also add rules to temporarily block IPs that have too many 404/410 responses.
```
ncloudioj: s/"bonus"/"bonus", /

docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
...
+
+This is the production storage engine that is used by Mozilla. To use it, compile the autopush code with `--features=bigtable`.
+
+It offers reasonable cost for the sort of scale that we see, and the performance of the data system is very good. While probably not advisable for any form of reliable production work, you can experiment with [using the Bigtable emulator](bigtable-emulation.md) fairly easily.
+
+#### Configuration
+The Bigtable DSN begins with the prefix `grpc` and points to the host and port of the Bigtable database (e.g. to point to a local emulator you would specify `grpc://localhost:8086`)
+
+Bigtable's database configuration options are stored in the `db_settings` as a serialized JSON string. The settings include:
+* `message_family` the Bigtable cell family that refers to the message data collections
```
ncloudioj: Maybe `s/the message data/the **message** data/` to be consistent with the above format.
Ditto for "router data collections" below.

docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
...
+The `scripts/setup_bt.sh` file contains a collection of commands that can be used to configure Bigtable storage, or at least aid in setting up the table.
+
+### Redis/Valkey
+
+Redis is a memory based key/value storage system. Valkey is an Open Source, API equivalent, "drop-in" replacement for Redis. For simplicity, I will refer to both systems as "Redis". To use it, compile the autopush code with `--features=redis`.
+
+Redis is very responsive, but also quite memory intensive. While initial load tests have shown surprising results, it should be noted that there is a drastic difference between the load test environment and reality. (It can be prohibitively expensive to replicate our traffic in a test environment, so our testing system focuses on a more aggressive use model at the expense of productions "long storage" stress.)
```
ncloudioj: > a more aggressive use model at the expense of productions "long storage" stress.

Hmm, this doesn't read clearly to me. What does "a more aggressive use model" refer to? Is it a specific test scenario in which most subscriptions do not have the "long storage" problem observed in production?

docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
...
+* `create_timeout`: number of seconds to wait to create a new connection to Redis
+* `router_ttl`: the number of seconds for a router record to live. The default is `MAX_ROUTER_TTL`, or `60 * 86,400` (60 days). Whenever the UAID connects, this TTL is reset. Only records for UAIDs that have not connected in 60 days are dropped.
+* `notification_ttl`: the maximum number of seconds for a message to live. Messages live by their
+provided `TTL` header. Messages that specify a `TTL` greater than this number will be set to this limit. By default, the `MAX_NOTIFICATION_TTL_SECS` value of `30 * 86,400` (30 days).
+
+Redis does not require any prep work before using.
+
+### Postgres
+
+Postgres is a SQL storage engine which has several features that allow it to perform similar to a Schemaless engine. While it is not as performant as some schemaless systems, it is also a fairly common data storage system which may be more familiar to some. To use it, compile the autopush code with `--features=postgres`.
```
ncloudioj: s/Schemaless/schemaless/

docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
...
+Redis' database configuration options are stored in `db_settings` as a serialized JSON string. These are optional and default values will be used.
+The settings include:
+* `create_timeout`: number of seconds to wait to create a new connection to Redis
+* `router_ttl`: the number of seconds for a router record to live. The default is `MAX_ROUTER_TTL`, or `60 * 86,400` (60 days). Whenever the UAID connects, this TTL is reset. Only records for UAIDs that have not connected in 60 days are dropped.
+* `notification_ttl`: the maximum number of seconds for a message to live. Messages live by their
+provided `TTL` header. Messages that specify a `TTL` greater than this number will be set to this limit. By default, the `MAX_NOTIFICATION_TTL_SECS` value of `30 * 86,400` (30 days).
+
+Redis does not require any prep work before using.
+
+### Postgres
```
ncloudioj: Nit: s/Postgres/PostgreSQL/

docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
...
+The postgres DSN begins with the prefix `postgres` and can contain the login information and schema (e.g. `postgres://dbuser:hunter2@localhost:5432/autopush`)
+
+Postgres' database configuration options are stored in `db_settings` as a serialized JSON strong.
+These are optional and default values will be used.
+The settings include:
+* `schema`: The name of the postgres schema to use (default `public`)
+* `router_table`: the name of the router table (default `router`)
+* `message_table`: the name of the message table (default `message`)
+* `meta_table`: the name of the table to store additional routing information for the user (default `meta`)
+* `reliability_table`: the name of the table to store reliability tracking information (default `reliability`)
```
ncloudioj: Is `reliability_table` specific to the Postgre backend? Didn't mention this config for BigTable and Redis.

docs/src/datastore.md
```diff
@@ -0,0 +1,76 @@
...
+* `max_router_ttl`: the number of seconds for a router record to live. The default is `MAX_ROUTER_TTL`, or `60 * 86,400` (60 days). Whenever the UAID connects, this TTL is reset. Only records for UAIDs that have not connected in 60 days are dropped.
```
ncloudioj, line 74: Does Postgres support `notification_ttl` as Redis?

docs/src/datastore.md
jrconlin: Good catch, I'll fix up the table name references elsewhere.

docs/src/datastore.md
jrconlin: "More aggressive" is what I'm calling the tests preference for a high rate of short term message storage and retrieval. We don't run "stress tests" where the test runs for longer than an hour or so. "Long storage" refers to messages that are either held in storage for extended periods (up to 30 days), or are simply never read because the UA has gone away. 

I will add some clarification text.

docs/src/datastore.md
jrconlin: Currently, Reliability requires redis. I'll add some clarity around that.

docs/src/datastore.md
jrconlin: Sort of?
Each notification can specify it's own TTL, up to the `MAX_NOTIFICATION_TTL_SECS` (~30 days). Redis & PostgreSQL also offers the ability to reduce the maximum router TTL in case that sites may want to not retain older router records as long, since that's one of the larger consumers of storage in our system, and there's a higher likelihood of that impacting "stand-alone" / "enterprise" users. I didn't specify the same option for a Message TTL  override for PostgreSQL since it's less of a memory consumption concern at this time. We may decide to add it later.

PR #1070, opened 2026-02-19
Bug/push 627 postgres

jrconlin:
docs: Document all configuration options in standard format

* Removed unused option
* Added rustdoc comments
* Added standard sort `-c` CLI option

Closes: [PUSH-622](https://mozilla-hub.atlassian.net/browse/PUSH-622)

---

bug: Address post PoC commit of Postgres support

* Handle "" in topics, which could falsely identify them as topic
  messages
* Correct the `remove_message` function to use the `chid_message_id`
* Add more Postgres error types for better formatting, identification
* Move `default` to the PostgresDbSettings to capture proper behaviors
* Correct psql var orders to be clearer.
* Change index in load-test to use string instead of byte array (because
  weird problems)
* Fix load test to look for correct key value
* make sure that load tests spot duplicate values (which is an error)

Closes: [PUSH-627](https://mozilla-hub.atlassian.net/browse/PUSH-627)

[PUSH-622]: https://mozilla-hub.atlassian.net/browse/PUSH-622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PUSH-627]: https://mozilla-hub.atlassian.net/browse/PUSH-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
@@ -226,7 +226,7 @@ impl WebPushClient {
                 // Some storage engines may set this to "".
                 let is_topic = acked_notification
                     .topic
-                    .clone()
+                    .as_ref()
                     .map(|t| !t.is_empty())
                     .unwrap_or(false);
                 debug!("✅ Ack notif: {:?}", &acked_notification);
--- a/autopush-common/src/db/postgres/mod.rs
+++ b/autopush-common/src/db/postgres/mod.rs
@@ -222,16 +222,16 @@ impl DbClient for PgClientImpl {
                     expiry=EXCLUDED.expiry
                     ;
             ", tablename=self.router_table()),
-            &[&user.uaid.simple().to_string(),              // 1 
+            &[&user.uaid.simple().to_string(),              // 1
             &(user.connected_at as i64),                    // 2
-            &user.router_type,                              // 3    
+            &user.router_type,                              // 3
             &json!(user.router_data).to_string(),           // 4
             &user.node_id,                                  // 5
             &user.record_version.map(|i| i as i64),    // 6
             &(user.version.map(|v| v.simple().to_string())),   // 7
             &user.current_timestamp.map(|i| i as i64), // 8
             &user.priv_channels.iter().map(|v| v.to_string()).collect::<Vec<String>>(),
-            &(self.router_expiry() as i64),                 // 10    
+            &(self.router_expiry() as i64),                 // 10
             ]
         ).await.map_err(|e| {
             DbError::PgDbError(Self::error_to_string(&e))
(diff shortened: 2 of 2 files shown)


docs/src/config_options.md
```diff
@@ -0,0 +1,117 @@
...
+| -c | --config _config/file/path_ | Path to a configuration file (e.g. `config.toml`) to use instead of environment variables |
+
+## Autoconnect
+
+The following configuration options can be specified either in the configuration file or by
+using their environment variable form.
+
+| **option** | **env var** | **type** | **default value** | **description** |
+|---|---|---|---|---|
+| actix_max_connections | AUTOCONNECT__ACTIX_MAX_CONNECTIONS | num | _None_ | Maximum number of concurrent connections for the Actix web server |
```
pjenvey: I'll note we added spans here so you can link to each env var directly, e.g.: https://github.com/mozilla-services/syncstorage-rs/blob/master/docs/src/config.md#SYNC_HOST 

https://raw.githubusercontent.com/mozilla-services/syncstorage-rs/refs/heads/master/docs/src/config.md

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -223,17 +223,28 @@ impl WebPushClient {
                 );
                 // Get the stored notification record.
                 let acked_notification = &mut self.ack_state.unacked_stored_notifs[pos];
-                let is_topic = acked_notification.topic.is_some();
+                // Some storage engines may set this to "".
+                let is_topic = acked_notification
+                    .topic
+                    .clone()
```
gruberb: nit: we could use `.as_ref()` instead of `.clone()` here.

autopush-common/src/db/postgres/mod.rs
```diff
@@ -831,7 +857,19 @@ impl DbClient for PgClientImpl {
         // Replace this with a proper health check.
         let client = self.pool.get().await.map_err(DbError::PgPoolError);
         let row = client?.query_one("select true", &[]).await;
-        Ok(!row?.is_empty())
+        if row?.is_empty() {
```
gruberb: `query_one()` returns a single `Row`. `Row::is_empty()` checks if the row has zero columns. For `select true`, this will always be `false`  (the row has 1 column). So this check is effecitevely dead code. It still works because of the `?` I believe. We can just use the `query_one` directly and just relying on the `?` to catch failures, or do `row?.try_get::<_, bool>(0)?` to validate the result.

gruberb (approved):
Just two minor things

autopush-common/src/db/postgres/mod.rs
jrconlin: I was mostly just looking to see if _any_ result was returned, since the query is kind of a nonce. The `is_empty()` was mostly just to check that there wasn't something unexpected in the response. Checking the actual response is a reasonable thing to do, so I'll add that.

PR #1072, opened 2026-02-20
feat: Make sentry reporting optional

jrconlin:
Introduces a config new flag `disable_sentry` (default `false`)
If enabled, sentry errors are reported via `error!()` to STDERR.

Closes #[PUSH-623](https://mozilla-hub.atlassian.net/browse/PUSH-623)

[PUSH-623]: https://mozilla-hub.atlassian.net/browse/PUSH-623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


autopush-common/src/middleware/sentry.rs
```diff
@@ -16,14 +16,16 @@ pub struct SentryWrapper<E> {
     metrics: Arc<StatsdClient>,
     metric_label: String,
     phantom: PhantomData<E>,
+    sentry_enabled: bool,
 }
 
 impl<E> SentryWrapper<E> {
-    pub fn new(metrics: Arc<StatsdClient>, metric_label: String) -> Self {
+    pub fn new(metrics: Arc<StatsdClient>, metric_label: String, sentry_enabled: bool) -> Self {
```
gruberb: That's negated. So callers pass `sentry_enabled: true`, but the settings have `disable_sentry`. So a `true` would actually enable sentry.

autoendpoint/src/settings.rs
```diff
@@ -75,6 +75,9 @@ pub struct Settings {
     /// The label to use for statsd metrics.
     pub statsd_label: String,
 
+    /// Do not report errors to sentry, instead log them to STDERR.
+    pub disable_sentry: Option<bool>,
```
gruberb: This seems complex. `disable_sentry` has three states right now. Why not just a `bool` and `#[serde(default)]` (which is `false`).

autopush-common/src/middleware/sentry.rs
```diff
@@ -136,6 +145,23 @@ where
     }
 }
 
+/// Log the Sentry event to STDERR. (This is copied from syncstorage-rs)
+fn log_event(event: &Event) {
+    let exception = event.exception.last();
```
gruberb, line 150: Do we really just want the `last` exception?

autopush-common/src/middleware/sentry.rs
```diff
@@ -136,6 +145,23 @@ where
...
+    let error_value = exception
+        .and_then(|e| e.value.as_deref())
+        .unwrap_or("No error message");
+    error!("{}", error_value;
+        "error_type" => error_type,
+        "tags" => ?event.tags,
+        "extra" => ?event.extra,
+        "url" => event.request.as_ref().and_then(|r| r.url.as_ref()).map(|u| u.to_string()).unwrap_or_default(),
+        "method" => event.request.as_ref().and_then(|r| r.method.as_deref()).unwrap_or_default(),
+        "stacktrace" => exception.and_then(|e| e.stacktrace.as_ref()).map(|s| format!("{:?}", s.frames)).unwrap_or_default(),
```
gruberb, line 161: Debug format is going to be very hard to read. Can we do something more structured?

autopush-common/src/middleware/sentry.rs
```diff
@@ -70,8 +74,8 @@ where
 
     actix_web::dev::forward_ready!(service);
 
+    /// Set up the hub to add request data to events
     fn call(&self, sreq: ServiceRequest) -> Self::Future {
```
gruberb, line 78: Since the doc comment is on a trait impl method, it won't be rendered. We could keep it inside the function via `//`. And it's about the implementation, so I think that should be fine?

autopush-common/src/middleware/sentry.rs
jrconlin: Fair, Going to rename things to `sentry_disabled` so that it's more consistent in the code.

autopush-common/src/middleware/sentry.rs
jrconlin: I believe we do. The exception stack contains the current thread states, with the last being the one that is most relevant to the error. We are dumping the stacktrace, so we should be able to see the core error. This is also very similar to the behavior that Sentry does.

autopush-common/src/middleware/sentry.rs
jrconlin: Whatever we dump to STDERR is going to be hard to read, even formatted. Dumping like this has the slight benefit that it appears on one line, which will be useful for capture.

PR #1077, opened 2026-02-25
Fix memory spike: add mimalloc, reqwest pool limits, and delivery met…

gruberb:
Fix memory spike: add mimalloc, reqwest pool limits, and delivery metrics

Added performance metrics to make future incidents diagnosable from dashboards alone (no logging, no privacy impact):
  - `notification.route_time`: full route_notification path
  - `notification.direct.delivery_time`: send_notification HTTP call
  - `notification.direct.delivery_status`: tagged counter (200/404/timeout/error)
  - `notification.storage.save_time`: db.save_message duration
  - `request.in_flight`: gauge of concurrent node requests (10s interval)

### DIFF
--- a/autoendpoint/src/routers/webpush.rs
+++ b/autoendpoint/src/routers/webpush.rs
@@ -46,10 +46,7 @@ impl Router for WebPushRouter {
         Ok(HashMap::new())
     }
 
-    async fn route_notification(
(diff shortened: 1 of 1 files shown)


jrconlin (approved):
Thanks!

PR #1078, opened 2026-02-26
perf: Comprehensive performance audit fixes

gruberb:
**DISCLAIMER** I did a comprehensive analysis via various LLMs (Claude, ChatGPT and Gemini) and did a deep dive into performance for autopush. These are the issues these tools came up with. Not sure if we want all of these, or just some. Just wanted to bring these up for discussion.

Critical fixes:
- Remove duplicate VAPID JWT validation that ran the same expensive ES256 signature verification twice per request (subscription.rs)
- Use Backtrace::new_unresolved() to defer symbol resolution until backtraces are actually read, avoiding ~1-5μs per error (error.rs)
- Replace futures_locks::RwLock with DashMap for ClientRegistry, eliminating global lock contention on connect/notify/disconnect operations across thousands of concurrent WebSocket connections
- Use bounded channels (capacity 128) instead of unbounded channels for client notifications, providing backpressure for slow consumers
- Remove unnecessary User clone in webpush router hot path (borrow instead of clone-then-borrow)
- Add to_common_notification() to avoid cloning the full Notification (including User with router_data HashMap) for DB storage

High impact fixes:
- Replace HashMap<&str, serde_json::Value> double-serialization with a typed NotificationForDelivery struct for single-pass serialization
- Implement batch INSERT for Postgres save_messages instead of N sequential INSERT queries during WebSocket disconnect
- Add circuit breaker pattern to BigTable gRPC operations to prevent retry storms during outages (5 consecutive failures -> 30s cooldown)
- Replace Vec-based ACK tracking with HashMap<(ChannelId, Version)> for O(1) notification lookup/removal instead of O(n)
- Add tokio::sync::Semaphore (capacity 64) to bound concurrent disconnect-cleanup task spawning during disconnect storms

Incremental improvements:
- Use HashSet instead of Vec for BroadcastSubs client subscriptions, making change_count_delta lookups O(1) instead of O(n)
- Cache fully-qualified table names in PgClientImpl to avoid format!/clone allocation on every database query

### DIFF
--- a/autoconnect/autoconnect-common/src/registry.rs
+++ b/autoconnect/autoconnect-common/src/registry.rs
@@ -14,6 +14,7 @@ const NOTIFICATION_CHANNEL_CAPACITY: usize = 128;
 #[derive(Debug)]
 struct RegisteredClient {
     /// The user agent's unique ID.
+    #[allow(dead_code)]
     pub uaid: Uuid,
     /// The local ID, used to potentially distinquish multiple UAID connections.
     pub uid: Uuid,
@@ -32,17 +33,16 @@ impl ClientRegistry {
     ///
     /// For now just registers internal state by keeping track of the `client`,
     /// namely its channel to send notifications back.
-    pub fn connect(
-        &self,
-        uaid: Uuid,
-        uid: Uuid,
-    ) -> mpsc::Receiver<ServerNotification> {
+    pub fn connect(&self, uaid: Uuid, uid: Uuid) -> mpsc::Receiver<ServerNotification> {
         trace!("ClientRegistry::connect");
         let (tx, snotif_stream) = mpsc::channel(NOTIFICATION_CHANNEL_CAPACITY);
         let client = RegisteredClient { uaid, uid, tx };
         if let Some(old_client) = self.clients.insert(uaid, client) {
             // Drop existing connection
-            let result = old_client.tx.clone().try_send(ServerNotification::Disconnect);
+            let result = old_client
+                .tx
+                .clone()
+                .try_send(ServerNotification::Disconnect);
             if result.is_ok() {
                 debug!("ClientRegistry::connect Ghosting client, new one wants to connect");
             }
@@ -83,7 +83,10 @@ impl ClientRegistry {
     /// The client specified by `uaid` has disconnected.
     pub fn disconnect(&self, uaid: &Uuid, uid: &Uuid) -> Result<()> {
         trace!("ClientRegistry::disconnect");
-        let client_exists = self.clients.get(uaid).is_some_and(|client| client.uid == *uid);
+        let client_exists = self
+            .clients
+            .get(uaid)
+            .is_some_and(|client| client.uid == *uid);
         if client_exists {
             self.clients.remove(uaid).expect("Couldn't remove client?");
             return Ok(());
--- a/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
+++ b/autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
@@ -159,10 +159,7 @@ impl WebPushClient {
     /// Disconnect this `WebPushClient` from the `ClientRegistry`
     pub fn registry_disconnect(&self) {
         // Ignore disconnect (Client wasn't connected) Errors
-        let _ = self
-            .app_state
-            .clients
-            .disconnect(&self.uaid, &self.uid);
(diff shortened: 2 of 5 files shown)


autoconnect/autoconnect-common/src/registry.rs
```diff
@@ -1,14 +1,15 @@
...
-use futures_locks::RwLock;
 use uuid::Uuid;
 
 use autopush_common::errors::{ApcErrorKind, Result};
 use autopush_common::notification::Notification;
 
 use crate::protocol::ServerNotification;
 
+/// Maximum number of buffered notifications per client before backpressure.
+const NOTIFICATION_CHANNEL_CAPACITY: usize = 128;
```
jrconlin, line 11: This should be a setting, if we ever plan on adjusting this (I can see where we would want to)

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
```diff
@@ -237,6 +243,13 @@ impl WebPushClient {
         let uaid = self.uaid;
         let connected_at = self.connected_at;
         rt::spawn(async move {
+            let _permit = match DISCONNECT_SEMAPHORE.acquire().await {
+                Ok(permit) => permit,
+                Err(_) => {
+                    warn!("Disconnect semaphore closed, skipping save of unacked direct notifs");
```
jrconlin, line 246: I need to look at this more, but we definitely don't want to discard unacked direct notifications.

Ok, if I understand correctly, this is trapping a violent disconnect event where we've lost connection to the database, so we cannot save any unacked messages. 
This is a significant error since we are basically discarding undelivered messages, and those will NOT be delivered. For now, we should do more than just issue a `warn!()` that will probably never be read. Let's create a metric for the time being, but we should consider what we can do to avoid being in this situation.

autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/mod.rs
```diff
@@ -347,13 +360,17 @@ pub struct SessionStatistics {
     existing_uaid: bool,
 }
 
+/// Key for looking up notifications in the ACK tracking maps.
+/// Uses (channel_id, version) as the unique identifier.
+type AckKey = (Uuid, String);
```
jrconlin, line 362: `version` here is the `Notification.message_id` which is a fernet encrypted compilation of the `uaid`, `channel_id` and `topic|timestamp`. It's globally unique. Possibly even more so because `fernet` introduces padding that creates a unique value for the same content. 
(Noting as a further optimization, since we should never have the same `version` for different `channel_id` values.

autoendpoint/src/extractors/subscription.rs
```diff
@@ -148,15 +148,6 @@ impl FromRequest for Subscription {
             trace!("user: {:?}", &user);
             validate_user(&user, &channel_id, &app_state).await?;
 
-            // Validate the VAPID JWT token and record the version
-            if let Some(vapid) = &vapid {
-                validate_vapid_jwt(vapid, &app_state.settings, &app_state.metrics)?;
-
-                app_state
```
jrconlin: note: this is already done @81

autoendpoint/src/routers/webpush.rs
```diff
@@ -48,17 +48,15 @@ impl Router for WebPushRouter {
         mut notification: Notification,
     ) -> ApiResult<RouterResponse> {
         // The notification contains the original subscription information
-        let user = &notification.subscription.user.clone();
-        // A clone of the notification used only for the responses
-        // The canonical Notification is consumed by the various functions.
+        let uaid = notification.subscription.user.uaid;
```
jrconlin, line 51: I can see where we'd want to remove the more verbose items later, but we use `notification.subscription.user` as a root pretty often. Would it make more sense to keep a reference to `&notification.subscription.user` and derive from that instead (`&notif_user`) or something?

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -445,8 +517,14 @@ impl BigTableClientImpl {
         &self,
         req: bigtable::MutateRowRequest,
     ) -> Result<(), error::BigTableError> {
+        if !self.circuit_breaker.allow_request() {
+            return Err(error::BigTableError::Status(
```
jrconlin, line 521: Add a metric?

gruberb:
Will open separate PRs

PR #1079, opened 2026-02-26
perf: Add circuit breaker for BigTable gRPC calls

gruberb:
Introduce a lightweight circuit breaker around `mutate_row` and `read_rows` to fail fast when BigTable is unhealthy, avoiding request pile-ups behind slow/dead gRPC connections.

The breaker trips after consecutive failures and auto-resets after a cooldown period. A new `error.bigtable.circuit_breaker` metric tracks trips.

### DIFF
--- a/autoconnect/autoconnect-common/src/broadcast.rs
+++ b/autoconnect/autoconnect-common/src/broadcast.rs
@@ -10,7 +10,7 @@
 ///
 /// see api discussion: https://docs.google.com/document/d/1Wxqf1a4HDkKgHDIswPmhmdvk8KPoMEh2q6SPhaz4LNE/edit#
 ///
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use serde_derive::{Deserialize, Serialize};
 use strum_macros::{AsRefStr, Display};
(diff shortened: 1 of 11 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -538,8 +623,17 @@ impl BigTableClientImpl {
         &self,
         req: ReadRowsRequest,
     ) -> Result<BTreeMap<RowKey, row::Row>, error::BigTableError> {
+        if !self.circuit_breaker.allow_request() {
+            self.metrics
```
jrconlin: Going to guess the AI doesn't know about the `ReportableError` trait we use.

Inside of `autopush-common::middleware::sentry::SentryWrapperMiddleware` we check to see if the error is reportable to Sentry or if it should be returned as a metric. 
I'm thinking that we might flood sentry with these, so we probably don't want them getting reported, but we do want them recorded as a metric. 

We should probably fix up `autopush-common::db::bigtable::bigtable_client::error::ReportableError` to contain this info so that happens automatically.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -445,8 +517,17 @@ impl BigTableClientImpl {
         &self,
         req: bigtable::MutateRowRequest,
     ) -> Result<(), error::BigTableError> {
+        if !self.circuit_breaker.allow_request() {
+            self.metrics
```
jrconlin: Same as below, we want to fix up `ReportableError` to use this `metric_label` and not report this error to Sentry, to avoid flooding. (We've had problems where we overload Sentry in the past.)

PR #1080, opened 2026-02-26
perf: Replace futures-locks with DashMap registry and bounded channels

gruberb:
- Replace `futures-locks::RwLock<HashMap>` with `DashMap` for the `ClientRegistry`, eliminating async lock contention on every connect/disconnect/notify
- Switch from `mpsc::unbounded` to bounded channels with configurable `client_channel_capacity` (default 50) to apply backpressure and cap per-client memory
- All registry methods become synchronous, removing `.await` from callers

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/error.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/error.rs
@@ -135,12 +135,17 @@ pub enum BigTableError {
 
     #[error("BigTable config error: {0}")]
     Config(String),
+
+    #[error("Circuit breaker open: BigTable temporarily unavailable")]
+    CircuitBreakerOpen,
 }
 
 impl BigTableError {
     pub fn status(&self) -> StatusCode {
         match self {
-            BigTableError::PoolTimeout(_) => StatusCode::SERVICE_UNAVAILABLE,
+            BigTableError::PoolTimeout(_) | BigTableError::CircuitBreakerOpen => {
+                StatusCode::SERVICE_UNAVAILABLE
+            }
             BigTableError::Status(e, _) => e.status(),
             _ => StatusCode::INTERNAL_SERVER_ERROR,
         }
@@ -151,7 +156,7 @@ impl ReportableError for BigTableError {
     fn is_sentry_event(&self) -> bool {
         #[allow(clippy::match_like_matches_macro)]
         match self {
-            BigTableError::PoolTimeout(_) => false,
+            BigTableError::PoolTimeout(_) | BigTableError::CircuitBreakerOpen => false,
             _ => true,
         }
     }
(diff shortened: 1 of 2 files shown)


autoconnect/autoconnect-common/src/registry.rs
```diff
@@ -82,18 +99,20 @@ impl ClientRegistry {
...
         trace!("ClientRegistry::disconnect");
-        let mut clients = self.clients.write().await;
-        let client_exists = clients.get(uaid).is_some_and(|client| client.uid == *uid);
+        let client_exists = self
+            .clients
+            .get(uaid)
+            .is_some_and(|client| client.uid == *uid);
         if client_exists {
-            clients.remove(uaid).expect("Couldn't remove client?");
+            self.clients.remove(uaid).expect("Couldn't remove client?");
```
jrconlin: This should be a Sentry reportable error. (Yes, it's an edge case, but if it happens, we want it to be trackable, and not lost in a sea of logging)

autoconnect/autoconnect-common/src/registry.rs
```diff
@@ -51,14 +69,14 @@ impl ClientRegistry {
...
         trace!("ClientRegistry::notify");
-        let clients = self.clients.read().await;
-        if let Some(client) = clients.get(&uaid) {
+        if let Some(client) = self.clients.get(&uaid) {
             debug!("ClientRegistry::notify Found a client to deliver a notification to");
             let result = client
                 .tx
-                .unbounded_send(ServerNotification::Notification(notif));
+                .clone()
+                .try_send(ServerNotification::Notification(notif));
```
jrconlin: Hmm... The logic here has changed subtly.
Previously, we presumed that with an unlimited number of channels available, a send would most likely succeed. Now, we have the potential of hitting an over-allocation event where a "backoff" may occur. That's not quite the same as "User not connected". We may want to return different errors in that case, since we may want to retry an over-allocation vs a channel failure.

autoconnect/autoconnect-common/src/registry.rs
```diff
@@ -68,11 +86,10 @@ impl ClientRegistry {
...
 
     /// A check for notification command has come for the uaid
-    pub async fn check_storage(&self, uaid: Uuid) -> Result<()> {
+    pub fn check_storage(&self, uaid: Uuid) -> Result<()> {
         trace!("ClientRegistry::check_storage");
-        let clients = self.clients.read().await;
-        if let Some(client) = clients.get(&uaid) {
-            let result = client.tx.unbounded_send(ServerNotification::CheckStorage);
+        if let Some(client) = self.clients.get(&uaid) {
+            let result = client.tx.clone().try_send(ServerNotification::CheckStorage);
```
jrconlin: Same as above. Let's use different error returns (possibly new `ApcErrorKind`s?) to differentiate the types of errors that are now occurring. Added bonus: we can use the `ReportableError` trait to make them non-sentry errors, with specific metrics so we know if we need to bump up the `DEFAULT_CHANNEL_CAPACITY`

PR #1081, opened 2026-02-26
perf: HashMap ACK tracking, batch inserts, HashSet broadcasts

gruberb:
- Replace Vec-based ACK tracking with HashMap keyed on notification version for O(1) lookup instead of O(n) linear scan
- Add disconnect semaphore (capacity 64) to bound concurrent cleanup tasks during disconnect storms
- Switch BroadcastSubs from Vec to HashSet for O(1) membership checks
- Batch save_messages in postgres with cached table names

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -519,6 +519,7 @@ dependencies = [
  "autopush_common",
  "docopt",
  "env_logger 0.11.8",
+ "mimalloc",
  "sentry",
  "serde",
  "slog-scope",
@@ -662,6 +663,7 @@ dependencies = [
  "hex",
  "jsonwebtoken",
  "lazy_static",
+ "mimalloc",
  "mockall",
  "mockito 1.7.2",
  "openssl",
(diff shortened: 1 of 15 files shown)


autoconnect/autoconnect-ws/autoconnect-ws-sm/src/identified/on_client_msg.rs
```diff
@@ -253,10 +240,11 @@ impl WebPushClient {
                         .record_reliability(&self.app_state.reliability, notif.reliability_state())
                         .await;
                 }
-                let _n = self.ack_state.unacked_stored_notifs.remove(pos);
                 #[cfg(feature = "reliable_report")]
                 if !is_topic {
-                    self.ack_state.acked_stored_timestamp_notifs.push(_n);
+                    self.ack_state
```
jrconlin, line 245: NOT FOR THIS PR:
I still _REALLY_ want to delete ack'd messages from storage. The batching was more a product of DynamoDB than Bigtable, and I think there may be folk abusing this "feature".

PR #1082, opened 2026-02-26
perf: Remove duplicate VAPID validation, unnecessary clones in autoendpoint

gruberb:
- Remove duplicate `validate_vapid_jwt` call in subscription extractor
- Use `Backtrace::new_unresolved()` instead of `Backtrace::new()` to defer expensive symbol resolution until actually needed
- Remove `User` clone in webpush router; use `notif_user` reference
- Add `to_common_notification()` typed serialization to avoid re-serializing the full `Notification` struct on every delivery

### DIFF
--- a/autoconnect/autoconnect-common/src/broadcast.rs
+++ b/autoconnect/autoconnect-common/src/broadcast.rs
@@ -10,7 +10,7 @@
 ///
 /// see api discussion: https://docs.google.com/document/d/1Wxqf1a4HDkKgHDIswPmhmdvk8KPoMEh2q6SPhaz4LNE/edit#
 ///
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use serde_derive::{Deserialize, Serialize};
 use strum_macros::{AsRefStr, Display};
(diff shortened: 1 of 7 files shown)


autoendpoint/src/extractors/notification.rs
```diff
@@ -184,43 +207,58 @@ impl Notification {
         self.headers.topic.is_some()
     }
 
+    /// Convert to a common Notification by borrowing, avoiding a full clone
```
jrconlin: This is what the `autoendpoint::extractors::notification::From<Notification>` should be doing. We could replicate this by adding it as  `From<&Notification>` (or maybe replacing the existing `From` with it)

autoendpoint/src/extractors/notification.rs
```diff
@@ -8,9 +8,32 @@ use autopush_common::util::{b64_encode_url, ms_since_epoch, sec_since_epoch};
 use cadence::CountedExt;
 use fernet::MultiFernet;
 use futures::{future, FutureExt};
+use serde::Serialize;
 use std::collections::HashMap;
 use uuid::Uuid;
 
+/// Wire format for delivering notifications to connection servers.
+/// Uses a single serialization pass instead of building a HashMap of serde_json::Values.
+#[derive(Debug, Serialize)]
```
jrconlin, line 17: Technically, this is a duplicate of `autopush_common::notification::Notification`, which is what we use to handle incoming and outbound push subscription messages. (Yes, we're terrible at naming things). 

One possible option is that we consider renaming one of the `Notification` structs to be something a bit more clear, like `StoredNotification` or `TransportNotification` so that it's easier for folk to differentiate.

PR #1101, opened 2026-03-02
feat: include discrete build for enterprise

jrconlin:
Closes: #[PUSH-619](https://mozilla-hub.atlassian.net/browse/PUSH-619)

[PUSH-619]: https://mozilla-hub.atlassian.net/browse/PUSH-619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -582,7 +582,7 @@ workflows:
           image: autoconnect-enterprise
           crate: autoconnect
           binary: autoconnect
-          build_args: \"--no-default-features --features=redis\"
+          build_args: "\"--no-default-features --features=redis\""
           filters:
             tags:
               only: /.*/
@@ -591,7 +591,7 @@ workflows:
           image: autoendpoint-enterprise
           crate: autoendpoint
           binary: autoendpoint
-          build_args: \"--no-default-features --features=redis\"
+          build_args: "\"--no-default-features --features=redis\""
           filters:
             tags:
               only: /.*/


.circleci/config.yml
```diff
@@ -467,12 +470,18 @@ jobs:
...
+        type: string
+        default: ${GCP_GAR_REPO}
     steps:
       # gcr-auth parameters:
       # https://circleci.com/developer/orbs/orb/circleci/gcp-gcr#commands-gcr-auth
       - gcp-gcr/gcr-auth:
           gcp_cred_config_file_path: ~/gcp_cred_config.json
-          google-project-id: GCP_GAR_PROJECT_ID
+          google-project-id: <<parameters.project_id>>  #GCP_GAR_PROJECT_ID
           google_project_number: GCP_OIDC_PROJECT_NUMBER
```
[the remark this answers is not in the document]
jrconlin: This is an existing pattern, so keeping it for now.

.circleci/config.yml
```diff
@@ -335,6 +337,8 @@ jobs:
         type: string
       binary:
         type: string
+      build_args:
+        type: string
```
gruberb, line 345: I think we have to add a `default: ""` here, because existing,  non-enterprise jobs will fail (on CircleCI, a string parameter without a default is required).

.circleci/config.yml
```diff
@@ -410,6 +415,8 @@ jobs:
         type: string
       path:
         type: string
+      build_args:
+        type: string
```
gruberb, line 424: Same here, we have to add a `default`.

.circleci/config.yml
```diff
@@ -435,6 +442,8 @@ jobs:
...
```
gruberb, line 452: Same here, we have to add a `default`.

.circleci/config.yml
```diff
@@ -590,10 +623,13 @@ workflows:
             tags:
               only: /.*/
 
-      # Comment out the following three sections for local CircleCI testing.
+      # Comment out the following five sections for local CircleCI testing.
+      # vvvv
       - deploy:
           name: deploy-autoconnect
           image: autoconnect
+          project_id: ${GCP_GAR_PROJECT_ID}
```
gruberb: This matches the defaults. I don't think we need to add `project_id` and `repo` here. Or, we add them to all deploy jobs.

.circleci/config.yml
```diff
@@ -467,12 +476,18 @@ jobs:
...
+      repo:
+        type: string
+        default: ${GCP_GAR_REPO}
     steps:
       # gcr-auth parameters:
       # https://circleci.com/developer/orbs/orb/circleci/gcp-gcr#commands-gcr-auth
       - gcp-gcr/gcr-auth:
           gcp_cred_config_file_path: ~/gcp_cred_config.json
-          google-project-id: GCP_GAR_PROJECT_ID
+          google-project-id: <<parameters.project_id>>  #GCP_GAR_PROJECT_ID
```
gruberb: Previously, we had `GCP_GAR_PROJECT_ID` as an ENV variable. Now, the `<<parameters.project_id>>` resolves to a shell-expanded value (`${GCP_GAR_PROJECT_ID}`). Which will fail. Not sure how to solve this differently or why we chose the `paramter` path here.

.circleci/config.yml
```diff
@@ -630,6 +666,45 @@ workflows:
...
+            - Unit Tests
+            - Integration Tests
+            - Rust Formatting Check
+          filters:
+            tags:
+              only: /.*/
+            branches:
+              only:
+                - master
+                - enterprise
```
gruberb, line 694: We just run this on `enterprise`, but do the `Rust Formatting Check` and `Integration Tests` run on this new branch? If not, it will be locked waiting for a job to run which is not triggered.

.circleci/config.yml
```diff
@@ -555,6 +570,24 @@ workflows:
           filters:
             tags:
               only: /.*/
+      - build:
+          name: build-autoconnect-enterprise
+          image: autoconnect-enterprise
+          crate: autoconnect
+          binary: autoconnect
+          build_args: "--no-default-features --features=redis"
```
gruberb: So no reliability reporting for enterprise? Did we document this somewhere?

.circleci/config.yml
jrconlin: Perhaps I'm confused.
It's my understanding that `parameters.project_id` would use the CI job parameters, defaulting to the environment variable `${GCP_GAR_PROJECT_ID}`. We specify the project and repo parameters on job invocations (See line 631+), so wouldn't those values take priority? Would it be clearer to not have a default value specified here which would trigger an error?

.circleci/config.yml
jrconlin: Enterprise does not need to do reliability tracking. It is not an included feature for the build. It is also meaningless in the context of Enterprise.

.circleci/config.yml
jrconlin: We should add them to all deploy jobs. 

IIUC, this circleci file uses job templates pretty heavily, switching the function based on the passed parameters. Since `project_id` and `repo` are very significant at this time (since I am now attempting to differentiate targets) I am being explicit.

.circleci/config.yml
jrconlin: IIUC, `rust-checks` and `test-integration` run on all commits.

.circleci/config.yml
gruberb: The issue is that the `google-project-id` are not execute via bash. It just replaces the value, and the new one will be `${GCP_GAR_PROJECT_ID}`, instead of `GCP_GAR_PROJECT_ID`, which then fails later because 
`${GCP_GAR_PROJECT_ID}` is not a ENV variable. 

We could add a separate paremeters:
```yml
  project_id_env:
    type: string
    default: GCP_GAR_PROJECT_ID 
  project_id:
    type: string
    default: ${GCP_GAR_PROJECT_ID}
```

With enterprise passing `project_id_env: GCP_GAR_ENTERPRISE_PROJECT_ID`.

That's at least what I gathered here: https://github.com/CircleCI-Public/gcp-gcr-orb/blob/master/src/commands/gcr-auth.yml

`google-project-id` is marked as an `env_var_name`.

PR #1109, opened 2026-03-09
docs: re-init docker compose setup documentation with Redis

Bnyro:
Hey,

This PR re-adds documentation about a Docker setup with Redis (apparently these docs never existed for autopush-rs) to simplify self-hosting it.

The changes are based on @p1gp1g's changes and notes from https://github.com/mozilla-services/autopush-rs/pull/813 and  https://codeberg.org/Sunup/android.

(rather unrelated) side notes:
- I noticed that [`config_options.md`](https://mozilla-services.github.io/autopush-rs/config_options.html) is not rendered for some reason? Locally, it didn't generate it either.
- [`config_options.md`](https://github.com/mozilla-services/autopush-rs/blob/master/docs/src/config_options.md) prefixes all autoend variables with `AUTOENDPOINT__` instead of `AUTOEND__`
-  Are there no pre-built Docker images? I've kept the build script to  I've built my own now (https://codeberg.org/bnyro/autopush-docker/packages), but it'd be nicer to have auto-built ones from upstream


docs/src/running.md
```diff
@@ -51,14 +40,28 @@ have the Autopush daemons use other ports.
...
     > ```
 
-> _**Note**_: The docker images used take approximately 1.5 GB of disk-space, make
+> _**Note**_: The docker images used take approximately 400 MB of disk-space, make
 sure you have appropriate free-space before proceeding.
 
-</div>
--->
+5. Modify the environment variables set in the `docker-compose.yml` file:
+    -  Replace `<CRYPTOKEYHERE>` with the output of `python3 scripts/fernet_key.py` (or [from this section](#generate-a-crypto-key)) in `AUTOCONNECT__CRYPTO_KEY` and `AUTOEND__CRYPTO_KEYS`.
```
jrconlin: So, there was [a patch](https://github.com/mozilla-services/autopush-rs/pull/1085) a while ago that makes this easier.
You can now do `make gen-key` and it will do the appropriate poetry set up and generate the keys for you.

It produces output like:
```
# Crypto Keys:
AUTOCONNECT__CRYPTO_KEY=["1JK...v-U="]
AUTOEND__CRYPTO_KEYS=["1JK...v-U="]


# Auth Key:
AUTOEND__AUTH_KEYS=["Ni...qw="]
```

(Yes, there's another pending ticket to allow for both `CRYPTO_KEY` and `CRYPTO_KEYS` because that annoys me too, but haven't had a chance to fix it yet.)

docs/src/running.md
```diff
@@ -51,14 +40,28 @@ have the Autopush daemons use other ports.
...
+
+    - Set `AUTOCONNECT__ENDPOINT_SCHEME`, `AUTOCONNECT__ENDPOINT_HOSTNAME`, `AUTOCONNECT__ENDPOINT_PORT` to the domain and `AUTOEND__ENDPOINT_URL` to match the public facing URL to reach `autoendpoint` (e.g. `update.push.example.com`).
```
jrconlin, line 52: FWIW, the Enterprise folk just had a problem where they failed to open a firewall for `AUTOCONNECT__ROUTER_PORT` between all the apps (Autoendpoint and autoconnect both use it to shuffle messages around.) It shouldn't be public, but not sure if we might want to include it here in the config. I updated [the docs](https://github.com/mozilla-services/autopush-rs/blob/master/docs/src/config_options.md) to talk about this.

(Not a huge deal for any config that only has one node, but if you have more than one, it's going to seriously trip you up. You won't get updates and have no idea why not unless you spelunk through the logs.)

docs/src/running.md
Bnyro: Good point, I've updated the docs accordingly (and added a hint that this should be executed inside the docker container or poetry has to be installed first)

> (Yes, there's another pending ticket to allow for both CRYPTO_KEY and CRYPTO_KEYS because that annoys me too, but haven't had a chance to fix it yet.)
>
That confused me too when setting things up, so good that you're aware of it :)

docs/src/running.md
Bnyro: I think that the comment at https://github.com/mozilla-services/autopush-rs/pull/1109/changes#diff-0cdc0193be4cc7c46bcc2eb217b04778e80360fb18d56a86ded959e374ae824dR16 should be enough, as usually this should work fine by default when using Docker (except if the admin decides to do some weird docker firewall changes)

jrconlin: Yeah, it's a hard one to call. There are some doc updates that I need to make now that I've gotten some feedback from the Enterprise folk. I'm fully expecting even more once folk start up their own docker instances. 
(of course, that doesn't solve the "most people never read docs" problem, but that's beside the point.)

PR #1110, opened 2026-03-10
PUSH-628 Replace postgres library with sqlx

gruberb:
Replace `tokio-postgres`/`deadpool-postgres` with `sqlx`.

  - Swapped out `tokio-postgres`, `postgres`, and `deadpool-postgres` for `sqlx` with built-in connection pooling
  - Unified error handling: `PgPoolError` variant is gone since `sqlx` handles pool and query errors under one type
  - Batch inserts (`add_channels`, `save_messages`) now use s`qlx::QueryBuilder::push_values` instead of manual parameter
  - Added initial migration file at `migrations/20240101000000_initial_schema.sql` (derived from `schema.psql`, meant to
  be run via `sqlx` migrate run)

For the migration: We have to figure out how to run the migration in the infra setup. 

### How to test

Spin up a local postgres:

```bash
docker run -d --name autopush-pg \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_PASSWORD=post_pass \
  -e POSTGRES_DB=autopush \
  -p 5432:5432 \
  postgres:16
```

Load the schema:
```bash
docker exec autopush-pg psql -U postgres -d autopush -c "CREATE SCHEMA IF NOT EXISTS autopush;"
docker exec -i autopush-pg psql -U postgres -d autopush \
  -c "SET search_path TO autopush, public; $(cat migrations/20240101000000_initial_schema.sql)"
```

  Run the tests:

```bash
POSTGRES_HOST="postgres:post_pass@localhost/autopush" \
  cargo test -p autopush_common --features postgres -- db::postgres
```

### DIFF
--- a/autopush-common/Cargo.toml
+++ b/autopush-common/Cargo.toml
@@ -62,7 +62,7 @@ prometheus-client = { version = "0.24", optional = true }
 protobuf = { version = "=2.28.0", optional = true } # grpcio does not support protobuf 3+
 form_urlencoded = { version = "1.2", optional = true }
 
-sqlx = { version = "0.8", features = ["runtime-tokio-rustls", "postgres"], optional = true }
+sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls", "postgres"], optional = true }
 deadpool-redis = { version = "0.22", features=["rt_tokio_1", "serde"], optional = true }
 redis-module = { version = "2.0", optional = true }
 redis = { version = "0.32", features=["aio", "tokio-comp"], optional = true } # blocked by deadpool-redis
--- a/autopush-common/src/db/postgres/mod.rs
+++ b/autopush-common/src/db/postgres/mod.rs
@@ -661,20 +661,21 @@ impl DbClient for PgClientImpl {
                 "INSERT INTO {tablename} ({field_names}) ",
                 tablename = &self.message_table(),
             ));
-        query_builder.push_values(msg_params.iter(), |mut b, mp| {
-            b.push_bind(uaid_str.clone());
-            b.push_bind(mp.channel_id.clone());
-            b.push_bind(mp.chidmessageid.clone());
-            b.push_bind(mp.version.clone());
-            b.push_bind(mp.ttl);
-            b.push_bind(mp.expiry);
-            b.push_bind(mp.topic.clone());
-            b.push_bind(mp.timestamp);
-            b.push_bind(mp.data.clone());
-            b.push_bind(mp.sortkey_timestamp);
-            b.push_bind(mp.headers.clone());
+        // `query` is provided by `push_values` to bind parameters for each row
+        query_builder.push_values(msg_params.iter(), |mut query, mp| {
+            query.push_bind(uaid_str.clone());
+            query.push_bind(mp.channel_id.clone());
+            query.push_bind(mp.chidmessageid.clone());
+            query.push_bind(mp.version.clone());
+            query.push_bind(mp.ttl);
+            query.push_bind(mp.expiry);
+            query.push_bind(mp.topic.clone());
+            query.push_bind(mp.timestamp);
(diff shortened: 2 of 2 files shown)


autopush-common/Cargo.toml
```diff
@@ -62,18 +62,16 @@ prometheus-client = { version = "0.24", optional = true }
 protobuf = { version = "=2.28.0", optional = true } # grpcio does not support protobuf 3+
 form_urlencoded = { version = "1.2", optional = true }
 
-tokio-postgres = { version="0.7", optional = true }
-postgres = {version="0.19", optional=true}
-
-deadpool-postgres = {version = "0.14", optional = true}
+sqlx = { version = "0.8", features = ["runtime-tokio-rustls", "postgres"], optional = true }
```
ncloudioj: Per the doc,

> For forward compatibility, you should use the separate runtime and TLS features as the combination features may be removed in the future.

```suggestion
sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls", "postgres"], optional = true }
```

ncloudioj: Do we also need the `migrate` feature?

ncloudioj (dismissed):
LGTM, r=nanj.

autopush-common/src/db/postgres/mod.rs
```diff
@@ -427,74 +405,40 @@ impl DbClient for PgClientImpl {
...
-            vars = Vec::from_iter((1..params.len() + 1).step_by(2).map(|v| format!(
-                "(${}, ${})",
-                v,
-                v + 1
-            )))
-            .join(",")
-        );
-        // finally, do the insert.
-        self.pool.get().await?.execute(&statement, &params).await?;
+        let mut query_builder: sqlx::QueryBuilder<sqlx::Postgres> =
```
jrconlin, line 409: Oh thank god. I hated that hack.

autopush-common/src/db/postgres/mod.rs
```diff
@@ -767,31 +656,45 @@ impl DbClient for PgClientImpl {
...
-            params.push(&mp.timestamp);
-            params.push(&mp.data);
-            params.push(&mp.sortkey_timestamp);
-            params.push(&mp.headers);
+        let mut query_builder: sqlx::QueryBuilder<sqlx::Postgres> =
+            sqlx::QueryBuilder::new(format!(
+                "INSERT INTO {tablename} ({field_names}) ",
+                tablename = &self.message_table(),
+            ));
+        query_builder.push_values(msg_params.iter(), |mut b, mp| {
```
jrconlin: I wonder if we should be more explicit about the variable names here, partly because `b` is kind of magicked in by the `push_values()` func. Not critical, and maybe a comment later.

autopush-common/src/db/postgres/mod.rs
```diff
@@ -920,32 +805,28 @@ impl DbClient for PgClientImpl {
         reliability_id: &str,
         new_state: crate::reliability::ReliabilityState,
     ) -> DbResult<()> {
-        let timestamp =
-            SystemTime::now() + Duration::from_secs(RELIABLE_LOG_TTL.num_seconds() as u64);
+        let timestamp_epoch = (std::time::SystemTime::now()
+            + Duration::from_secs(RELIABLE_LOG_TTL.num_seconds() as u64))
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap()
```
jrconlin: :twitch:

This should never fail, but still, `log_report` is not critical. We should log the error but not panic the thread.

autopush-common/Cargo.toml
jrconlin: +1 on separating out TLS.

gruberb: I don't think we need the `migrate` feature. We run the migrations outside of the binary.

PR #1125, opened 2026-03-19
fix: push enterprise gars via the GHA workflow

pjenvey:
where the auth seems simpler than in circleci

Closes PUSH-679

(This seems like the easiest fix, as it works: my test branch https://github.com/mozilla-services/autopush-rs/actions/runs/23274825321/job/67675479316 successfully published)

### DIFF
--- a/.github/workflows/mozcloud-publish.yaml
+++ b/.github/workflows/mozcloud-publish.yaml
@@ -3,6 +3,11 @@ name: Build, Tag and Push Container Images to GAR Repository
 on:
   pull_request:
     types: [opened, labeled, unlabeled, synchronize]
+  push:
+    branches:
+      - master
+    tags:
+      - '**'
   workflow_dispatch: {}
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/mozcloud-publish.yaml
```diff
@@ -17,7 +22,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
-    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@1b87069d293273436a84dff04954a8950d3ff9ca # v6.1.0
+    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@4784cb70739a4f32ce010921f60fb1ebbc791a38 # v6.2.2
```
jrconlin, line 25: Heh, it'd be nice if this was an actual tag rather than a commit reference, but that's a bigger issue.

.github/workflows/mozcloud-publish.yaml
```diff
@@ -26,6 +31,32 @@ jobs:
       docker_build_args: |
         CRATE=autoconnect
         BINARY=autoconnect
+
+  build-and-push-autoconnect-enterprise:
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      (
+        github.event_name == 'push' &&
+        (github.ref_name == 'master' || startsWith(github.ref, 'refs/tags/'))
```
jrconlin, line 40: Guess I should add in the `enterprise` ref_name here once I build that. (Kinda wondering if I should anyway, just to keep things discrete.)

PR #1127, opened 2026-03-21
feat: Add number of in-process requests to health check

jrconlin:
Closes #PUSH-678

### DIFF
--- a/autoendpoint/src/extractors/notification.rs
+++ b/autoendpoint/src/extractors/notification.rs
@@ -79,13 +79,7 @@ impl Drop for Notification {
         let _ =
             self.in_process_counter
                 .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
-                    if current > 0 {
-                        Some(current - 1)
-                    } else {
-                        // Should we add a metric here to track if we hit the floor?
-                        debug!("⚠️🧹 in_process_counter underflow.");
-                        None
-                    }
+                    current.checked_sub(1)
                 });
         trace!(
             "🧹 Dropping notification with message_id: {}",


autoendpoint/src/routes/health.rs
```diff
@@ -49,6 +49,7 @@ pub async fn health_route(state: Data<AppState>) -> Json<serde_json::Value> {
         "router_table": router_health,
         "message_table": message_health,
         "routers": routers,
+        "request_count":state.in_process_subscription_updates.load(std::sync::atomic::Ordering::Relaxed),
```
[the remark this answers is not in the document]
jrconlin: Consumers of this endpoint are machines. Clarity is not important.

autoendpoint/src/extractors/notification.rs
```diff
@@ -54,6 +60,38 @@ pub struct Notification {
...
+        // `Clone` or `Copy` can cause the counter to decrement too many times.
+        // We'll set a floor for now.
+        let _ =
+            self.in_process_counter
+                .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
+                    if current > 0 {
+                        Some(current - 1)
+                    } else {
+                        // Should we add a metric here to track if we hit the floor?
+                        debug!("⚠️🧹 in_process_counter underflow.");
```
ncloudioj: Would `debug!` be logged as well, if so, I think it's sufficient.

autoendpoint/src/extractors/notification.rs
```diff
@@ -54,6 +60,38 @@ pub struct Notification {
...
+                        None
+                    }
+                });
```
ncloudioj, line 83: Or use `usize::checked_sub()`
```suggestion
        if self
            .in_process_counter
            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
                current.checked_sub(1)
            })
            .is_err()
        {
            debug!("⚠️🧹 in_process_counter underflow.");
        }
```

ncloudioj (approved):

autoendpoint/src/extractors/notification.rs
jrconlin: It wouldn't be logged. We only log >info. 
This is here mostly because I wanted to have something local for some testing.

autoendpoint/src/extractors/notification.rs
jrconlin: Ah! Thanks! I forgot that gets inherited.

PR #1128, opened 2026-03-23
feat: Add memory check to __health__

jrconlin:
Add a snapshot of the percentage of memory currently in use by this node to our health check. This should align with the kubernetes metric, so this should just be a sanity check, but it may help spot nodes that could OOM soon.

Issue: PUSH-632

### DIFF
--- a/autoendpoint/src/routes/health.rs
+++ b/autoendpoint/src/routes/health.rs
@@ -22,7 +22,7 @@ use autopush_common::metrics::StatsdClientExt;
 #[cfg(feature = "reliable_report")]
 use autopush_common::util::b64_encode_url;
 
-/// get the local memory usage in bytes (presumes running under kubernetes)
+/// get the local memory usage in percentage of limit (presumes running under kubernetes)
 pub async fn memory_usage_percentage(memory_path: &str) -> Option<f64> {
     // If we can read (and there is a limit)
     if let Ok(mem_limit_str) = read_to_string(format!("{}/{}", memory_path, "memory.max")) {


autoendpoint/src/settings.rs
```diff
@@ -137,6 +139,10 @@ impl Default for Settings {
             reliability_retry_count: autopush_common::redis_util::MAX_TRANSACTION_LOOP,
             max_notification_ttl: Duration::from_secs(MAX_NOTIFICATION_TTL_SECS),
             disable_sentry: false,
+            // From imperical observation, kubernetes stores this in the main
```
ncloudioj, line 142:
```suggestion
            // From empirical observation, kubernetes stores this in the main
```

autoendpoint/src/routes/health.rs
```diff
@@ -21,6 +22,28 @@ use autopush_common::metrics::StatsdClientExt;
 #[cfg(feature = "reliable_report")]
 use autopush_common::util::b64_encode_url;
 
+/// get the local memory usage in percentage of limit (presumes running under kubernetes)
+pub async fn memory_usage_percentage(memory_path: &str) -> Option<f64> {
```
ncloudioj, line 26: No async code involved below, this function can be sync.

autoendpoint/src/routes/health.rs
```diff
@@ -21,6 +22,28 @@ use autopush_common::metrics::StatsdClientExt;
...
+    // If we can read (and there is a limit)
+    if let Ok(mem_limit_str) = read_to_string(format!("{}/{}", memory_path, "memory.max")) {
```
ncloudioj, line 28: If we're on 2024 edition, we can chain those if lets into one:

```rs
    // If we can read (and there is a limit)
    if let Ok(mem_limit_str) = read_to_string(format!("{}/{}", memory_path, "memory.max")) 
        && mem_limit_str.trim() != "max"
        && let Ok(mem_limit) = mem_limit_str.trim().parse::<u64>()
        && let Ok(mem_current_str) = read_to_string(format!("{}/{}", memory_path, "memory.current"))
        && let Ok(mem_current) = mem_current_str.trim().parse::<u64>()
    {
        // Stars have aligned, and we can return a value.
        return Some((mem_current as f64 / mem_limit as f64) * 100.0);
    }
```

jrconlin: Right, I'll apply this once the `update` PR lands.

PR #1138, opened 2026-04-06
docs: Update docs for enterprise

jrconlin:
* Remove docker references
* Fix discrepency between "crypto_key" vs "crypto_keys"
* add metrics discussion
* updated release docs
* added boot process

Closes: PUSH-653, PUSH-624, PUSH-664, PUSH-604


autoconnect/autoconnect-settings/src/lib.rs
```diff
@@ -78,7 +80,8 @@ pub struct Settings {
     /// The optional port override for the endpoint URL
     pub endpoint_port: u16,
     /// The seed key to use for endpoint encryption
-    pub crypto_key: String,
+    pub crypto_key: Option<String>, // obsolete, use `crypto_keys` instead
```
gruberb, line 83: Comment says obsolete?

autoconnect/autoconnect-settings/src/app_state.rs
```diff
@@ -45,19 +45,32 @@ pub struct AppState {
...
+                "Missing required configuration: {}__CRYPTO_KEY or {}__CRYPTO_KEYS",
+                ENV_PREFIX.to_uppercase(),
+                ENV_PREFIX.to_uppercase()
+            )));
+        };
+        let crypto_keys = &settings
+            .crypto_keys
+            .clone()
+            .unwrap_or_else(|| settings.crypto_key.clone().unwrap());
+        // TODO: switch to make_fernet(); move make_fernet() to autopush_common.
```
gruberb, line 59: Should we implement the TODO?

autoconnect/autoconnect-settings/src/app_state.rs
```diff
@@ -45,19 +45,32 @@ pub struct AppState {
...
+        if !(crypto_keys.starts_with('[') && crypto_keys.ends_with(']')) {
```
gruberb, line 62: Can we add a comment here, why starting with [ is invalid, what is this magic [?

autoconnect/autoconnect-settings/src/lib.rs
gruberb: can we give `s` a batter name? When reading the code, it's confusing what `s` means.

autoconnect/autoconnect-settings/src/lib.rs
```diff
@@ -247,6 +259,8 @@ impl Settings {
     pub fn test_settings() -> Self {
         // Provide test settings based on enabled features.
         // semi-hack to satisfy clippy --all --all-features
+        let crypto_keys = Some(format!("[\"{}\"]", Fernet::generate_key()));
```
gruberb, line 262: What is the semi-hack here? Why can't we use the usual format? And, could we not implement the `Dispaly` trait on `Fernet` so we can stringify them without `format` here?

autoconnect/autoconnect-settings/src/lib.rs
jrconlin: It is, but we want to persist the key to allow for older installs to not break suddenly.

autoconnect/autoconnect-settings/src/app_state.rs
jrconlin: Ideally, yes, but this PR is already complex. 

https://mozilla-hub.atlassian.net/browse/PUSH-174 should cover the work.

autoconnect/autoconnect-settings/src/app_state.rs
jrconlin: This is to be consistent with autoendpoint's usage. The `[]` imply a mock JSON "list" definition indicating multiple keys

autoconnect/autoconnect-settings/src/lib.rs
jrconlin: We're testing the way that the string should be provided. (e.g. as a JSON like set of potentially multiple values.) 
`Fernet` is foreign, and while we could add a Display trait, we shouldn't.

Issue #1118, opened 2026-03-17
Release Container/Docker Images public accessible


wrenix:
It would be nice, that i do not have to build the images at my own.
Please publish the container images public e.g. ghcr.io / here on github

┆Issue is synchronized with this [Jira PUSH-668](https://mozilla-hub.atlassian.net/browse/PUSH-668)

wrenix:
I believe the easiest way is, to create a github-action which build and push direct into ghcr.io.

If i write such an PR, do you like to merge it? (I know that this project has sometimes slow reviewing/reactions, see the urgancy PR, so i do not like to much spend time into github-based systems instatt of gitlab, when it does not get merged)

jrconlin:
I mean, I'm not one to turn away community contributions. (Sorry, really haven't had the band-width to dig into the Urgency PR, mostly because I'm trying to work out if there might be something missing, particularly with stored notifications. Also need to clarify with the front-end folk that it won't cause weird issues, and they've been swamped too.)

PR #1145, opened 2026-04-15
feat(ci): add container build support

wrenix:
solved #1118


wrenix:
The pipeline has build on my master-branch the correct latest images:
https://github.com/wrenix/autopush-rs/actions/runs/24522299084/job/71683097524

as matrix (with just latest and latest-redis), for:
  - autoconnect: https://github.com/wrenix/autopush-rs/pkgs/container/autopush-rs%2Fautoconnect
  - autoendpoint: https://github.com/wrenix/autopush-rs/pkgs/container/autopush-rs%2Fautoendpoint

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Generate container meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            ghcr.io/${{ github.repository }}/${{ matrix.component }}
+          tags: |
+            type=schedule
+            type=raw,value=latest,enable=${{matrix.features == 'production'}}
```
jrconlin, line 49: Let's not include `production` as a docker key here. We use a separate path for that deployment and I feel that it would confuse and frustrate folk looking to run stand-alone servers.

Let's specify the `-no-default-features` & `enterprise` feature set here.
(Technically, you could specify both `production` and `enterprise` if you wanted the most flexibility for data storage, but building the `grpc` stuff takes _a long time_ and kinda bloats things for folk that don't want it.)

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,82 @@
...
+            type=semver,pattern={{major}}.{{minor}}-${{matrix.features}},value=${{ steps.checkout.outputs.tag }}
+            type=semver,pattern={{major}}-${{matrix.features}},value=${{ steps.checkout.outputs.tag }}
+
+      - name: Build image
+        uses: redhat-actions/buildah-build@v2
+        id: build-image
+        with:
+          oci: true
+          archs:
+          build-args: |
```
jrconlin, line 64: Again, let's make sure we're building images that are friendly for "Stand-alone" folk, so we should specify `enterprise`.

.github/workflows/build-image.yaml
wrenix: Hmm, current we build the following images:
  - production (which is also pinned as latest)
  - redis

What do you think is usable images with feature sets?
If we has no-default-features is activated, what are we losing? (reliable reports)

wrenix: Should we build an:
   - redis with reliable report
   - postgresql without reliable report

should we create builds with log_vapid and/or stub?

And what is the default image?

wrenix: no i create just redis and postgresql:
https://github.com/wrenix/autopush-rs/actions/runs/24688906645

PR #1146, opened 2026-04-15
docs: Add discussion about post-quantum encryption

jrconlin:
I wanted to capture some discussion about post-quantum cryptography concerns.


p1gp1g:
Hey ! I've just seen your notes, and that's something I'm thinking about, so here are mines:
- AES128 is considered quantum safe. You have stated differently on your doc, I'm curious if I miss something or not. Otherwise, autopush endpoints are safe, as they are based on tokens AES encrypted
- aes128gcm push notification content is encrypted with 2 secrets: one calculated from ECDH (not PQ-safe) and a 128bits pre-shared key: the authentication secret. If I'm correct, this makes the encryption PQ-safe, as an attacker needs to know the 2 secrets:

RFC8291:
```
PRK_key = HMAC-SHA-256(auth_secret, ecdh_secret);
IKM = HMAC-SHA-256(PRK_key, key_info || 0x01)
```

- VAPID, a JWT asymmetrically signed with ECDSA on the P-256 curve is not PQ-safe. But authentication is not vulnerable to harvest now decrypt later. Nevertheless, we should probably start working on it. There are already some work on pq-jwts. In addition, it is good to note that endpoints are considered by RFC8030 as Capability endpoints (they should be managed as a secret), and VAPID is supposed to be an additional layer of security

I shared part of my thoughts to Filippo (Go crypto maintainer and cryptographer), he has likely not look in details, but he didn't invalidate my thoughts:
https://abyssdomain.expert/@filippo/116363341997258536
> PSKs are fine if you can keep them from being compromised. The scheme you excerpted should be fine if auth_secret is not known to the attacker.

jrconlin:
> AES128 is considered quantum safe.

I'm probably being overly paranoid. AES128 hasn't been identified as a method that can be broken by Quantum computing, and is generally considered safe, but since it's essentially it's based on EC, it may be possible to derive a potential set of valid values from the coordinates in the public key, thus making things a bit "fuzzy". Since it hadn't been cleared as Quantum Safe, I erred on the side of paranoia.

In reality, much like everything else, it always comes down to threat vectors. If you're using AES128 to secure your multi-billion dollar financial network, municipal infrastructure back-end, or other high value target system, then yeah, worth thinking about. If you're using it to encrypt notifications that the cat video post on Mastodon got a like, a lot less likely to be attacked. 

Same with Push notifications. To send a properly spoofed push notification to a victim you need to get:
* the VAPID key (which is stored on the server)
* The encryption key (which is stored on the client)
* The endpoint (which is also encrypted, and semi-ephemeral)

Is it possible to do nefarious things using this? Yeah, possibly, but holy wow is that a lot of work for very little return. There are far, far easier ways to compromise someone's systems (with things like Click-fix). 

I mean, one way to secure push notifications right now is to make the endpoint have a 48 hour lifespan (or 1/2 as long as the predicted breaking period, whatever that might be. The remote service could have it's client simply refresh or request a new endpoint and key set which resets all of the clocks.

Issue #1114, opened 2026-03-11
Endpoint to proxy to other push servers


p1gp1g:
UnifiedPush uses web push to get a decentralized push infrastructure for mobile. This is perfectly fine to use a self-hosted push server (autopush being one of them) with most applications, but it may be different with some others:
1. some (web) apps may use the push server domain to fingerprint the user
2. other apps restrict which push server they allow. This is done by simplicity, developers want to avoid any kind of SSRF, and they don't necessarily know there could be some other push servers than the big ones. For example:
    - [Matrix's sygnal gateway](https://github.com/matrix-org/sygnal/blob/4700184bac220125a51fed7ece428e233fb0c824/docs/applications.md?plain=1#L297)
    - [Firefox sync account](https://github.com/mozilla/fxa/blob/509d11e756f80ab156dd5c95301433825e908b53/packages/fxa-auth-server/config/index.ts#L1653)
    - [ntfy](https://github.com/binwiederhier/ntfy/blob/3c8ac4a1e14919705b305d7eddc89e546c7bac56/server/server_webpush.go#L20)
    - And many more proprietary apps (a few of them are listed [in this issue](https://codeberg.org/UnifiedPush/wishlist/issues/27))

As a solution, we can set a public service that works as a proxy for web push requests. The targeted endpoint would be encrypted, the proxy wouldn't have to store anything, and the requests' body length and content-encoding would be checked.

This solution would fix the first point: we can use the proxy with apps that we don't want to fingerprint us (mostly on web).
Regarding the second point, it depends. If the proxy is on a new domain that isn't already allowed, we need the different applications to change their code to allow this new service (and I didn't imagine there were that many apps doing that kind of allow-list).

Therefore if we had an endpoint on autopush to forward web push requests, autopush could be used for that purpose too. I don't think I'd need a lot of capacity, as nothing needs to be stored.

If that's OK, we could even benefit from Mozilla's hosted server already being in all application allow lists (for apps that have one): fixing the 2 issues in a row. In this case, I would be very happy to implement the feature. I don't expect it to be very complex.

And, if you think it is unlikely to be merged, do you think we could have a subdomain CNAMEd to a domain we own ? Most of the allow lists seems to use wildcards, it would fix the 2nd issue on many apps, but not all.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/PUSH-658)

PR #1150, opened 2026-04-17
Add endpoint to forward push notifications to another push server

p1gp1g:
For the context: #1114

This PR adds 2 endpoints:
- one to create a *forward endpoint*. It takes a JSON in input, with 3 parameters to the forward endpoint (cf. bellow)
- and the forward endpoint, to proxy the notification

The forward endpoint contains a fernet-encrypted token, that contains the targeted URL, the digest of the appserver public key (uncompressed format), and a private key (cf bellow)

It contains a private key to sign outgoing VAPID authorizations, because:
- we can't simply forward the incoming VAPID authorization header (different audience)
- we must use different VAPID key for all endpoints, else it would be possible to request any endpoint configured to work with autopush as soon as we know the (targeted) endpoint.

The private key is tested to not be the one used to send from the server, to avoid applications leaking their keys.

I've added tests for the 2 extractors.

I've some pending questions:
- Should I create a new ApiErrorKind for the endpoint to create new forward endpoints ? I often use ApiErrorKind::General there.
- Is it OK to disable redirections for app_state.http ? The redirection policy can't be set per-request for now, and it is only used for webpush requests to our own nodes.
- Should we had some metrics ?

The PR is a bit long, but half of the lines are the tests/documentation. I'm happy to edit the patch, and answer any question. I hope it will be considered, it answers a real need :)

### DIFF
--- a/autoendpoint/src/extractors/forward_request.rs
+++ b/autoendpoint/src/extractors/forward_request.rs
@@ -1,0 +1,582 @@
+use std::str::FromStr;
+
+use actix_http::header::{self, HeaderMap};
+use actix_web::{FromRequest, HttpRequest, web::Data};
+use autopush_common::{
+    metric_name::MetricName, metrics::StatsdClientExt as _, util::b64_encode_url,
+};
+use cadence::StatsdClient;
+use futures::{FutureExt, future};
+use openssl::{ec::PointConversionForm, hash::MessageDigest};
+
+use crate::{
+    error::{ApiError, ApiErrorKind, ApiResult},
+    extractors::subscription::{decode_public_key, repad_base64, validate_vapid_jwt},
+    headers::vapid::{VapidClaims, VapidError, VapidHeader, VapidHeaderWithKey, VapidVersionData},
+    server::AppState,
+    settings::Settings,
+};
+
+/// Contains data extracted from the encrypted token and the request, required to push to the new endpoint
+#[derive(Debug)]
+struct ForwardData {
+    /// Endpoint to forward notifications
+    url: String,
+    /// VAPID claims from the incoming authorization header, reused for outgoing request
+    vapid_claims: Option<VapidClaims>,
+    /// VAPID private key for outgoing request
+    vapid_privkey: Option<[u8; 32]>,
+}
+
+/// Contains the [reqwest::RequestBuilder] made from data extracted from the encrypted token
+/// via [ForwardToken]
+pub struct ForwardRequest(pub reqwest::RequestBuilder);
+
+impl FromRequest for ForwardData {
+    type Error = ApiError;
+    type Future = future::LocalBoxFuture<'static, Result<Self, Self::Error>>;
+
+    fn from_request(req: &HttpRequest, _payload: &mut actix_http::Payload) -> Self::Future {
+        let req = req.clone();
+        async move {
+            let app_state: Data<AppState> =
+                Data::extract(&req).await.expect("No server state found");
+
+            let encoded_token = req
+                .match_info()
+                .get("token")
+                .expect("{token} must be part of the webpush path")
+                .to_string();
+
+            let raw_token = app_state
+                .fernet
+                .decrypt(&repad_base64(&encoded_token))
+                .map_err(|e| {
+                    // Since we're decrypting and endpoint, we get a lot of spam links.
+                    // This can fill our logs.
+                    trace!("🔐 fernet: {:?}", e.to_string());
+                    ApiErrorKind::InvalidToken
+                })?;
+            forward_data_from_token(
+                raw_token,
+                &req.headers(),
+                &app_state.settings,
+                &app_state.metrics,
+            )
+        }
+        .boxed_local()
+    }
+}
+
+/// Get the [ForwardData] from the token and the request headers
+fn forward_data_from_token(
+    token: Vec<u8>,
+    req_headers: &HeaderMap,
+    settings: &Settings,
+    metrics: &StatsdClient,
+) -> ApiResult<ForwardData> {
+    if token.len() < 64 {
+        trace!("⏩️forward: token shorter than 64 bytes");
+        Err(ApiErrorKind::InvalidToken)?;
+    }
+    let i_pubkey = token.len() - 64;
+    let i_privkey = i_pubkey + 32;
+
+    let url = String::from_utf8(token[..i_pubkey].to_vec())
+        .ok()
+        .map(|s| url::Url::parse(&s).ok())
+        .flatten()
+        .ok_or_else(|| ApiErrorKind::InvalidToken)?;
+
+    let pubkey_hash: Option<[u8; 32]> = if token[i_pubkey..i_privkey] == [0u8; 32] {
+        None
+    } else {
+        token[i_pubkey..i_privkey].try_into().ok()
+    };
+
+    let vapid_privkey: Option<[u8; 32]> = if token[i_privkey..] == [0u8; 32] {
+        None
+    } else {
+        token[i_privkey..].try_into().ok()
+    };
+
+    trace!(
+        "⏩️forward: new request to {}, inbound vapid {}, outboud vapid {}",
+        url,
+        pubkey_hash.is_some(),
+        vapid_privkey.is_some()
+    );
+
+    let vapid_claims = get_vapid_claims(
+        req_headers,
+        pubkey_hash,
+        url.origin().unicode_serialization(),
+        settings,
+        metrics,
+    )?;
+
+    Ok(ForwardData {
+        url: url.into(),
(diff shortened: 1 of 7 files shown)


jrconlin:
Huh. 
Sorry, I missed the initial issue because I was swamped with other stuff.

So, if I understand, the idea is that you want to add a different endpoint so that Autopush becomes a forwarding proxy for some other service. Someone would register that endpoint using the new `/fpush/v1/new` endpoint, pass along the URL and get back a masked URL that could later be called with a payload. 

How would this not be used to deliver malware using Mozilla's servers as a masked gateway?

If I understand correctly, I, the sketchy owner of "https://EvilOnAStick.com", could compose a request to `https://updates.push.services.mozilla.org/fpush/v1/new` with a URL that goes to any server. I could then use that URL to deliver a payload (albeit, with some weird headers) to that remote server via the mozilla.com domain. I mean, that would be the very first thing I'd try if I saw this. 

It just seems to me that essentially, this PR turns autopush into an open proxy with a few more steps involved, and that's not great. Fortunately, this does limit the size of the payload, but that's not a huge barrier. We already have a few folk possibly doing questionable things with Autopush, and I'd like to not continue that trend.

(I'd also argue that even if there was a good way to protect from that, we might want to use a different router protocol system similar to how we use `webpush` to send messages between autopush connection nodes, but that still would mean delivering blind payloads to random servers where we're masking the origin.)

p1gp1g:
> Huh. Sorry, I missed the initial issue because I was swamped with other stuff.
> 
> So, if I understand, the idea is that you want to add a different endpoint so that Autopush becomes a forwarding proxy for some other service. Someone would register that endpoint using the new `/fpush/v1/new` endpoint, pass along the URL and get back a masked URL that could later be called with a payload.

That's the idea, but the requester doesn't get the response body. Autopush becomes an *application server* for the receiving push server. The reasons are detailed in the issue.

> How would this not be used to deliver malware using Mozilla's servers as a masked gateway?

Hm, so your concern would be someone using the endpoint to send a malware to another website via an open form? If that's the case, I think that's safe to remove the `Cookie` header (`Authorization` header is already removed or replaced), and replace `Content-Type` with `application/octet-stream` (as defined by RFC8188) - it should not be possible to upload an arbitrary form then.

Edit: We can even copy webpush defined headers only (TTL, Topic, and set Content-Type)

> If I understand correctly, I, the sketchy owner of "https://EvilOnAStick.com", could compose a request to `https://updates.push.services.mozilla.org/fpush/v1/new` with a URL that goes to any server. I could then use that URL to deliver a payload (albeit, with some weird headers) to that remote server via the mozilla.com domain. I mean, that would be the very first thing I'd try if I saw this.

If that's about the open form, then see my previous point. If that's about any DOS amplification: there is no point doing it, 1. autopush becomes a bottle neck, and 2. the outgoing payload is smaller than the request.

> It just seems to me that essentially, this PR turns autopush into an open proxy with a few more steps involved, and that's not great. Fortunately, this does limit the size of the payload, but that's not a huge barrier. We already have a few folk possibly doing questionable things with Autopush, and I'd like to not continue that trend.
> 
> (I'd also argue that even if there was a good way to protect from that, we might want to use a different router protocol system similar to how we use `webpush` to send messages between autopush connection nodes, but that still would mean delivering blind payloads to random servers where we're masking the origin.)

We can also add `X-Forwarded-For` header, which is used for that purpose

p1gp1g:
I've pushed a new commit that restrict the forwarded headers and enforce content-encoding, which prevents the API to fill arbitrary forms.

Regarding `X-Forwarded-For`: we should handle that differently depending if autopush is used behind a reverse-proxy. If it is being a reverse proxy: the header should be handled there, else we can use `req.peer_addr()`: Should it simply forward that header too?

EDIT: I think we can set `X-Forwarded-For` if `req.peer_addr()` is a global ip (cf. [is_global](https://doc.rust-lang.org/std/net/enum.IpAddr.html#method.is_global), a crate exposes it, we don't have to use nightly), else it must be set up by the reverse proxy

jrconlin:
By the way, my preemptive apologies for not staying on top of this. I've been pulled into a different project with a tight deadline, so I don't have a lot of bandwidth to look at this. I've set a few calendar reminders so that it doesn't fall off my radar again.

p1gp1g:
Thanks for the info and for having setup these reminders :+1:

jrconlin:
Sigh.

So, I've decided to step away from Mozilla, and thus, this PR is probably not going to land. For what it's worth, I understand the objectives and goal, but I'm not quite sure I agree with them.
My general sense is that while there are reasonable safeguards in place, it still essentially boils down to using Mozilla servers as a proxy for other servers. That feels questionable for a few reasons, including things that are not technical in nature. (e.g. A bad actor using our service as a shield, which might also expose Mozilla to various additional actions.) Yes, it's possible for them to use our servers directly to do bad things, but that's a risk we're taking on directly. Without a larger team, I'm very hesitant to add any more work. 

I'm sorry that this is both sudden, and probably not the news you wanted to hear.

p1gp1g:
> So, I've decided to step away from Mozilla

Well, it sounds like Mozilla will miss someone soon

> Without a larger team, I'm very hesitant to add any more work.

Well, I understand. Maybe it can be taken again later with the next shift.

Given the limited body size, the content-type, and if we remove all the other headers, the non-amplificiation of the body, it would make the feature very unattractive to exploit compare to other online services.

The other options to solve the different problems are:
- either patch the different application services to allow arbitrary push servers (including Mozilla account)
- or write another service that is client of Google push server or autopush, and does the proxy afterward

Anyway, if you're about to leave, it's not worth using your time discussing it.

**Thanks for all your work on the subject!** I wish you the best

jrconlin:
To be fair, I've been thinking really hard about this as well. One of the things I didn't really mention was that I was wondering if going through a bottleneck like this is really a good idea. In essence, you're placing trust in that bottleneck to be a good actor.  (Not saying that my fellows at Google, Apple, Microsoft, and AWS aren't, they're pretty stand-up folk when I dealt with them, but the future is never certain.) Problem is, i'm not really sure what the alternative might be. Adopting something more distributed feels better, but would absolutely impact delivery time. 

I'll probably continue to think about this for a while, and maybe might approach this from a more civilian angle in the future.

p1gp1g:
This feature is mainly for the browsers (Firefox forks like Fennec and IronFox, hopefully Firefox itself and Linux browsers - for Linux mobiles - one day).

To give more insights, I've been thinking about adding Web Push support to autopush, as we have FCM and APNS (WNS maybe a day), to support de-Google AOSP phones and Linux phones.

But that way, we end up with 2 web push servers. To me, it's less elegant and efficient. This is how I ended with this proxy.

If we look at the big picture, it is a bottleneck to browsers exactly the same way it is on Android with the Play Services for FCM, or on iOS.

Finally, I expect some users to use the proxy for some other apps, to hide their self-hosted push server. But I don't expect it to be the main goal of the feature. As I often say, mobile apps are usually "trusted" applications where we are logged in, usually with the same email address or phone number, that we access without any VPN (or alike), and that have access to many other fingerprint. The push server origin is more concerning for untrusted context, on websites in the browsers.

jrconlin:
So, obviously, the decision to implement this is up to the team that's going to own the service, but here are a few of my thoughts:

Push does a really good job protecting the identity of the sender and receiver. It does this by essentially making the server well known. There's really no idea of "hand off" between various push providers (good and bad about that), and in some respects, your PR may actually do that. The problem, though, still remains that in this scenario "Mozilla" is the "front and center" server. 

So, let's say you've got a group that does, I dunno, something evil. For whatever reason, that gets a bunch of folk to grab the pitchforks and torches and go after the thing that's visible. In that case, it's going to be mozilla, because we're the enabler of the kitten pornography or whatever. It's not really a great look for mozilla. As dumb as it sounds, brand protection is a consideration. 

You could argue that the evil folk could just use mozilla's push server directly, and you'd be right, but that would also let us have direct control over the connection. We could do things to block servers from using push directly, or otherwise prevent things. It's a little bit harder if we're a carrier system handling wrapped traffic. 

The next one isn't quite as bad, but it's worth thinking about. While Push is getting a lot of much needed attention right now, it's mostly in prep for it to be placed in "Maintenance Mode". That's basically a mode where it's left to itself, with occasional updates, possibly done by an LLM. I think it's going to be really hard to convince anyone at mozilla to add a new feature right now, particularly one that potentially carries either some risk or increased costs. 

Maybe it's worth considering a general update to WebPush? Maybe something a bit more distributed, with better encryption, that doesn't require a centralized server. Maybe one that takes what we learned from Push and tries a few improvements (like broadcast using MLS, or distribution using something like ActivityPub or revisit FIDO) There have been a lot of great advancements in tech over the past 12 years.

PR #1154, opened 2026-04-24
fix: preserve redis notification metadata

napiany:
## Description

Preserve `ttl`, `topic`, and `sortkey_timestamp` when serializing Redis-backed `StorableNotification` records.

The Redis save path creates zset members using `Notification::chidmessageid()`, which depends on `topic` or `sortkey_timestamp`. The current stored JSON omits those fields, so after deserialization the notification falls back to the legacy `channel_id:version` message ID. That makes reconnect/backlog cleanup delete the wrong Redis member/key and can cause repeated replay of stored notifications.

This patch keeps the metadata in the stored JSON.


pjenvey:
Have you (or anyone else) ran the redis backend against the integration test suite (and are they currently runnable against it)?

I'm wondering if they caught/would have caught such a bug, or should a new test showing this issue be added.

kschelonka:
Thank you for your contribution! Apologies that I missed this and duplicated the fix https://github.com/mozilla-services/autopush-rs/pull/1191

Let me see about getting this possibly merged in for contribution.

PR #1156, opened 2026-04-28
fix(redis): defaults for explicitly null keys in settings

kschelonka:
Apply same defaults for missing keys to null keys
for RedisDbSettings. Otherwise when using
unwrap_or_default the default is set to 0,
which is not a valid TTL and results in
errors on SET.

[PUSH-659]

[PUSH-659]: https://mozilla-hub.atlassian.net/browse/PUSH-659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autopush-common/src/db/redis/mod.rs
+++ b/autopush-common/src/db/redis/mod.rs
@@ -32,8 +32,10 @@ pub struct RedisDbSettings {
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
     pub create_timeout: Option<Duration>,
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
+    // Minimum value is 1 (second), defaults to MAX_ROUTER_TTL_SECS
     pub router_ttl: Option<Duration>,
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
+    // Minimum value is 1 (second), defaults to MAX_NOTIFICATION_TTL_SECS
     pub notification_ttl: Option<Duration>,
 }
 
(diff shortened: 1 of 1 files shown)


autopush-common/src/db/redis/mod.rs
```diff
@@ -73,6 +73,17 @@ impl TryFrom<&str> for RedisDbSettings {
                 "notification_ttl must be greater than 0".to_string(),
             ));
         }
+        // Supply defaults for explicitly null values (deserializer handles missing keys)
```
jrconlin, line 78: We should be very clear about this.

These are semi-override values for the defaults. Technically you _could_ have `0` for these, but we should probably make that more explicit than implicit because you never specified the value. Not really sure the best way to convey that, maybe in a doc comment?

kschelonka: If it's set to 0 by the user it will throw above though won't it?

kschelonka: specifically this [line](https://github.com/mozilla-services/autopush-rs/blob/a19ba4a53636c1cc5f295973366dc26449d1d0e7/autopush-common/src/db/redis/redis_client/mod.rs#L81-L82) will default to 0 if we don't explicitly handle the null case

jrconlin: Yeah, good point. We can say that the smallest minimum TTL for message and router storage is 1s. Partly because I don't want to change more code.

kschelonka: Will update the docs

jrconlin (approved):

PR #1165, opened 2026-06-10
chore(grpc): Replace grpcio with tonic

gruberb:
This PR attempts to fix the ongoing `grpcio` issues. Example:
https://github.com/mozilla-services/autopush-rs/issues/898

We cannot build Push on macOS for example, because of the `grpcio` dependency. This is needed, because we communicate via gRPC to Googles BigTable. 

The types come from https://crates.io/crates/googleapis-tonic-google-bigtable-v2 crate. 
How it works:

* The crate is auto-generated from Google's official [googleapis](https://github.com/googleapis/googleapis) proto definitions (google/bigtable/v2/*.proto) using tonic-build + prost. It's published from [bouzuya/googleapis-tonic](https://github.com/bouzuya/googleapis-tonic), which regenerates the whole `googleapis` surface as individual crates and is what `bigtable_rs` uses under the hood as well.
* It ships the prost message types (`RowFilter`, `Mutation`, `ReadRowsRequest`, ...) and the tonic client stub (BigtableClient) pre-compiled as Rust source, so cargo build just compiles Rust, which is why the macOS cc-rs/cmake breakage is gone.

It also removes dead code: `mutate_rows`, `delete_cells`, the admin-API `delete_rows` (plus `BigtableTableAdminClient`, `admin_metadata()`, the admin channel accessor, and the `Admin` error variant). None had callers.

To test: clone the repo and checkout the PR locally on macOS, and run `cargo check` or `cargo build`, and it should all be green. 

### How to test

1. Start docker 
2. Run `gcloud beta emulators bigtable start` 
3. In another terminal, set `BIGTABLE_EMULATOR_HOST` to `localhost:8086`
4. And then `./scripts/setup_bt.sh` 

Then you can 
1. `cd autopush-common` 
2. `cargo test --features=emulator,bigtable -- db::bigtable --nocapture`

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/merge.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/merge.rs
@@ -2,9 +2,10 @@ use std::collections::{BTreeMap, HashMap};
 use std::mem;
 use std::time::{Duration, SystemTime};
 
-use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::ReadRowsResponse;
-use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::read_rows_response::CellChunk;
-use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::read_rows_response::cell_chunk::RowStatus;
+use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::{
+    ReadRowsResponse,
+    read_rows_response::{CellChunk, cell_chunk::RowStatus},
+};
 use tonic::Streaming;
 
 use super::{FamilyId, Qualifier, RowKey, cell::Cell, error::BigTableError, row::Row};
--- a/autopush-common/src/db/bigtable/bigtable_client/merge.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/merge.rs
@@ -2,11 +2,22 @@ use std::collections::{BTreeMap, HashMap};
 use std::mem;
 use std::time::{Duration, SystemTime};
 
-use futures::StreamExt;
-use google_cloud_rust_raw::bigtable::v2::bigtable::{ReadRowsResponse, ReadRowsResponse_CellChunk};
-use grpcio::ClientSStreamReceiver;
+use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::ReadRowsResponse;
+use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::read_rows_response::CellChunk;
+use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::read_rows_response::cell_chunk::RowStatus;
+use tonic::Streaming;
 
 use super::{FamilyId, Qualifier, RowKey, cell::Cell, error::BigTableError, row::Row};
+
+/// Return whether this chunk signals a row reset.
+fn is_reset_row(chunk: &CellChunk) -> bool {
+    matches!(chunk.row_status, Some(RowStatus::ResetRow(true)))
+}
+
+/// Return whether this chunk commits the row.
+fn has_commit_row(chunk: &CellChunk) -> bool {
+    matches!(chunk.row_status, Some(RowStatus::CommitRow(_)))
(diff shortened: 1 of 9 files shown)


jrconlin:
I am mostly concerned that we're replacing our `google-cloud-rust` with another one, generated by a single user. (A prolific single user, but still, one without a lot of support.) We've had issues in the past with this (see `grpcio`) and I would prefer that we used something with a bit more of a sustainable structure to it, but I'm fine if the team is ok with doing that or will take on forking and supporting that package at some point in the future.

autopush-common/src/db/bigtable/bigtable_client/merge.rs
```diff
@@ -2,12 +2,23 @@ use std::collections::{BTreeMap, HashMap};
 use std::mem;
 use std::time::{Duration, SystemTime};
 
-use futures::StreamExt;
-use google_cloud_rust_raw::bigtable::v2::bigtable::{ReadRowsResponse, ReadRowsResponse_CellChunk};
-use grpcio::ClientSStreamReceiver;
+use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::ReadRowsResponse;
+use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::read_rows_response::CellChunk;
+use googleapis_tonic_google_bigtable_v2::google::bigtable::v2::read_rows_response::cell_chunk::RowStatus;
```
jrconlin: nit: These feel like they could be combined.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -12,16 +12,17 @@ use async_trait::async_trait;
...
-use google_cloud_rust_raw::bigtable::admin::v2::bigtable_table_admin_grpc::BigtableTableAdminClient;
-use google_cloud_rust_raw::bigtable::v2::bigtable::ReadRowsRequest;
-use google_cloud_rust_raw::bigtable::v2::bigtable_grpc::BigtableClient;
-use google_cloud_rust_raw::bigtable::v2::data::{RowFilter, RowFilter_Chain};
-use google_cloud_rust_raw::bigtable::v2::{bigtable, data};
-use grpcio::{Channel, Metadata, RpcStatus, RpcStatusCode};
-use protobuf::RepeatedField;
+use gcp_auth::TokenProvider;
+// The generated stubs keep both the service requests/responses and the data
+// types in a single `v2` module. Alias it twice to retain the historical
```
jrconlin: I appreciate history, but this seems weirdly confusing for folk who don't care about that (e.g. someone fixing this in the future.)

If we're going to swap to the new construct, we should use the new construct without the legacy stuff.

(Looking over the code, I can see the value of using either `bigtable` or `data`, but picking one and converting the other is fine.)

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -659,25 +595,25 @@ impl BigTableClientImpl {
...
-                // a `Timestamp granularity mismatch` error
-                set_cell.set_timestamp_micros((timestamp.as_millis() * 1000) as i64);
                 debug!("🉑 expiring in {:?}", timestamp.as_millis());
-                mutation.set_set_cell(set_cell);
-                mutations.push(mutation);
+                mutations.push(data::Mutation {
+                    mutation: Some(data::mutation::Mutation::SetCell(data::mutation::SetCell {
+                        family_name: family_id.clone(),
+                        column_qualifier: cell.qualifier.clone().into_bytes(),
+                        // Yes, this is passing milli bounded time as a micro. Otherwise I get
```
jrconlin: Curious if this is still a problem.

gruberb: Updated the comment. Based on what I researched, yes: Bigtable tables default to millisecond timestamps.

jrconlin (approved):
🎊 whoo-hoo! 🎊

PR #1166, opened 2026-06-10
feat(apns): Add ability for token-based connection to APNS

gruberb:
This adds support for token-based APNS auth using Apple's .p8 provider keys, alongside the existing certificate-based auth. 

The motivation is operational: APNS certificates expire yearly and renewing them means a manual trip through the Apple Developer portal, PEM conversion, newline escaping into the channels JSON, and a redeploy. Provider auth keys don't expire and one key covers every bundle ID in the team.

Also fixes a small bug found while in here: `create_client` was building its `ClientConfig` from `ApnsSettings::default()` instead of the configured settings, so `request_timeout_secs` and `pool_idle_timeout_secs` were silently ignored.

Closes https://mozilla-hub.atlassian.net/browse/PUSH-720


autoendpoint/src/routers/apns/settings.rs
```diff
@@ -27,6 +33,10 @@ pub struct ApnsChannel {
     /// e.g. `-----BEGIN PRIVATE KEY-----\n`
     pub cert: String,
     pub key: String,
+    /// The 10-character Apple key ID for the `.p8` auth key (token-based auth)
```
kschelonka, line 36: comment(non-blocking): this is a little clunky, wondering if it warrants a refactor of these fields to group together and/or user an enum for discriminated union. It's odd to have to pass the cert string despite having token

if so, might want to update the deserializer to be backwards compatible to be kind to external users

gruberb: True. I didn't want to break the existing config, but I guess we should focus on ergonomics here as well.

autoendpoint/src/routers/apns/router.rs
```diff
@@ -139,49 +149,80 @@ impl ApnsRouter {
         })
     }
 
-    /// Create an APNS client for the channel
+    /// Create an APNS client for the channel.
+    ///
+    /// If `key_id` and `team_id` are set, the client uses token-based
+    /// authentication with the `.p8` provider auth key in `key`. Otherwise it
+    /// falls back to certificate-based authentication with `cert`/`key`.
```
jrconlin, line 158: 👍🏻 Always have backups.

autoendpoint/src/routers/apns/settings.rs
jrconlin: Well, there may be a period of overlap, while you're testing out the new auth, where you could have both. It's also not a bad idea to have some form of backup, particularly when dealing with an external service. 

Noting  a preference or a fallback isn't a terrible idea.

jrconlin (dismissed):

kschelonka (approved):

PR #1179, opened 2026-06-29
chore(ci): move from circleci to gha

kschelonka:
Migrate to GHA actions and workflows from circleci, with mostly equivalent CI steps. 
Similar to circleci config.yml, the actions have been consolidated into one orchestrator (build-test-deploy.yml) and the workflows are invoked from there. This allows for cross-workflow dependencies. The exception is the preview action which builds without requiring tests, on the preview label.

Changes:

* Stricter requirements for image deployments -- now require all tests and checks in order to proceed with publishing:
    * enterprise images
    * test images (e2e and integration test)
    * reliability cron images
   
   (Previously these were functionally required because the checks must pass due to `main` branch protections, but this ensures that the workflow won't be invoked for tags created with failing tests)

* Decoupled artifact upload from deployment flow to ensure GCS hiccups don't block image deployments
* Small updates to dockerfiles per linting output

[PUSH-610]

There's still some duplicative stuff that could use another pass (the ways we set up docker emulator and building test docker images) and potential caching improvements for rust if we incorporate 3rd party actions. But I think this is an acceptable first pass. Longer-term streamline example is the integration test images that we build and use in analysis runs on argo are totally duplicative with the integration tests run in gha. Consider building real canary instead of re-running integration tests. Using cargo-chef for the service images (using the pattern of the integration test image build) can reduce build time by caching more layers.

[PUSH-610]: https://mozilla-hub.atlassian.net/browse/PUSH-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kschelonka:
The preview environments deployed successfully to the registry but are failing due to secret configuration

.github/actions/setup-rust/action.yml
```diff
@@ -0,0 +1,53 @@
+name: Set up Rust
```
ncloudioj: There is a [setup-rust-toolchain action](https://github.com/marketplace/actions/setup-rust-toolchain-for-github-ci) available.

ncloudioj (dismissed):
LGTM, thanks!

ncloudioj (approved):

PR #1191, opened 2026-07-06
fix(redis): persist topic/ttl/sortkey_timestamp in stored notifications

### LINKED ISSUE #1189: legacy message causes infinite log spam and high CPU utilisation
Currently autoconnect starts using the entire CPU core to spam log messages and fails to deliver new messages. I am using redis as a backing store if relevant.
Here is a example log message i get:
```
Jul 05 20:32:03.266 WARN 🚨 LEGACY MESSAGE!? Notification { channel_id: 189a1c07-12d5-438e-b253-84c893b0aaaf, version: "gAAAAABqSqLCWmut0KBxGBT4RyYS5wE7lhru7lL40Xj5PGdv6_JutK4cPqyP-MMrRpHErcQelOXhsL6n7Po9QcEvtgOocW2uJOJIILLDjBb7nrSdl5ji9b7yCPsRrgHa4DMqceujisQYFnzLzawxMBIoa0JKBF8Y-ncGyfz9FhsedHqRuOWmKnFPMKZu_Pe8skP29exYhSFY", timestamp: 1783276226, ttl: 0, topic: None, data: Some("eyJ1cmdlbnQiOnRydWV9"), sortkey_timestamp: None, headers: Some({"encoding": "aes128gcm"}), reliability_id: None, reliable_state: None }
```
which is repeated a few million times over the last hour. Right now my workaround is to restart both autoconnect and autoendpoint, which makes me wonder why it works as the message should still be in the redis db.
System information:

```
 - system: `"aarch64-linux"`
 - host os: `Linux 6.18.37, NixOS, 26.11 (Zokor), 26.11.19700101.dirty`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.95.2
System type: aarch64-linux
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/store/nw8x08z5y8xr0nlpslgn0h397d4j2jxc-source`
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/PUSH-736)

kschelonka:
Previously 'ttl', 'topic', and 'sortkey_timestamp' were omitted from from notification stored in Redis. This resulted chidmessageid() assuming legacy format and returning an incorrect key. Since the key didn't match a stored record, the deletion was a no-op and the check_storage loop never exited once the client reconnects and attempts to retrieve stored messages. This resulted in log spam that caused excess CPU usage. The loop only stops when the message's content key hits its ttl.
    
Only the Redis backend was affected since these fields are serialized in Bigtable.

Fixes #1189

Jira: [PUSH-736]

[PUSH-736]: https://mozilla-hub.atlassian.net/browse/PUSH-736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autopush-common/src/db/redis/mod.rs
+++ b/autopush-common/src/db/redis/mod.rs
@@ -101,14 +101,17 @@ pub struct StorableNotification {
     pub version: String,
     pub timestamp: u64,
     // Possibly stored values, provided with a default.
-    #[serde(default = "default_ttl", skip_serializing)]
+    // Note: Unlike client-facing `Notification`, these fields
+    // should round-trip faithfully through Redis and not
+    // `skip_serializing` unless truly None.
+    #[serde(default = "default_ttl")]
     pub ttl: u64,
     // Optional values, which imply a "None" default.
-    #[serde(skip_serializing)]
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub topic: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub data: Option<String>,
-    #[serde(skip_serializing)]
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub sortkey_timestamp: Option<u64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub headers: Option<HashMap<String, String>>,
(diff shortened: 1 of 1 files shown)


jrconlin:
Ah! Good Catch! (and d'oh).

That also explains why the gauntlet check never caught it as well.

Thanks!

jrconlin (approved):

PR #1199, opened 2026-07-15
fix(grpc): PUSH-746 Use connect_lazy for BigTable connections

gruberb:
The `grpcio` to `tonic` migration replaced `grpcio`'s lazily-connected channel with tonic's eager `connect().await`, which moved connection establishment into `pool.get()` so that connection failures returned before reaching the circuit breaker's `record_failure()` accounting in read_rows/mutate_row. As a result, a BigTable connectivity event produced connection failures the breaker never observed, so it never opened, and normal churn amplified into an unbounded reconnect storm. 

This switches channel creation to `connect_lazy()` so dialing happens on the first RPC inside the breaker-guarded paths (matching the old grpcio behavior), and drops the now-unnecessary `async` from the channel constructors since `connect_lazy()` performs no I/O.

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -62,7 +62,7 @@ const RELIABLE_LOG_FAMILY: &str = "reliability";
 /// /// In most use cases, converted to seconds through .num_seconds().
 pub const RELIABLE_LOG_TTL: TimeDelta = TimeDelta::days(60);
 
-pub(crate) const RETRY_COUNT: usize = 5;
+pub(crate) const RETRY_COUNT: usize = 2;
 
 /// Maximum gRPC message size (256MB), matching the prior grpcio configuration.
 const MAX_MESSAGE_LEN: usize = 1 << 28;
@@ -341,9 +341,13 @@ pub fn retry_policy(max: usize) -> RetryPolicy {
 
 fn retryable_internal_err(status: &Status) -> bool {
     match status.code() {
-        // tonic surfaces transport-level failures (connection resets, etc.)
-        // as `Unknown`.
-        Code::Unknown => true,
+        // Generated tonic clients synthesize this status when the Channel is
+        // not ready, before sending the RPC. Other Unknown statuses are
+        // ambiguous and must not be replayed indiscriminately.
+        Code::Unknown => status
+            .message()
+            .to_ascii_lowercase()
+            .starts_with("service was not ready:"),
         Code::Internal => [
             "rst_stream",
             "rst stream",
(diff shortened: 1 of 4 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -377,15 +470,27 @@ pub fn retryable_status(metrics: &Arc<StatsdClient>) -> impl Fn(&Status) -> bool
     }
 }
 
+pub fn retryable_read_status(metrics: &Arc<StatsdClient>) -> impl Fn(&Status) -> bool + '_ {
+    move |status| {
+        info!("GRPC read failure: {:?}", status);
+        let retry = retryable_read_err(status);
+        if retry {
+            metric(metrics, "RpcFailure", Some(&format!("{:?}", status.code())));
```
kschelonka: can this be moved up and replace log in 496?

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -138,7 +134,6 @@ pub struct BigtableClientManager {
     settings: BigTableDbSettings,
     dsn: Option<String>,
     connection: String,
-    metrics: Arc<StatsdClient>,
```
kschelonka, line 141: was this unused? why remove?

gruberb: We used `metrics` here for the `recycle` method. We now no longer do a RPC call to BigTable in this recycle method:

https://github.com/mozilla-services/autopush-rs/pull/1199/changes#diff-fea2e93a85f467f075dcf1ac88947b21fd6b6f9ddf33bf78fe934c602329df47L236-L244

kschelonka: This is public but seems just internal so I don't think we need a version bump for the api change

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -366,35 +459,30 @@ pub fn metric(metrics: &Arc<StatsdClient>, err_type: &str, code: Option<&str>) {
...
-        }
-        retry
-    }
-}
-
 pub fn retryable_bt_err(
     metrics: &Arc<StatsdClient>,
 ) -> impl Fn(&error::BigTableError) -> bool + '_ {
     move |err| {
-        debug!("🉑 Checking BigTableError...{err}");
```
kschelonka, line 384: No harm in keeping this debug statement I think

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -366,35 +459,30 @@ pub fn metric(metrics: &Arc<StatsdClient>, err_type: &str, code: Option<&str>) {
...
             error::BigTableError::Auth(_) => {
                 metric(metrics, "Auth", None);
-                true
+                return true;
             }
-            _ => false,
+            _ => return false,
+        };
+
+        info!("GRPC Failure: {:?}", status);
```
kschelonka, line 488: ah I didn't realize that this log existed previously, was worried about increasing expenses

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -62,7 +62,7 @@ const RELIABLE_LOG_FAMILY: &str = "reliability";
 /// /// In most use cases, converted to seconds through .num_seconds().
 pub const RELIABLE_LOG_TTL: TimeDelta = TimeDelta::days(60);
 
-pub(crate) const RETRY_COUNT: usize = 5;
+pub(crate) const RETRY_COUNT: usize = 2;
```
kschelonka, line 71: this default retry count change should be documented

kschelonka: This also affects health check retries (which aren't configurable via settings, only uses this value). I'm ok with moving forward since we don't want to retry storm a failing health check anyway, but it might be helpful to decouple data retries from health check retries in the future

kschelonka (approved):

PR #1203, opened 2026-07-20
fix(config)!: require user-set AUTOEND__AUTH_KEYS

kschelonka:
Guard against booting the server with default (insecure) keys. Startup now fails unless auth_keys contains at least one non-empty key. Default no longer provides a static key.

BREAKING CHANGE: servers depending on the default auth_keys will no longer start until a real key is set.

[PUSH-748]

Note - this does not affect any Mozilla deployments (the value was always set).

I also updated the table in the docs to use the correct prefix (`AUTOEND__` not `AUTOENDPOINT__`)

[PUSH-748]: https://mozilla-hub.atlassian.net/browse/PUSH-748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/settings.rs
+++ b/autoendpoint/src/settings.rs
@@ -119,7 +119,7 @@ impl Default for Settings {
             // presume base64 encoding, so we can bump things up to 5630 bytes max.
             max_data_bytes: 5630,
             crypto_keys: format!("[{}]", Fernet::generate_key()),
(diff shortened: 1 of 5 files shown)


gruberb (commented):
Looks good!! 

It seems like we need to supply `AUTOEND__AUTH_KEYS` to our CI docker image.

gruberb (approved):

PR #1208, opened 2026-07-21
fix(bigtable): restore connection keepalive; retry transport errors

kschelonka:
gRPC's C-core provided connection-liveness behavior implicitly; tonic does not have this, which surfaced as transport error (BrokenPipe) failures and elevated p95 5xx latency (~30s, consistent with TCP retransmission timeout).

This change improves dead connection detection and retries transient broken pipe failures on reads.


gruberb:
I added a commit to this PR. We have the handle the whole BigTable RPC lifecycle on our side. This is the smallest change I could possible achieve (for now). We should defintely write this up in a doc.

kschelonka:
superseded by https://github.com/mozilla-services/autopush-rs/pull/1209

PR #1209, opened 2026-07-23
fix(bigtable): tonic connection and retry management

gruberb:
This PR addresses the three issues observed during the tonic rollout: excessive Bigtable connections, elevated p95 error latency, and failures caused by reaped transports. [c808beb](https://github.com/mozilla-services/autopush-rs/commit/c808beb442477d3d90d28cd62653415cebefa58c) separates logical pool handles from a small, configurable set of shared tonic channels. [c39195c](https://github.com/mozilla-services/autopush-rs/commit/c39195c7782144193f8ea71bcd587d33c5cae14c) adds per-operation attempt deadlines and total retry budgets. [3e826cc](https://github.com/mozilla-services/autopush-rs/commit/3e826cca22af2ea86b490f6aafc98222b4717e36) proactively refreshes channels and safely retries eligible operations on a fresh transport.


autopush-common/src/db/reporter.rs
```diff
@@ -22,20 +22,27 @@ pub fn spawn_pool_periodic_reporter(
...
+        metrics
+            .gauge_with_tags("database.pool.waiting", status.waiting as u64)
+            .send();
+    }
+
+    // Tonic multiplexes many logical client handles over each HTTP/2 channel,
+    // so deadpool size is not the physical connection count.
+    if let Some(count) = db.configured_channel_count() {
+        metrics
+            .gauge_with_tags("database.pool.channels.configured", count as u64)
```
kschelonka: Suggest renaming to 

"database.connections" or "database.channels" because this is actually the open connection count (and is independent of inflight requests)

autopush-common/src/db/reporter.rs
```diff
@@ -22,20 +22,27 @@ pub fn spawn_pool_periodic_reporter(
...
+            )
+            //.with_tag("hostname", hostname)  // Do not include hostname due to cardinality
+            .send();
+        metrics
+            .gauge_with_tags("database.pool.idle", status.available as u64)
+            .send();
+        metrics
+            .gauge_with_tags("database.pool.waiting", status.waiting as u64)
+            .send();
+    }
```
kschelonka, line 42: Suggest renaming these to reflect that they're rpc operations and not channels/connections, which is how I interpret the current dashboard. Maybe something like:

"database.pool.active" -> "database.ops.inflight"
"database.pool.idle" -> "database.ops.available"
"database.pool.waiting" -> "database.ops.queued"

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -1,36 +1,215 @@
...
+    ) {
+        for index in 0..channels.len() {
+            let weak = Arc::downgrade(channels);
+            let metrics = metrics.clone();
+            channels.runtime.spawn(async move {
+                channel_refresh_loop(weak, index, min_period, max_period, metrics).await;
+            });
+        }
+    }
+}
```
kschelonka: I've been using the bigtable client for reference implementation and this is a significant behavioral difference I'm concerned about. the [bigtable client](https://github.com/googleapis/google-cloud-cpp/blob/95511adfa1a6e733ca20dd17aaefd898611fe2f5/google/cloud/bigtable/internal/connection_refresh_state.cc#L1) never recreates a new channel, only pokes it to reconnect.

especially in a busy service like this with a relatively small number of channels per pod, where channels are pretty unlikely to go idle in the first place, I think we should try to avoid proactive connection churn. Tonic will rebuild dead transports on subsequent requests, and we're implementing the read-retry for the request that hit the dead transport, so rotation isn't necessary for reconnection.

tonic doesn't expose an AsyncWaitConnectionReady equivalent but we could use the PingAndWarm nudge as a cheap alternative (like the stub refresher in the linked code).

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -192,21 +427,15 @@ impl Manager for BigtableClientManager {
...
-            &self.settings.table_name,
-        );
-        debug!("🏊 Bigtable connection acquired");
+        let entry = BigtableDb::new(self.token_provider().await?);
+        debug!("🏊 Bigtable client handle acquired");
         Ok(entry)
     }
 
-    /// Recycle if the connection has outlived it's lifespan.
+    /// Recycle if the client handle has outlived its lifespan.
```
kschelonka: I think this is vestigial; the client doesn't have a meaningful lifespan anymore since it no longer has the channel. just reduce to `Ok(())`?

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -560,7 +1019,7 @@ impl BigTableClientImpl {
             .unwrap_or(Duration::from_secs(MAX_ROUTER_TTL_SECS))
     }
 
-    /// Spawn a task to periodically evict idle connections
+    /// Spawn a task to periodically evict idle logical client handles.
```
kschelonka: per recycle comment, I don't think we need this

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -110,32 +319,44 @@ impl BigTablePool {
...
+    }
+
+    /// Select a shared channel for one RPC attempt. Retried operations call
+    /// this again so a dead transport does not consume the entire retry budget.
+    pub(super) fn next_channel(&self) -> Channel {
+        self.pool.manager().channels.next()
+    }
 }
 
 fn sweeper(pool: &deadpool::managed::Pool<BigtableClientManager>, max_idle: Duration) {
```
kschelonka: per recycle and spawn_sweeper, pretty sure we don't need this anymore since not a real connection

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -216,14 +445,14 @@ impl Manager for BigtableClientManager {
             && Instant::now() - metrics.created > timeout
         {
             debug!("🏊 Recycle requested (old).");
-            return Err(RecycleError::message("Connection too old"));
+            return Err(RecycleError::message("Client handle too old"));
         }
         if let Some(timeout) = self.settings.database_pool_max_idle
```
kschelonka: note that removing sweeper removes all reference to `database_pool_connection_ttl` and `database_pool_max_idle`, should clean these up too if remove

autopush-common/src/db/bigtable/mod.rs
```diff
@@ -60,32 +96,81 @@ pub struct BigTableDbSettings {
...
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
     pub database_pool_wait_timeout: Option<Duration>,
-    /// Max time(in seconds) to recycle a connection
+    /// Max time (in seconds) to recycle a client handle.
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
     pub database_pool_recycle_timeout: Option<Duration>,
-    /// Max time (in seconds) a connection should live
+    /// Max time (in seconds) a pooled client handle should live. Tonic channel
```
kschelonka: Per recycle comment we can remove this I believe

docs/src/datastore.md
```diff
@@ -32,7 +32,15 @@ Bigtable's database configuration options are stored in the `db_settings` as a s
 * `message_family` the Bigtable cell family that refers to the **message** data collections
 * `router_family` the Bigtable cell family that refers to the **router** data collections
 * `table_name` the Bigtable internal path URI to the table. You construct this by using the following template `projects/{YOUR PROJECT}/instances/{INSTANCE NAME}/tables/{TABLE NAME}`. For example, if I had created a GCP project named "test-prod", that has an instance ID of "test-prod-us" which contained a table named "auto-test", the `table_name` would be `projects/test-prod/instances/test-prod-us/tables/auto-test`.
-* `retry_count` the number of retries after the initial Bigtable data operation. It defaults to 2 (reduced from 5 to limit retry storms). Health checks also use 2 retries, but their retry count is not separately configurable.
+* `retry_count` the number of retries after the initial Bigtable data operation. It defaults to 2 (reduced from 5 to limit retry storms). Health checks use the same configured retry count.
+* `database_pool_max_size` the maximum number of concurrent Bigtable operations. Pool entries are lightweight client handles, not physical connections.
```
kschelonka: clarify that this is the maximum number of operations we're essentially allowing to be inflight. the actual maximum number of concurrent operations depends on the number of channels and how many concurrent streams are allowed per channel. If this value is set above the actual capacity of the channel pool then operations will queue silently via tonic

autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -1583,12 +2123,22 @@ impl DbClient for BigTableClientImpl {
     }
 
     async fn health_check(&self) -> DbResult<bool> {
-        Ok(self
-            .pool
-            .get()
-            .await?
-            .health_check(&self.metrics.clone(), &self.settings.app_profile_id)
-            .await?)
+        // Use a random key so health checks do not create a hot tablet. The
```
kschelonka, line 2126: nice idea

autopush-common/src/db/bigtable/mod.rs
```diff
@@ -31,7 +31,37 @@ use tonic::metadata::MetadataMap;
...
+
+const DEFAULT_GRPC_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const DEFAULT_GRPC_POINT_ATTEMPT_TIMEOUT: Duration = Duration::from_secs(5);
+const DEFAULT_GRPC_SCAN_ATTEMPT_TIMEOUT: Duration = Duration::from_secs(20);
+const DEFAULT_GRPC_POINT_TOTAL_TIMEOUT: Duration = Duration::from_secs(15);
+// Kept strictly under the 30s GCP backend-service default `timeoutSec`, so a
+// stalled scan returns our own error before the load balancer gives up on the
+// request. At exactly 30s the two race and the client sees an LB 5xx instead,
+// which would also hide this budget from any p95 measured on inbound requests.
+const DEFAULT_GRPC_SCAN_TOTAL_TIMEOUT: Duration = Duration::from_secs(25);
```
kschelonka, line 44: 👍

autopush-common/src/db/bigtable/mod.rs
```diff
@@ -60,32 +90,58 @@ pub struct BigTableDbSettings {
...
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
     pub database_pool_create_timeout: Option<Duration>,
-    /// Max time (in seconds) to wait for a socket to become available
+    /// Max time (in seconds) to wait for a logical operation slot.
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_opt_u32_to_duration")]
     pub database_pool_wait_timeout: Option<Duration>,
-    /// Max time(in seconds) to recycle a connection
+    /// Max time (in seconds) to recycle a client handle.
```
kschelonka: nit: this is an orphaned config now that `recycle` does nothing

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -16,21 +20,103 @@ use crate::db::bigtable::{BigTableDbSettings, BigTableError, bigtable_client::Bi
...
+/// Default number of shared channels.
+///
+/// Google sizes a Bigtable pool from measured peak concurrency `C`, at about
+/// twice the count needed for saturation: `2 * (C / 50) = C / 25`. We do not
+/// have `C` yet, so this default is picked for failure isolation instead. A slot
+/// stalled inside tonic's `Reconnect` parks every request queued behind it for
+/// the duration of the dial, so four is the smallest count where one stalled
+/// slot taxes a quarter of attempts rather than half. Set `grpc_channel_count`
+/// from `ceil(C / 25)` once peak `database.ops.inflight` is known.
+/// https://cloud.google.com/bigtable/docs/configure-connection-pools
```
kschelonka, line 44: just noting, this is the reasoning behind https://mozilla-hub.atlassian.net/browse/PUSH-755

need to tag by workload since bigtable shared by autoconnect and autoendpoint, and also divide by number of pods

autopush-common/src/db/bigtable/pool.rs
```diff
@@ -16,21 +20,103 @@ use crate::db::bigtable::{BigTableDbSettings, BigTableError, bigtable_client::Bi
...
+}
+
+/// A fixed set of lazily-connected channels, selected round-robin.
+///
+/// Slots are never replaced. A tonic `Channel` owns a `Reconnect` that re-dials
+/// on the next `poll_ready` after a transport failure, so a channel recovers on
+/// its own. Bigtable's middleware deletes a connection that has not seen a
+/// request in five minutes, but that reap is a graceful GOAWAY: the retry path
+/// classifies it as retryable and reselects a different slot, and the documented
+/// cost is a cache-miss latency blip rather than a failure.
```
kschelonka: let's document here that conditional writes won't be replayed, so e.g. add_user on a freshly reaped channel will be a failure and retried by the client.

kschelonka (dismissed):
some comments but non-blocking, thanks!

kschelonka (approved):

PR #1218, opened 2026-07-29
Fix/bigtable reconnect storm

gruberb:
When redeploying autoconnect, the pods are restarting, and existing clients are connecting to existing, healthy pods. This increases the amount of messages for each endpoint. 

1. Our channel limit of 4 was slightly too conservative. This PR increases it to 16.
2. We don't count messages waiting in a queue towards the circuit breaker anymore.


autopush-common/src/db/bigtable/mod.rs
```diff
@@ -90,7 +90,7 @@ pub struct BigTableDbSettings {
     pub message_topic_family: String,
     #[serde(default)]
     pub database_pool_max_size: Option<u32>,
-    /// Number of shared tonic channels used for Bigtable RPCs. Defaults to four.
+    /// Number of shared tonic channels used for Bigtable RPCs. Defaults to sixteen.
```
kschelonka, line 93: I configured this using env vars, don't need to use a deployment to manage this (it's set to 6)

PR #1219, opened 2026-07-30
fix(docs): add missing AUTHEND__AUTH_KEYS docs to docker compose setup guide

Bnyro:
Since https://github.com/mozilla-services/autopush-rs/commit/1080af8b8dda8e1251cbf24431ed7dfe15bbf678, `AUTOEND__AUTH_KEYS` are required, so the docs for the docker compose with redis template also have to be updated.


kschelonka (approved):
Thank you!

PR #1223, opened 2026-08-08
fix(autoendpoint): update FCM error mappings to HTTP status codes

kschelonka:
Report RESOURCE_EXHAUSTED as 429 instead of 502, with errno 201 to hint exponential backoff. Forward the Retry-After header from FCM if provided, clamped between 10 and 3600s, otherwise use the original default of 120s.

Report transient UNAVAILABLE error as 503 instead of 502, which hints exponential backoff as well.

The original 502 status codes do not carry backoff hints, and retries plausibly worsen throttling or traffic congestion for FCM.

Note that this does have implications for 5xx tracking -- ~90% of autoendpoint's 5xx errors are currently due to FCM throttling, which will be moved to 429. This is intentional and desired, so that we can separate service degradation from being rate-limited and have appropriate retry hints.

[PUSH-759], [PUSH-760]

[PUSH-759]: https://mozilla-hub.atlassian.net/browse/PUSH-759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PUSH-760]: https://mozilla-hub.atlassian.net/browse/PUSH-760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/routers/fcm/client.rs
+++ b/autoendpoint/src/routers/fcm/client.rs
@@ -2,16 +2,46 @@ use crate::routers::RouterError;
 use crate::routers::common::message_size_check;
 use crate::routers::fcm::error::FcmError;
 use crate::routers::fcm::settings::{FcmServerCredential, FcmSettings};
+use actix_web::http::header::HttpDate;
 use reqwest::StatusCode;
 use serde::Deserialize;
 use std::collections::HashMap;
 use std::path::Path;
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 use url::Url;
 use yup_oauth2::authenticator::DefaultAuthenticator;
 use yup_oauth2::{ServiceAccountAuthenticator, ServiceAccountKey};
 
 const OAUTH_SCOPES: &[&str] = &["https://www.googleapis.com/auth/firebase.messaging"];
+
+// Firebase docs recommend a minimum 10 second wait for any retry.
+const MIN_RETRY_AFTER_SECS: u64 = 10;
+const MAX_RETRY_AFTER_SECS: u64 = 3600;
+
+/// Parse an upstream `Retry-After` into delta-seconds. Accepts delta-seconds
+/// or an HTTP-date.
+///
+/// A date already in the past means "retry now" and is floored
+/// to the minimum rather than discarded. Anything unparseable yields `None`,
+/// leaving the caller to its own default.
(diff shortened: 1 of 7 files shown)


autoendpoint/src/routers/fcm/client.rs
```diff
@@ -2,17 +2,47 @@ use crate::routers::RouterError;
...
+use std::time::{Duration, SystemTime};
 use url::Url;
 use yup_oauth2::authenticator::DefaultAuthenticator;
 use yup_oauth2::{ServiceAccountAuthenticator, ServiceAccountKey};
 
 const OAUTH_SCOPES: &[&str] = &["https://www.googleapis.com/auth/firebase.messaging"];
 
+// Firebase docs recommend a minimum 10 second wait for any retry.
+const MIN_RETRY_AFTER_SECS: u64 = 10;
+const MAX_RETRY_AFTER_SECS: u64 = 3600;
```
gruberb, line 20: Why did we choose 3600? What if FCM sends us 7200? We would overwrite it and then try again sooner? I guess we don't want to wait... 2 days (not sure if FCM would ever send such a high value), but then we should solve this by either disregarding it or doing something different.

gruberb (commented):
Not sure about the `3600` value. Seems like if FCM sends a higher number, we just make the problem worse again by trying sooner?

autoendpoint/src/routers/fcm/client.rs
kschelonka: I'll admit it was somewhat arbitrary, based on defaults I've used before for maximum exponential backoff limit. I don't anticipate FCM will exceed this value, but if it does then it would be after continual retries and likely fall into the [category described here](https://firebase.google.com/docs/cloud-messaging/scale-fcm): "If a request is continually retried with exponential backoff and is still failing 60 minutes later, it is either miscategorized as a retryable error, or FCM is experiencing an outage where retries may be inadvertently exacerbating the situation." 

Is there an alternative value you'd propose?

gruberb: As discussed, I think we can leave the 3600, but add a comment why we choose it and the likely impact it has?

gruberb (approved):

PR #1224, opened 2026-08-10
fix(bigtable): enforce router record expiry at read time

kschelonka:
fix(bigtable): enforce router record expiry at read time

Fold expiry into policy filter for router records.

Router reads and check-and-mutate predicates now
hide expired records immediately instead of waiting on server-side GC, ahead of enabling the router family max_age GC policy.

Adds an emulator test for the partial-expiry skew case.

Server-side GC enabled here: https://github.com/mozilla/webservices-infra/pull/12098

[PUSH-29]

[PUSH-29]: https://mozilla-hub.atlassian.net/browse/PUSH-29?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autopush-common/src/db/bigtable/bigtable_client/mod.rs
+++ b/autopush-common/src/db/bigtable/bigtable_client/mod.rs
@@ -175,30 +175,46 @@ pub struct BigTableClientImpl {
     circuit_breaker: Arc<CircuitBreaker>,
 }
 
-/// Return a a RowFilter matching the GC policy of the router Column Family
-fn router_gc_policy_filter() -> bigtable::RowFilter {
+/// Return a RowFilter matching the `maxversions=1` portion of a family's GC
+/// policy (keep only the most recent cell per column). Shared by the router and
+/// message family filters below.
+fn max_versions_filter() -> bigtable::RowFilter {
     bigtable::RowFilter {
         filter: Some(bigtable::row_filter::Filter::CellsPerColumnLimitFilter(1)),
     }
 }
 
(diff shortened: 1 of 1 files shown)


autopush-common/src/db/bigtable/bigtable_client/mod.rs
```diff
@@ -175,30 +175,46 @@ pub struct BigTableClientImpl {
     circuit_breaker: Arc<CircuitBreaker>,
 }
 
-/// Return a a RowFilter matching the GC policy of the router Column Family
-fn router_gc_policy_filter() -> bigtable::RowFilter {
+/// Return a RowFilter matching the `maxversions=1` portion of a family's GC
+/// policy (keep only the most recent cell per column). Shared by the router and
+/// message family filters below.
+fn max_versions_filter() -> bigtable::RowFilter {
```
ncloudioj, line 181: Nit: this can be a constant (`MAX_VERSION_FILTER`).

ncloudioj (approved):
LGTM, r=nanj.

autopush-common/src/db/bigtable/bigtable_client/mod.rs
kschelonka: yes, but it reads consistently with the others this way, and this is just a rename of the original function without code changes (it's no longer the full router gc policy but a component)

PR #1226, opened 2026-08-20
fix(autoconnect): add crypto_keys setting, deprecate crypto_key

kschelonka:
Multiple keys were already accepted in the crypto_key setting; this updates the naming to make this clear and be the same as autoendpoint.

Non-breaking, deprecation warning for old setting.

[PUSH-174]

[PUSH-174]: https://mozilla-hub.atlassian.net/browse/PUSH-174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/reuse-check-and-test.yml
+++ b/.github/workflows/reuse-check-and-test.yml
@@ -79,7 +79,7 @@ jobs:
       - name: Check formatting and clippy
         run: |
           cargo fmt -- --check
-          cargo clippy --all --all-targets --all-features -- -D warnings --deny=clippy::dbg_macro
+          cargo clippy --all --all-targets --all-features -- -D warnings -A deprecated --deny=clippy::dbg_macro
 
   test-unit:
     runs-on: ubuntu-latest
--- a/autoconnect/autoconnect-settings/src/lib.rs
+++ b/autoconnect/autoconnect-settings/src/lib.rs
@@ -78,6 +78,7 @@ pub struct Settings {
     /// The optional port override for the endpoint URL
     pub endpoint_port: u16,
     /// The seed key to use for endpoint encryption (deprecated: use crypto_keys instead)
+    #[deprecated(since = "{version}", note = "Use `crypto_keys` instead")]
     #[serde(default)]
     pub crypto_key: String,
     /// The cryptographic keys to use for endpoint encryption. Format: [key1,key2,...].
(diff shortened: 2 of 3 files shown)


autoconnect/autoconnect-settings/src/lib.rs
```diff
@@ -77,8 +77,13 @@ pub struct Settings {
     pub endpoint_hostname: String,
     /// The optional port override for the endpoint URL
     pub endpoint_port: u16,
-    /// The seed key to use for endpoint encryption
+    /// The seed key to use for endpoint encryption (deprecated: use crypto_keys instead)
+    #[serde(default)]
```
ncloudioj, line 82: Rust has this attribute for deprecation. 
```suggestion
    #[deprecated(since = "{version}", note = "Use `crypto_keys` instead")]
    #[serde(default)]
```

ncloudioj (dismissed):

autoconnect/autoconnect-settings/src/lib.rs
kschelonka: thank you!

ncloudioj (dismissed):
Oops, looks like some tests using the deprecated field are failing.

kschelonka:
> Oops, looks like some tests using the deprecated field are failing.

I had to allow deprecation warnings in the linter

.github/workflows/reuse-check-and-test.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       - name: Check formatting and clippy
         run: |
           cargo fmt -- --check
-          cargo clippy --all --all-targets --all-features -- -D warnings --deny=clippy::dbg_macro
+          cargo clippy --all --all-targets --all-features -- -D warnings -A deprecated --deny=clippy::dbg_macro
```
ncloudioj, line 82: You can also attach `#[allow(deprecated)]` to those tests to reduce the exemption area.

ncloudioj (approved):

.github/workflows/reuse-check-and-test.yml
kschelonka: the 2-stage deprecation was intentional, so I guess I'd rather have it be hinted in the IDE (shows up as a strikethrough for me) rather than having to do #[allow(deprecated)] in multiple areas?

PR #1231, opened 2026-09-01
feat(bigtable): count records that would be gc'd

kschelonka:
Emulating server-side GC with filters resulted in greater 410s and reduced send tab success. Re-deploy like a COUNT rule, matching read/write behavior in 1.83.1, so we can get metrics on impact with destination metadata

[PUSH-29]

[PUSH-29]: https://mozilla-hub.atlassian.net/browse/PUSH-29?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/autoendpoint/src/routes/registration.rs
+++ b/autoendpoint/src/routes/registration.rs
@@ -169,29 +169,30 @@ pub async fn get_channels_route(
     path_args: RegistrationPathArgsWithUaid,
     app_state: Data<AppState>,
 ) -> ApiResult<HttpResponse> {
-    let uaid = path_args.user.uaid;
+    // The extractor already read the user (and 410s when there isn't one), so
+    // reuse it rather than doubling check-in's db round trips
+    let mut user = path_args.user;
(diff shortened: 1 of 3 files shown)


autoendpoint/src/routes/registration.rs
```diff
@@ -169,29 +169,30 @@ pub async fn get_channels_route(
     path_args: RegistrationPathArgsWithUaid,
     app_state: Data<AppState>,
 ) -> ApiResult<HttpResponse> {
-    let uaid = path_args.user.uaid;
+    // The extractor already read the user (and 410s when there isn't one), so
+    // reuse it rather than doubling check-in's db round trips
```
kschelonka, line 173: removed this duplication so we're not double-counting impact of gc

gruberb (approved):

PR #1232, opened 2026-09-02
feat(metrics): check in additional ua types

kschelonka:
Include mobile devices in ua parsing (note: affects multiple metrics). Previously mobile devices were collapsed to "Other". 

Also add router destination to mobile device check-in metric, which is a good proxy for at least ios vs. android if ua parsing fails to get a more specific device.

https://mozilla-hub.atlassian.net/browse/PUSH-767

### DIFF
--- a/autopush-common/src/util/user_agent.rs
+++ b/autopush-common/src/util/user_agent.rs
@@ -27,7 +27,7 @@ const VALID_UA_OS: &[&str] = &[
 /// The tag value for a user agent we could not resolve.
 const UA_METRIC_UNKNOWN: &str = "Other";
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct UserAgentInfo {
     _user_agent_string: String,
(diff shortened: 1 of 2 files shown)


autopush-common/src/util/user_agent.rs
```diff
@@ -9,9 +9,25 @@ const VALID_UA_BROWSER: &[&str] = &["Chrome", "Firefox", "Safari", "Opera"];
...
+    "iPhone",
+    "iPad",
+    "iPod",
+];
 
-#[derive(Clone, Debug, Default)]
+/// The tag value for a user agent we could not resolve.
+const UA_METRIC_UNKNOWN: &str = "Other";
+
+#[derive(Clone, Debug)]
```
gruberb: Bit of a nit, but we don't need `Clone` here.

gruberb (dismissed):

gruberb (approved):

PR #1234, opened 2026-09-04
delete acked notifications in increment_storage to stop redelivery

### LINKED ISSUE #1228: Notifications still sent after ACK
# Notifications still sent after ACK

I noticed that notifications continue to be sent even after the client ACKs them. When I reconnect, I get the same messages all over again.

Here's what's happening. When a client sends an ACK like this:

```json
{"messageType":"ack","updates":[{"channelID":"<id>","version":"<version>","code":100}]}
```

The ACK reaches the server, but the message isn't actually removed from storage. So on the next reconnect, `check_storage` fetches it again and delivers it a second time.

I tracked this down to a few issues in how the ACK handling talks to Redis:

## The main issue: `increment_storage` is using the wrong Redis key

In `redis_client/mod.rs:412`, `increment_storage` queries `exp_list_key` to find messages to delete:

```rust
let exp_id_list = con.zrangebyscore(&exp_list_key, 0, timestamp).await?;
```

The problem is that `exp_list_key` is sorted by **expiry time** (`timestamp + ttl`), but the ACK system is passing in the message's **timestamp**. So when ACKing a message with `timestamp=100`, it's looking for messages with `expiry <= 100`, which finds nothing since expiry times are ~86400 seconds into the future.

It should be querying `msg_list_key` instead, which is sorted by the actual message timestamps.

## A few related issues

There are also some other problems that compound this:

1. **`save_message` uses the wrong score** (`redis_client/mod.rs:382`): It's using `sec_since_epoch()` instead of `storable.timestamp` for the `msg_list_key` score, so messages aren't sorted by their actual timestamp.

2. **`fetch_timestamp_messages` uses an inclusive range** (`redis_client/mod.rs:513`): When fetching messages after `timestamp=102`, it includes messages with `timestamp=102`. This should be exclusive (`timestamp + 1`) to avoid re-fetching the boundary message.

3. **`unacked_stored_highest` isn't cleared** (`on_server_notif.rs:366`): After `increment_storage` runs, this pointer stays set, so the next storage check uses the same timestamp again.

## Why this wasn't caught

This is a bit of a perfect storm:
- It only shows up when you ACK messages and then reconnect before they expire
- No errors are thrown, so it's a silent failure
- The natural Redis TTL eventually cleans up messages, which masks the bug
- Tests cover the individual operations but not the full ACK→reconnect cycle

## Fix

I've got a fix for this that:
- Uses `storable.timestamp` for the `msg_list_key` score
- Queries `msg_list_key` in `increment_storage` instead of `exp_list_key`
- Makes the fetch range exclusive by adding 1 to the timestamp
- Clears `unacked_stored_highest` after incrementing storage

Happy to submit a PR if this makes sense.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/PUSH-765)

sorooshm78:
Fixes https://github.com/mozilla-services/autopush-rs/issues/1228

  ACKed notifications were never removed from Redis storage, so clients
  received them again on reconnect:

  - `increment_storage` queried `exp_list_key` (scored by expiry time) with a
    message timestamp, so the range was always empty and nothing was deleted.
    It now queries `msg_list_key`, which is scored by message timestamp.
  - `save_message` scored `msg_list_key` with the wall clock instead of
    `storable.timestamp`.
  - `fetch_timestamp_messages` used an inclusive lower bound, re-fetching the
    message at the resume point; now exclusive (+1), matching Bigtable's
    `StartKeyOpen`.

  The `unacked_stored_highest` clearing mentioned in the issue was dropped as
  unnecessary (see issue discussion).

  Adds `increment_storage_removes_acked`, a deterministic regression test
  (the existing `test_expiry`/`wipe_expired` only fail on the old code when
  the wall clock ticks a second mid-test). Redis tests need a live
  `REDIS_HOST` — CI coverage is coming via #1229.

### DIFF
--- a/autopush-common/src/db/redis/mod.rs
+++ b/autopush-common/src/db/redis/mod.rs
@@ -6,7 +6,7 @@
 /// `autopush/co/{uaid}` u64 to store the last time the user has interacted with the server
 /// `autopush/timestamp/{uaid}` u64 to store the last storage timestamp incremented by the server, once messages are delivered
 /// `autopush/channels/{uaid}` List to store the list of the channels of the user
-/// `autopush/msgs/{uaid}` SortedSet to store the list of the pending message ids for the user
+/// `autopush/msgs/{uaid}` SortedSet to store the list of the pending message ids for the user, scored by the message's millisecond `sortkey_timestamp` (matching the ordering of Bigtable's message row keys)
 /// `autopush/msgs_exp/{uaid}` SortedSet to store the list of the pending message ids, ordered by expiry date, this is because SortedSet elements can't have independent expiry date
 /// `autopush/msg/{uaid}/{chidmessageid}`, with `{chidmessageid} == {chid}:{version}` String to store
 /// the content of the messages
--- a/autopush-common/src/db/redis/redis_client/mod.rs
+++ b/autopush-common/src/db/redis/redis_client/mod.rs
@@ -379,7 +379,7 @@ impl DbClient for RedisClientImpl {
             // The function [fetch_timestamp_messages] takes a timestamp in input,
             // here we use the timestamp of the record
             .zadd(&exp_list_key, msg_id, expiry)
-            .zadd(&msg_list_key, msg_id, sec_since_epoch());
+            .zadd(&msg_list_key, msg_id, storable.timestamp);
 
         let _: () = pipe.exec_async(&mut con).await?;
         self.metrics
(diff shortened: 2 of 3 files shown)


kschelonka:
https://github.com/mozilla-services/autopush-rs/pull/1229 was merged; please also include removing the added xfails so we can get accurate test report

sorooshm78:
I'll check the CI failure soon and fix it.

kschelonka:
Thanks for chasing this down. You're right that `increment_storage` is querying the wrong key. Unfortunately this approach regresses `test_msg_limit` because the range delete is operating on second-level resolution, so it also deletes undelivered messages enqueued in the same second. Timestamp messages deliberately persist in the database to be eventually cleaned up by TTL, and delivery is managed by the ack pointer. Bigtable doesn't have this same problem because the keys operate on millisecond resolution which dramatically reduces the collision likelihood. Updating redis to use the same millisecond parity as well as fixing the pointer is probably the way to go, but this is sneakily complicated...

sorooshm78:
One migration note for existing deployments: messages saved before this change carry second-resolution scores. An old seconds-valued pointer just acts as a very low floor (harmless over-delivery that self-heals via TTL); once a client acks a batch of new-format (ms-scored) messages, the pointer moves to ms and any leftover old-format entries simply expire by TTL instead of redelivering. Strictly better than the pre-fix behavior (endless redelivery), and bounded by the notification TTL.

sorooshm78:
Hi, could someone please take a look at this PR when you get a chance? Thanks!

kschelonka:
Sorry about the delay. We don't support autopush full time. Since this has implications for enterprise deployments we need to do due diligence about implications for over-delivery due to the ms-resolution change.

